PR #10, opened 2016-10-18
Styling für PasswordUpdateDialog

niklaushug:
Aktuell sind noch alle Styles im Package PasswordUpdateDialog enthalten. Ich beabsichtige das Styling entsprechend den Todo's im Code und dem heute verschickten Mail zu erweitern. Aus meiner Sicht macht es Sinn zuerst die Styling Grundlagen zu schaffen, bevor dieser Commit in das Haupt-Repository übernommen wird. 

Da die Dependencies (Bootstrap und Font Awesome) die gleichen wie in Nice 2.11 sind, sind wenn überhaupt, nur geringe Side-Effects zu erwarten. Probleme könnte die hohe Spezifität in Kundenprojekten verursachen.


MethodenMann:
please dont commit changes to version in lerna.json. otherwise travis will fail

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordInput.js
```diff
@@ -22,6 +23,7 @@ PasswordInput.propTypes = {
   value: React.PropTypes.string,
   onChange: React.PropTypes.func.isRequired,
   readOnly: React.PropTypes.bool,
+  disabled: React.PropTypes.bool,
```
rzueger: Why did you add this prop type? You don't use the `disabled` prop.

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordInput.js
```diff
@@ -10,6 +10,7 @@ const PasswordInput = props => (
       value={props.value}
       onChange={e => props.onChange(e.target.value)}
       readOnly={props.readOnly === true}
+      disabled={props.readOnly === true}
```
rzueger: I think we shouldn't use the `readOnly` prop to set both `readOnly` and `disabled`, because `readOnly` and `disabled` is not the same thing.

Do you want the password inputs to be read only or disabled? (see this post for a detailled expanation: http://stackoverflow.com/questions/7730695/whats-the-difference-between-disabled-disabled-and-readonly-readonly-for-ht)

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordMatchDisplay.js
```diff
@@ -2,14 +2,14 @@ import React from 'react'
 import {FormattedMessage} from 'react-intl'
 
 const PasswordMatchDisplay = props => {
+  // TODO state should change immediately also if typing in other fields, this field can only be valid if all other checks are fullfilled
   const message = (props.passwordRepeat && props.password !== props.passwordRepeat)
-    ? <FormattedMessage id="client.action.passwordUpdate.noMatch"/>
-    : '\u00a0'
-
+    ? <li className="text-danger"><FormattedMessage id="client.action.passwordUpdate.noMatch"/></li>
+    : <li className="text-success"><FormattedMessage id="client.action.passwordUpdate.Match"/></li>
```
rzueger: `match` lower case: `client.action.passwordUpdate.match`

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordUpdateDialog.scss
```diff
@@ -1,54 +1,206 @@
...
+$fa-font-path: "../../../../../node_modules/font-awesome/fonts/";
+@import "../../../../../node_modules/font-awesome/scss/font-awesome.scss";
+
+
+/*****
+* tocco-client boilerplate
+* for styling convenience while developing, only used inside monorepo tocco-client
+*/
+
+// TODO replace #root with #tocco-client-development
```
rzueger: why do you want to change this id?

And should this rule really be placed in this file? It's only for the development environment.

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordUpdateDialog.scss
```diff
@@ -1,54 +1,206 @@
...
+      position: absolute;
+      left: -$icon-list-indent;
+      content: $icon-list-item-1;
+    }
   }
+}
 
-  .FailureMessage {
-    margin-top: 1em;
+// replace icon for list items with a special class
```
rzueger: should this comment be a `TODO` too (like the other comments)?

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordInput.js
niklaushug: My intention was do prevent focus readonly fields by using tab key. If an user want to add new informations this is advantage, if an user want to copy data from a readonly field this is a disadvantage. My personal choice is the first option.

From my perspective sending readonly data has no point, it only takes more bandwidth.

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordUpdateDialog.scss
niklaushug: Id's has to be unique in html document by convention. On package level we can not assure that. I would like to replace the id by a css class. Semantically is "root" the wrong term, because it indicates the very top level element of a whole, which is only true in a single package. Right now we use this selector for boilerplate styles, which helps while developing but is unwanted in production. There are better class names than my proposal. Do you have preference?

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordUpdateDialog.scss
niklaushug: This is meant as information for developers to understand concealed purpose of a declaration or property. Every comment which starts with TODO is a open task. All other comments remain until the end of time. Would you like to indicate informations by INFO or something like that.

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordMatchDisplay.js
niklaushug: I will change that.

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordInput.js
rzueger: In this case we should set the input to `disabled` and remove the `readOnly` attribute. It doesn't make sense to set both of them (or does it?)

niklaushug: It makes sense to use attribute disabled only. I have tested all possible attribute combinations with https://jsfiddle.net/niklaushug/yd1LL9kf/ for newest browser versions: FF, CH, SA, OP as well as IEE, IE11, IE10 and IE9. Apart of varying default stylings for attribute readonly the behaviour is for all browsers the same.

attribute readonly: protects value, adds some default styling
attribute disabled: protects value, prevent focus by tab key, does not submit value with post request

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordUpdateDialog.scss
niklaushug: /Users/tocco/Projects/tocco-client/packages/action-password-update/src/main.js:
   58:   const mountElement = document.getElementById('root')

#root ist not specific enough (naming conflict with other widgets). Mount will fail if login widget is embedded more than once.

packages/action-password-update/src/components/PasswordUpdateDialog/PasswordUpdateDialog.scss
niklaushug: I have extended TODO with WEB and CLIENT to indicate, which team has the focus for a certain issue. It is only meant as a proposal.

niklaushug:
Fork is not under maintenance.

PR #17, opened 2016-10-28
refactor(action-merge): add permission info to the merge result

ottrob:
- Relations which could not be merged are already displayed
  in the merge response. But If the user has no read
  permissions to see these relations we must not show them
  Therefore the service now returns a new flag
  `showPermissionMessage`. If this flag is set to `true`
  An additional message will be displayed.

### DIFF
--- a/packages/action-merge/src/components/MergeWizard/MergeResponse.js
+++ b/packages/action-merge/src/components/MergeWizard/MergeResponse.js
@@ -53,10 +53,9 @@ class MergeResponse extends React.Component {
           responseEntities={this.props.mergeResponse.notDeletedEntities}
         />
         {(this.props.mergeResponse.showPermissionMessage)
-          ? <div className="alert alert-info">
-            <FormattedMessage id="client.entityoperation.action.merge.missingReadPermissions"/>
-          </div>
-          : null
+        && <div className="alert alert-info">
+          <FormattedMessage id="client.entityoperation.action.merge.missingReadPermissions"/>
+        </div>
         }
         <button
           className="btn btn-primary close-button"
--- a/packages/action-merge/src/components/MergeWizard/MergeResponse.js
+++ b/packages/action-merge/src/components/MergeWizard/MergeResponse.js
@@ -1,9 +1,9 @@
 import React from 'react'
 import {ExternalEvents} from 'tocco-util'
-import {FormattedMessage} from 'react-intl'
+import {FormattedMessage, intlShape} from 'react-intl'
 
 export const EntityResponseTable = props => {
-  if (!props.responseEntities || props.responseEntities.length === 0) return (<div/>)
+  if (!props.responseEntities || props.responseEntities.length === 0) return <div/>
   return (
     <div>
       <h4>{props.title}</h4>
@@ -38,22 +38,8 @@ EntityResponseTable.propTypes = {
   responseEntities: React.PropTypes.array.isRequired
 }
 
-export const PermissionMessageBox = props => {
-  if (!props.showPermissionMessage || props.showPermissionMessage === false) return <div/>
-  return (
-    <div className="alert alert-info">
-      <FormattedMessage id="client.entityoperation.action.merge.missingReadPermissions"/>
-    </div>
-  )
-}
-
-PermissionMessageBox.propTypes = {
-  showPermissionMessage: React.PropTypes.bool.isRequired
-}
-
 class MergeResponse extends React.Component {
   render() {
-    console.log('this.props.mergeResponse', this.props.mergeResponse)
     return (
       <div className="merge-response">
         <h1><FormattedMessage id="client.entityoperation.action.merge.responseTitle"/></h1>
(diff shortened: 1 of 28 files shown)


packages/action-merge/src/components/MergeWizard/MergeResponse.js
```diff
@@ -39,6 +38,19 @@ EntityResponseTable.propTypes = {
   responseEntities: React.PropTypes.array.isRequired
 }
 
+export const PermissionMessageBox = props => {
+  if (!props.showPermissionMessage || props.showPermissionMessage === false) return <div/>
```
MethodenMann: i would move the logic to show/hide this box to MergeResponse component

MethodenMann: || props.showPermissionMessage === false is not needed, !props.responseEntities covers all cases

packages/action-merge/src/components/MergeWizard/MergeResponse.js
```diff
@@ -1,10 +1,9 @@
 import React from 'react'
 import {ExternalEvents} from 'tocco-util'
-import {FormattedMessage, injectIntl} from 'react-intl'
+import {FormattedMessage} from 'react-intl'
 
-const EntityResponseTable = props => {
-  console.log('props.responseEntites', props.responseEntities)
-  if (!props.responseEntities || props.responseEntities.length === 0) return <div/>
+export const EntityResponseTable = props => {
+  if (!props.responseEntities || props.responseEntities.length === 0) return (<div/>)
```
MethodenMann: remove brackets around <div/> or add also in showPermissionMessage to be consistent

packages/action-merge/src/components/MergeWizard/MergeResponse.js
```diff
@@ -39,6 +38,19 @@ EntityResponseTable.propTypes = {
...
+  )
+}
+
+PermissionMessageBox.propTypes = {
+  showPermissionMessage: React.PropTypes.bool.isRequired
+}
+
 class MergeResponse extends React.Component {
   render() {
     console.log('this.props.mergeResponse', this.props.mergeResponse)
```
MethodenMann: remove productive console.log

packages/action-merge/src/components/MergeWizard/MergeResponse.spec.js
```diff
@@ -0,0 +1,86 @@
...
+          notCopiedRelations: [
+            { pk: "pk", entity: "entity", name: "name" }
+          ],
+          notDeletedEntities: [],
+          showPermissionMessage: false
+        }
+
+        const wrapper = shallow(
+          <MergeResponse
+            mergeResponse={{
```
MethodenMann: why not use mergeResponse const from above?

packages/action-merge/src/components/MergeWizard/MergeResponse.spec.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react'
+import MergeResponse from './MergeResponse'
+import {EntityResponseTable, PermissionMessageBox} from './MergeResponse'
+import {mount, render, shallow} from 'enzyme'
+import {FormattedMessage} from 'react-intl'
+
+const intl = {
```
rzueger: There is a file called `intlStub.js` in the `login` package which contains all required methods (`formatDate`, `formatTime`, ...). Maybe you could move this file somewhere else so that it can be used here too.

The `intl` prop type in `<MergeResponse>` should be `intlShape.isRequired` instead of `PropTypes.object`.

packages/action-merge/src/components/MergeWizard/MergeResponse.js
rzueger: I personally would remove the brackets

packages/action-merge/src/components/MergeWizard/MergeResponse.spec.js
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+        expect(wrapper.find(EntityResponseTable)).to.have.length(2)
+        expect(wrapper.find(PermissionMessageBox)).to.have.length(1)
+      })
+
+      it('should render a table', () => {
+        const wrapper = shallow(
+          <EntityResponseTable
+            title=''
```
rzueger: Our linting configuration enforces double quotes in JSX. We should follow this rule in the tests as well, even though they're not linted at the moment.

packages/action-merge/src/components/MergeWizard/MergeResponse.spec.js
```diff
@@ -0,0 +1,86 @@
...
+          />
+        )
+
+        expect(wrapper.find(FormattedMessage)).to.have.length(0)
+      })
+
+      it('should render info-box', () => {
+        const wrapper = shallow(
+          <PermissionMessageBox
+            showPermissionMessage={true}
```
rzueger: Our linting configuration wants us to omit values for boolean attributes. We should follow this rule in the tests as well, even though they're not linted at the moment.

packages/action-merge/src/components/MergeWizard/MergeResponse.js
```diff
@@ -54,6 +52,12 @@ class MergeResponse extends React.Component {
           title={this.props.intl.formatMessage({id: 'client.entityoperation.action.merge.notDeletedEntitiesTitle'})}
           responseEntities={this.props.mergeResponse.notDeletedEntities}
         />
+        {(this.props.mergeResponse.showPermissionMessage)
+          ? <div className="alert alert-info">
+            <FormattedMessage id="client.entityoperation.action.merge.missingReadPermissions"/>
+          </div>
+          : <div/>
```
MethodenMann: i think the program will behave the same without the empty div in case of `false`?

PR #19, opened 2016-10-28
refactor(login): Improve error handling in sagas

rzueger:
If fetch requests fail, errors are caught and logged and a generic failure
response is returned.

### DIFF
--- a/packages/login/src/modules/sagas.js
+++ b/packages/login/src/modules/sagas.js
@@ -17,15 +17,30 @@ export const DEFAULT_TIMEOUT = 30
 export const textResourceSelector = (state, key) => state.intl.messages[key] || key
 export const loginSelector = state => state.login
 
-export function doLoginRequest(data) {
-  return fetch(`${__BACKEND_URL__}/nice2/login`, getOptions(data))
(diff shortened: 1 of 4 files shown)


packages/login/src/modules/sagas.js
```diff
@@ -17,15 +17,30 @@ export const DEFAULT_TIMEOUT = 30
...
-export function doLoginRequest(data) {
-  return fetch(`${__BACKEND_URL__}/nice2/login`, getOptions(data))
+export function doRequest(url, options) {
+  return new Promise(resolve => {
+    fetch(url, options)
+      .then(resp => {
+        resp.json().then(json => resolve(json))
+      })
+      .catch(e => {
+        if (console && console.error) {
```
MethodenMann: does console.error work in IE? I just know that is not standard: https://developer.mozilla.org/es/docs/Web/API/Console/error
And if console.error is undefined, the error message will never be shown

rzueger: Yes. According to your link there is basic support since IE 8. And I checked it in IE 11 and Edge and it's available and seems to work.

PR #22, opened 2016-10-31
feat(tocco-test-util): add new package tocco-test-util

ottrob:
A new package which provides the most needed mocks for writing tests.


packages/tocco-test-util/README.md
```diff
@@ -0,0 +1,9 @@
+# Tocco-test-util
+This package contains the most needed mocks for tests.
```
MethodenMann: i wouldn't reduce this package to mocks. It could also contain some useful utils like helper classes / methods or generic test data. but t

PR #30, opened 2016-11-11
Introduce 'fetch-mock'

uwolfer:


packages/login/src/modules/passwordUpdate/password/sagas.js
```diff
@@ -27,61 +27,33 @@ function doRequest(data, username, action) {
...
-            }))
-          }
-        })
+  return doRequest(data, username, 'password-validation')
+    .then(resp => {
+      if (resp.ok) {
+        return resp.json().then(json => json)
+      } else {
+        return resp.json().then(() => ({
+          // validation request failed for some reason. we ignore that.
```
MethodenMann: i would move comment outside of object (above return statement)

PR #31, opened 2016-11-17
refactor: mock text resources

MethodenMann:
- add util function to mock textresource by passing an array of keys
- use in merge and login package


packages/tocco-util/src/dev/fetchMock/textResource.spec.js
```diff
@@ -0,0 +1,42 @@
+import textResourceFetchMock from './textResource'
+
+describe('tocco-util', () => {
+  describe('dev', () => {
+    describe('utilFetchMock', () => {
+      describe('textResourceFetchMock', () => {
+        it('should contain a matcher for textresources', () => {
+          let urlRegex
+
+          let fetchMockStuck = {
```
rzueger: - `const` instead of `let`?
- what does `fetchMockStuck` mean?

packages/tocco-util/src/dev/fetchMock/textResource.spec.js
```diff
@@ -0,0 +1,42 @@
...
+
+          textResourceFetchMock(fetchMockStuck, [])
+
+          expect(urlRegex.toString()).to.contains('/textresource')
+        })
+
+        it('should return an object with readable text resources', () => {
+          let fetchResponse
+
+          let fetchMockStuck = {
```
rzueger: see comment above

PR #36, opened 2016-11-22
Pr/entity browser/pagination component

MethodenMann:


packages/tocco-ui/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,154 @@
...
+    this.state = {
+      currentPage: 1,
+      totalPages: Math.ceil(props.totalRecords / props.recordsPerPage)
+    }
+
+    this.callOnPageChanged = _debounce(this.callOnPageChanged, 800)
+  }
+
+  onLastPage = () => this.state.currentPage >= this.state.totalPages
+  onFirstPage = () => this.state.currentPage <= 1
```
ottrob: this.state.currentPage === 0  or this.state.currentPage === 1? In the function `handleInputUpdate` the value will be set to 0 if the input is less than 0 so I guess the index of the first page has to be zero?

packages/tocco-ui/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,154 @@
...
+
+  handleToLast = () => {
+    if (!this.onLastPage()) {
+      this.updateCurrentPage(this.state.totalPages)
+    }
+  }
+
+  handleToFirst = () => {
+    if (!this.onFirstPage()) {
+      this.updateCurrentPage(1)
```
ottrob: here the first page is 1

packages/tocco-ui/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,154 @@
...
+    let value = parseInt(event.target.value)
+
+    if (value < 0) {
+      value = 0
+    }
+    if (value > this.state.totalPages) {
+      value = this.state.totalPages
+    }
+
+    this.updateCurrentPage(value)
```
ottrob: and here the value might be zero in case the value is lower than 0.

packages/tocco-ui/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,154 @@
...
+    return (
+      <div className="tocco-pagination">
+        <button
+          id="toFirstButton"
+          type="button"
+          className="btn"
+          onClick={this.handleToFirst}
+          disabled={this.onFirstPage()}
+        >
+          <span className="glyphicon glyphicon-fast-backward"/>
```
ottrob: Maybe use Tocco-UI-Button? Could already handle the glyphicon.

packages/tocco-ui/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * Amount of records that are shown on one page (is used to calculate numbers of pages)
+   */
+  recordsPerPage: React.PropTypes.number.isRequired,
+  /**
+   * Get called everytime the current page changes. Given the current number as first argument.
+   */
+  onPageChange: React.PropTypes.func,
+  /**
+   * If set to false (default true) a label, instead of a input field to enter page, is displayed.
```
ottrob: an input instead of a input

packages/tocco-ui/src/Pagination/styles.scss
```diff
@@ -0,0 +1,24 @@
+.tocco-pagination {
+  .total{
```
ottrob: missing space before curly bracket -> .total {, button {, #currentPage {

PR #39, opened 2016-11-28
Pr/entity browser/search component

ottrob:
Add new component `SearchBox`. The html structure of the component has been adopted from the bootstrap page (http://getbootstrap.com/components/#input-groups-buttons).

### DIFF
--- a/packages/tocco-ui/src/SearchBox/SearchBox.js
+++ b/packages/tocco-ui/src/SearchBox/SearchBox.js
@@ -11,7 +11,7 @@ class SearchBox extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      inputValue: ""
+      inputValue: ''
     }
 
     this.liveSearch = _debounce(this.liveSearch, props.debounce)
@@ -35,7 +35,7 @@ class SearchBox extends React.Component {
 
   keyDownHandler = evt => {
     if (this.props.liveSearch) {
-      this.liveSearch();
+      this.liveSearch()
     } else if (evt.key === 'Enter') {
       this.explicitSearch()
     }
--- a/packages/tocco-ui/src/SearchBox/SearchBox.js
+++ b/packages/tocco-ui/src/SearchBox/SearchBox.js
@@ -11,26 +11,33 @@ class SearchBox extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      inputValue: ''
+      inputValue: ""
     }
-    this.search = _debounce(this.search, props.debounce)
+
+    this.liveSearch = _debounce(this.liveSearch, props.debounce)
   }
 
   updateValue = evt => {
-    this.setState({'inputValue': evt.target.value})
+    this.setState({inputValue: evt.target.value})
   }
 
-  search = () => {
+  liveSearch = () => {
     const value = this.state.inputValue
-    if (this.props.liveSearch || (value && value.length >= this.props.minInputLength)) {
+    if (value && value.length >= this.props.minInputLength) {
       this.props.onSearch(value)
     }
   }
 
+  explicitSearch = () => {
+    const value = this.state.inputValue
+    this.props.onSearch(value)
+  }
+
   keyDownHandler = evt => {
-    const KEY_ENTER_CODE = 13
-    if (this.props.liveSearch || evt.keyCode === KEY_ENTER_CODE) {
-      this.search()
+    if (this.props.liveSearch) {
+      this.liveSearch();
+    } else if (evt.key === 'Enter') {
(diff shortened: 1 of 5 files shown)


packages/tocco-ui/src/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,99 @@
...
+      this.search()
+    }
+  }
+
+  render() {
+    return (
+      <div
+        className="tocco-searchbox row">
+        <div
+          className="col-lg-6">
```
MethodenMann: sure we want to set this class? i think a generic component should not be limited in this way

packages/tocco-ui/src/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,99 @@
...
+  constructor(props) {
+    super(props)
+    this.state = {
+      inputValue: ''
+    }
+    this.search = _debounce(this.search, props.debounce)
+  }
+
+  updateValue = evt => {
+    this.setState({'inputValue': evt.target.value})
```
MethodenMann: single quotation mark not necessary

packages/tocco-ui/src/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Placeholder of the input field.
+   */
+  placeholder: React.PropTypes.string,
+  /**
+   * If true, the `onSearch` function will be triggered on the fly.
+   */
+  liveSearch: React.PropTypes.bool,
+  /**
+   * Amount of milli seconds before the search starts. The default is set to `200`.
```
MethodenMann: "before the search starts." is not really correct. Its more like a delay of miliseconds that have to elapse since the last invocation.

packages/tocco-ui/src/SearchBox/SearchBox.spec.js
```diff
@@ -0,0 +1,140 @@
+import React from 'react'
+import SearchBox from './SearchBox'
+import Button from '../Button'
+
+import {shallow} from 'enzyme'
+
+describe('tocco-ui', function() {
+  describe('SearchBox', function() {
+    const DEFAULT_DEBOUNCE = 200
```
MethodenMann: with this value e unit test takes at least 200ms? i would change it to 50

ottrob: This is the default value of the SearchBox component. I would have to set it manually lower than 200 but then the default behaviour cannot be tested.

packages/tocco-ui/src/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,99 @@
...
+      inputValue: ''
+    }
+    this.search = _debounce(this.search, props.debounce)
+  }
+
+  updateValue = evt => {
+    this.setState({'inputValue': evt.target.value})
+  }
+
+  search = () => {
```
ottrob: I discussed this behaviour also with [uwolfer]. I think this is the better approach. Otherwise it could be confusing. With this approach it is always possible to set the minInputLength to `1` or the debounce value to `0`. If I would implement your suggested approach, these two properties would have no effect.

packages/tocco-ui/src/SearchBox/example.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import SearchBox from './'
+// real-import:import {SearchBox} from 'tocco-ui'
+
+export default () => {
+  return (
+    <div>
+      {/* start example */}
+      <SearchBox
+        placeholder="Live search..."
```
rzueger: i18n

packages/tocco-ui/src/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,107 @@
...
+              onKeyDown={this.keyDownHandler}
+              placeholder={this.props.placeholder}
+              value={this.state.inputValue}
+              onChange={this.updateValue}
+             />
+            <span
+              className="input-group-btn">
+              <Button
+                type="button"
+                className="btn"
```
MethodenMann: default class of ToccoUI.Button

packages/tocco-ui/src/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,107 @@
...
+                onClick={this.explicitSearch}
+                icon="glyphicon-search"
+                disabled={this.props.liveSearch}
```
MethodenMann: Why is the button disabled when LiveSearch is on?

PR #46, opened 2016-12-01
Pr/entity browser/search form

ottrob:
Integrate search function


packages/entity-browser/src/components/EntityBrowser/EntityBrowser.js
```diff
@@ -24,14 +25,35 @@ export class EntityBrowser extends React.Component {
     return <span>_</span>
   }
 
+  renderSearchForm(formDefinition) {
```
MethodenMann: I think SearchForm should be a component

packages/entity-browser/src/components/EntityBrowser/EntityBrowser.js
```diff
@@ -24,14 +25,35 @@ export class EntityBrowser extends React.Component {
...
+    if (formDefinition && formDefinition.length > 0) {
+      return (
+        <div className="search-form">
+          <form>
+            {formDefinition.map(definition => {
+              return (
```
MethodenMann: can be shortened :    definition => <InputFactory...

packages/entity-browser/src/components/EntityBrowser/InputFactory/InputFactory.js
```diff
@@ -0,0 +1,28 @@
...
+import TextFieldInputHandlerContainer from './../../../containers/TextFieldInputHandlerContainer'
+
+class InputFactory extends React.Component {
+
+  showField = () => {
+    return this.props.fieldDefinition.displayType !== 'HIDDEN'
+  }
+
+  getTypeMap = () => ({
+    'ch.tocco.nice2.model.form.components.simple.TextField': TextFieldInputHandlerContainer
```
MethodenMann: Create a container for FieldInputHandler seams a bit inconvenient for me. why no keep it a simple component and pass the properties down?

packages/entity-browser/src/components/EntityBrowser/EntityBrowser.js
```diff
@@ -24,14 +25,35 @@ export class EntityBrowser extends React.Component {
...
+                <InputFactory key={definition.name} fieldDefinition={definition}/>
```
MethodenMann: property key is not used?

packages/entity-browser/src/modules/entityBrowser/reducer.js
```diff
@@ -14,14 +14,24 @@ const clearRecordStore = state => ({
   recordStore: {}
 })
 
+const setSearchTerm = (state, {payload}) => {
+  const term = payload.searchTerm
```
MethodenMann: why do you pass an object containing a name and value and only set the value in the state? i would either save the whole object to the state (and then use the singleTransferRecuder) is name attribute is needed or only pass value (and use singleTransferReducer)

packages/entity-browser/src/components/EntityBrowser/EntityBrowser.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react'
 import * as ToccoUI from 'tocco-ui'
 import './styles.scss'
+import {InputFactory} from './InputFactory'
 
 export class EntityBrowser extends React.Component {
```
rzueger: We should use stateless functional components instead of classes wherever possible.

See also https://medium.com/@housecor/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc#.c4yqbjs15

packages/entity-browser/src/components/EntityBrowser/SearchForm/SearchForm.js
```diff
@@ -0,0 +1,23 @@
...
+          setSearchTerm={setSearchTerm}
+        />)}
+      </form>
+    </div>
+  )
+}
+
+SearchForm.propTypes = {
+  formDefinition: React.PropTypes.array.isRequired,
+  setSearchTerm: React.PropTypes.func
```
MethodenMann: i thing this should also be "isRequired"

packages/entity-browser/src/components/EntityBrowser/InputFactory/InputFactory.js
```diff
@@ -1,28 +1,28 @@
...
-      return React.createElement(typeHandler, {'fieldDefinition': this.props.fieldDefinition})
-    }
-    return null
+  const showField = fieldDefinition => {
+    return fieldDefinition.displayType !== 'HIDDEN' && getTypeMap().hasOwnProperty(fieldDefinition.type)
   }
+
+  return (
+    showField(fieldDefinition)
+    && React.createElement(
```
MethodenMann: if (showField(fieldDefinition)){
return   getTypeMap()[fieldDefinition.type], {'fieldDefinition': fieldDefinition, 'setSearchTerm': setSearchTerm}
 }
return <span/>


i would prefer this notation. its easier to read and fallback case is explicitly handled

packages/entity-browser/src/modules/entityBrowser/reducer.js
ottrob: I don't think the name should be saved in the redux state. The name only can be used to set/change the search term directly in the reducer. There is no need for the name in the state.

PR #48, opened 2016-12-05
feat(entity-browser): enhance input parameters

MethodenMann:
- add limit and formBase as new input parameter
- generic input validation (can eventually be extracted later for other packages)

### DIFF
--- a/packages/entity-browser/src/appFactory.js
+++ b/packages/entity-browser/src/appFactory.js
@@ -79,10 +79,8 @@ const validateInput = input => {
   inputsFields.forEach(f => {
     if (input[f.name]) {
       dispatches.push(f.action(input[f.name]))
-    } else {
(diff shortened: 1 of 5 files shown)


packages/entity-browser/src/appFactory.js
```diff
@@ -71,3 +72,37 @@ const factory = (id, input = {}, externalEvents, publicPath, resourcePrefix, red
...
 }
+
+const validateInput = input => {
+  const dispatches = []
+
+  inputsFields.forEach(f => {
+    if (input[f.name]) {
+      dispatches.push(f.action(input[f.name]))
+    } else {
+      if (f.mandatory) {
```
rzueger: could be merge to

```
...
} else if (f.mandatory) {
   ...
}

PR #49, opened 2016-12-05
Pr/entity browser/search form fix

ottrob:


packages/entity-browser/src/components/EntityBrowser/InputFactory/TypeHandler/TextFieldInputHandler.js
```diff
@@ -3,7 +3,8 @@ import * as ToccoUI from 'tocco-ui'
 
 const TextFieldInputHandler = props => {
   return (
-    <ToccoUI.SearchBox
+    props.fieldDefinition.name === 'txtFulltext'
```
MethodenMann: if (....) { 
 return <SearchBox>
}

is clearer in my opinion

PR #65, opened 2016-12-20
feat(tocco-ui): add EditableValue component

MethodenMann:
- basic implementation for type string and multi-select
- tests
- integrate in showcase


packages/tocco-ui/src/EditableValue/EditableValue.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+EditableValue.propTypes = {
+  /**
+   * Type of value. e.g. 'string'
+   */
+  type: React.PropTypes.oneOf([
+    'multi-select',
+    'string'
+  ]).isRequired,
```
rzueger: This list should be coming from a source like the map in typeEditorFactory. It might get cumbersome to keep this list up to date at all times here.

PR #68, opened 2016-12-21
Pr/entity browser/search form

MethodenMann:


packages/entity-browser/src/appFactory.js
```diff
@@ -56,7 +57,10 @@ const factory = (id, input = {}, externalEvents, publicPath, resourcePrefix, red
       <Provider store={store}>
         <LoadMask promises={[initIntlPromise]}>
           <IntlProvider>
-            <EntityBrowserContainer/>
+            <div>
+              <SearchFormContainer/>
+              <EntityBrowserContainer/>
+            </div>
```
MethodenMann: yes its required. IntlProvider allows only a single child.

PR #73, opened 2016-12-30
run tests on SauceLabs

uwolfer:


packages/tocco-ui/src/FormattedValue/typeFormatters/DurationFormatter.spec.js
```diff
@@ -19,7 +19,7 @@ describe('tocco-ui', function() {
 
         const wrapper = mount(<IntlProvider locale="en"><DurationFormatter
           value={durationMilliseconds}/></IntlProvider>)
-        expect(wrapper.text()).to.equal('01:30:01')
+        expect(wrapper.text().replace(/\u200E/g, '')).to.equal('01:30:01')
```
MethodenMann: by reading this test its neither clear why you do the replace in the first place or what exactly gets replaced. i would move the regex into a variable.

PR #75, opened 2017-01-03
chore: Include static type checker `flow`

rzueger:
- Type checking can be enabled for a file by adding `// @flow` at the top
- Type annotations are removed from output code automatically
  (see babel plugin `transform-flow-strip-types`)
- Prop type definitions for JSX components are converted to React prop types
  automatically (so that there are still runtime errors if wrong props are
  passed to a component)
  (see babel plugin `flow-react-proptypes`)
- Type checking is included as ESLint plugin (`flowtype-errors`).
  However, I encountered one type error which wasn't reported by the plugin:
  I created a generic PayloadAction<P> (exact) type with a string property
  `type` and the generic `payload` property. Then I imported this type in the
  reducers and action creator files. Then I created an action object of type
  `PayloadAction<MyPayloadType>` without a payload property. The ESLint plugin
  didn't complain about the missing `payload` property, but `flow` did (when
  `flow` was run using `npm run flow`). That's why `npm run flow` is added
  to the Travis config, even though `npm run lint` should already have reported
  all errors.
- see flowtype.org


.eslintrc
```diff
@@ -28,7 +32,8 @@
     "describe": false,
     "sinon": false,
     "before": false,
-    "beforeEach": false
+    "beforeEach": false,
+    "Generator": true
```
MethodenMann: where do we need this global?

PR #79, opened 2017-01-04
chore(login): Add first flow type annotations in login package

rzueger:
At least one file of each of the following file types is annotated:
- React component
- actions creators
- reducer
- sagas


rzueger:
we decided not to use flowtype atm

PR #83, opened 2017-01-05
Pr/entity browser/detail state

MethodenMann:


packages/entity-browser/src/components/ListView/ListView.js
```diff
@@ -0,0 +1,71 @@
...
+          recordsPerPage={props.limit}
+          onPageChange={this.onPageChange}
+          currentPage={props.currentPage}
+        />
+        <ToccoUI.Button
+          onClick={props.refresh}
+          icon="glyphicon-refresh"
+          className="refresh-button"
+        >
+          Reset
```
rzueger: l10n

packages/entity-browser/src/components/ListView/ListView.js
```diff
@@ -0,0 +1,71 @@
...
+        </ToccoUI.Button>
+      </div>
+    )
+  }
+}
+
+ListView.propTypes = {
+  changePage: React.PropTypes.func.isRequired,
+  records: React.PropTypes.array.isRequired,
+  orderBy: React.PropTypes.object.isRequired,
```
rzueger: can we define a shape for the `orderBy` prop?

packages/entity-browser/src/components/ListView/ListView.js
```diff
@@ -0,0 +1,71 @@
...
+  currentPage: React.PropTypes.number,
+  limit: React.PropTypes.number,
+  columnDefinitions: React.PropTypes.array,
```
rzueger: can we define a shape for the `columnDefinition` prop?

packages/entity-browser/src/components/EntityBrowser/EntityBrowser.js
```diff
@@ -1,77 +1,21 @@
...
-        </span>
-      )
-    }
-  }
+import './styles.scss'
 
-  render() {
-    const props = this.props
+export const EntityBrowser = props => {
+  props.initialize()
```
rzueger: Is it a good idea to call `initialize()` in the render function? Maybe componentWillMount/componentDidMount should be used?

packages/entity-browser/src/components/ListView/ListView.js
```diff
@@ -0,0 +1,87 @@
+import React from 'react'
+import {intlShape} from 'react-intl'
+import * as ToccoUI from 'tocco-ui'
+import './styles.scss'
+
+export class ListView extends React.Component {
```
rzueger: doesn't need to be a class, does it?

PR #85, opened 2017-01-16
chore: introduce plop

MethodenMann:
- add generator for react component generation
- add generator for redux action generation

### DIFF
--- a/plopfile.js
+++ b/plopfile.js
@@ -1,0 +1,123 @@
+const getAllPackages = require('./bin/packages').getAllPackages
+
+module.exports = function(plop) {
+  plop.addHelper('ifIn', (val, arr, options) => {
+    if (arr.includes(val)) {
+      return options.fn(this)
+    }
+    return options.inverse(this)
+  })
+
+  const packagePrompt = {
+    type: 'rawlist',
+    name: 'package',
+    message: 'Package Name',
+    choices: getAllPackages()
+  }
+
+  plop.setGenerator('Component', {
+    description: 'Create a React Component',
+    prompts: [
+      packagePrompt,
+      {
+        type: 'input',
+        name: 'componentName',
+        message: 'Component Name',
+        validate: value => {
+          if ((/.+/).test(value)) {
+            return true
+          }
+          return 'name is required'
+        }
+      }, {
+        type: 'confirm',
+        name: 'stateless',
+        message: 'Stateless component?'
+      }, {
+        type: 'checkbox',
+        name: 'features',
+        message: 'What features would you like?',
+        choices: [
+          {name: 'Tests', value: 'tests', checked: true},
+          {name: 'Tocco-Ui', value: 'toccoui'}
+        ]
+      }],
+    actions: data => {
+      let actions = []
+
+      const baseTemplateFolder = 'templates/component'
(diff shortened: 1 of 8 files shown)


templates/component/stateful.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+{{#ifIn 'toccoui' features}}
+import * as ToccoUi from 'tocco-ui'
+{{/ifIn}}
+
+class {{pascalCase componentName}} extends React.Component {
+  constructor(props) {
+    super()
```
rzueger: `props` should be forwarded to `super()`, shouldn't it?

plopfile.js
```diff
@@ -0,0 +1,123 @@
...
+
+      actions.push({
+        type: 'add',
+        path: `${targetFolder}/{{pascalCase componentName}}.js`,
+        templateFile: data.stateless ? `${baseTemplateFolder}/stateless.js` : `${baseTemplateFolder}/stateful.js`
+      })
+
+      actions.push({
+        type: 'add',
+        path: `${targetFolder}//index.js`,
```
uwolfer: doubled slashes?

templates/component/stateful.js
```diff
@@ -0,0 +1,26 @@
...
+    return (
+      <div>
+        <h1>{{titleCase componentName}}</h1>
+      </div>
+    )
+  }
+}
+
+{{pascalCase componentName}}.propTypes = {
+  val: React.PropTypes.func.isRequired
```
rzueger: I personally would leave the prop types empty. I don't like too much stuff in templates which has to be removed or changed anyway.

plopfile.js
```diff
@@ -0,0 +1,123 @@
...
+          return 'name is required'
+        }
+      }, {
+        type: 'confirm',
+        name: 'stateless',
+        message: 'Stateless component?'
+      }, {
+        type: 'checkbox',
+        name: 'features',
+        message: 'What features would you like?',
```
uwolfer: "Which features would you like to use?"

templates/component/tests.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import {EntityBrowser} from './EntityBrowser'
+import {shallow} from 'enzyme'
+
+const EMPTY_FUNC = () => {}
```
rzueger: I personally would remove this function. Seems quite useless to me.

plopfile.js
```diff
@@ -0,0 +1,123 @@
...
+  })
+
+  const packagePrompt = {
+    type: 'rawlist',
+    name: 'package',
+    message: 'Package Name',
+    choices: getAllPackages()
+  }
+
+  plop.setGenerator('Component', {
```
rzueger: IMO each generator should be moved to its own source file as soon as possible.

templates/component/tests.js
MethodenMann: To avoid prop validation errors during a test run you often have to pass such empty functions. i added it in the template to have a consistency in the naming and ussage. If you dont need this function you are forced to delete it because eslint does not allow unussed vars.

plopfile.js
MethodenMann: yes agree

PR #86, opened 2017-01-17
pr/entity-browser/validation

MethodenMann:


packages/entity-browser/src/components/DetailView/LabeledField.js
```diff
@@ -2,6 +2,24 @@ import React from 'react'
 import _clone from 'lodash/clone'
 import * as ToccoUi from 'tocco-ui'
 
+const ErrorList = props => {
+  const errorValues = Object.values(props.errors)
+
+  return (
+    <ul className="error-list">
+      { errorValues.map((value, idx) => (
```
rzueger: (I would remove the space after `{`)

packages/entity-browser/src/components/DetailView/LabeledField.js
```diff
@@ -2,6 +2,24 @@ import React from 'react'
 import _clone from 'lodash/clone'
 import * as ToccoUi from 'tocco-ui'
 
+const ErrorList = props => {
+  const errorValues = Object.values(props.errors)
```
rzueger: This leads to an error if there is no `error` prop.

Possibilities:

- Declare prop as required
- Declare default prop `{}`
- Check here if errors are set

packages/entity-browser/src/components/DetailView/LabeledField.js
```diff
@@ -12,10 +30,10 @@ const LabeledField = ({input, label, type, meta: {touched, error, warning}}) =>
     <div className="form-group">
       <label className="control-label col-sm-5">{label}:</label>
       <div className="col-sm-7">
-        <p className="form-control-static">
+        <div className="form-control-static">
           <ToccoUi.EditableValue value={input.value} onChange={input.onChange} type={type} events={events}/>
-          {touched && ((error && <span>{error}</span>) || (warning && <span>{warning}</span>))}
-        </p>
+          {touched && (error && <ErrorList errors={error}/>)}
```
rzueger: Parentheses are not required and I would remove them to increase readability

PR #87, opened 2017-01-18
entity-browser refactoring

MethodenMann:


packages/entity-browser/src/util/api/entities.js
```diff
@@ -0,0 +1,105 @@
...
+
+export const entitiesListTransformer = json => {
+  return json.data.map(entity => {
+    const result = {
+      id: entity.key,
+      values: {}
+    }
+
+    const paths = entity.paths
+    for (let path in paths) {
```
ottrob: `const path` instead of `let path`?

packages/entity-browser/src/modules/listView/sagas.js
```diff
@@ -55,7 +57,10 @@ export function* fetchEntitiesAndAddToStore(page) {
   if (!entityStore[page]) {
     const searchInputs = yield call(getSearchInputs)
 
-    const entities = yield call(api.fetchEntities, entityName, page, orderBy, limit, columnDefinition, searchInputs)
+    const fields = _union(...columnDefinition.map(field => (field.value)))
```
MethodenMann: const fields = columnDefinition.map(field => (field.value))
would get you an array of arrays with the fields. but i have rewritten it a bit and removed the duplicate check since REST API should handle that well.

packages/entity-browser/src/dev/fetchMockHelpers.js
```diff
@@ -0,0 +1,72 @@
...
+
+const allEntities = createUsers(1003)
+
+export const createValidateResponse = (url, opts) => {
+  const fields = {}
+  const entity = opts.body
+  const requiredFields = ['firstname', 'lastname']
+  requiredFields.forEach(requiredField => {
+    if (!entity.paths[requiredField].value.value
+      || entity.paths[requiredField].value.value === '') {
```
rzueger: `... === ''` already covered by first condition (`!...`)

packages/entity-browser/src/modules/detailView/reducer.spec.js
```diff
@@ -0,0 +1,32 @@
...
+
+const EXPECTED_INITIAL_STATE = {
+  entityName: '',
+  formDefinition: [],
+  entity: {}
+}
+
+describe('entity-browser', () => {
+  describe('modules', () => {
+    describe('listView', () => {
```
rzueger: `'detailView'` ?

packages/entity-browser/src/modules/detailView/sagas.spec.js
```diff
@@ -0,0 +1,58 @@
...
+            expect(generator.next().done).to.be.true
+          })
+        })
+
+        describe('initialize saga', () => {
+          it('should set entityname and load detail form definition', () => {
+            const formBase = 'Base_form'
+            const entityName = 'User'
+
+            const gen = sagas.initialize({payload: {entityName, formBase}})
```
rzueger: why aren't you using the action creators to test the sagas (same below)?

packages/entity-browser/src/util/api/forms.js
```diff
@@ -0,0 +1,59 @@
...
+export const columnDefinitionTransformer = json => {
+  const {form} = json
+
+  const columns = form.children.find(child => child.name === 'table')
+    .children.filter(column => column.displayType !== 'HIDDEN')
+
+  return columns.map(c => ({
+    label: c.label,
+    value: c.children
+      .filter(child =>
```
rzueger: is it possible to improve readability here (indentation)?

packages/entity-browser/src/util/reduxForms.js
```diff
@@ -1,7 +1,33 @@
...
+
+  return entity
+}
+
+export const entityToFormValues = entity => {
+  if (!entity || !entity.paths) return {}
+  const result = {}
+  const paths = entity.paths
+  Object.keys(entity.paths).forEach(key => {
+    if (paths[key].value !== null) {
```
rzueger: what about `undefined`? Maybe use `!= null` to cover `undefined` as well?

packages/entity-browser/src/util/api/forms.js
MethodenMann: yes

PR #90, opened 2017-01-23
pr/entity-browser/date

MethodenMann:


packages/tocco-ui-showcase/src/components/ShowCaseApp.js
```diff
@@ -1,29 +1,53 @@
...
+
+    addLocaleData([...de, ...en, ...fr, ...it])
+  }
+
   static propTypes = {
     componentsTree: React.PropTypes.array.isRequired
   }
 
+  localeChange(locale) {
+    console.log('Change locale to: ', locale)
```
ottrob: remove console.log?

MethodenMann: its only showcase app code. i left that there on purpose so you see the exact locale name in console.

PR #92, opened 2017-01-23
pr/entity-browser/datetime

MethodenMann:


packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -0,0 +1,79 @@
...
+      </span>
+    )
+  }
+}
+
+DateAbstract.propTypes = {
+  intl: intlShape.isRequired,
+  onChange: React.PropTypes.func,
+  value: React.PropTypes.string,
+  options: React.PropTypes.object
```
ottrob: maybe with React.PropTypes.shape({...})?

MethodenMann: Not really possible. Options can have any attributes. I just add them to the options of flatpickr and dont want to restrict them in any way.

PR #94, opened 2017-01-24
Pr/entity browser/async store load

ottrob:


packages/entity-browser/src/components/DetailView/DetailForm.js
```diff
@@ -80,7 +84,19 @@ DetailForm.propTypes = {
   submitting: React.PropTypes.bool,
   submitFailed: React.PropTypes.bool,
   submitSucceeded: React.PropTypes.bool,
-  anyTouched: React.PropTypes.bool
+  anyTouched: React.PropTypes.bool,
+  loadRelationEntities: React.PropTypes.func,
+  stores: React.PropTypes.shape({
```
MethodenMann: name 'store' does not really explain what exactly is stored. Its only for relatedEntities, right? I would name id like that

packages/entity-browser/src/modules/detailView/actions.js
```diff
@@ -44,3 +48,33 @@ export const setEntity = entity => ({
 export const submitForm = () => ({
   type: SUBMIT_FORM
 })
+
+export const loadRelationEntities = entityName => ({
```
MethodenMann: its only for one => loadRelationEntity

packages/entity-browser/src/modules/detailView/sagas.js
```diff
@@ -43,6 +47,22 @@ export function* loadEntity({payload}) {
   yield put(actions.setEntity(entity))
 
   const formValues = yield call(entityToFormValues, entity)
+
+  // initialize store for fields of the types `entity` and `entity-list`
+  const keys = Object.keys(entity.paths)
```
MethodenMann: Refactoring suggestion: 
- Create Util method "getRelatedEntitesStoreInitialValues" (or something like that) that returns an object with this structure:
 { key: fieldstore, key2: fieldstrore}
- In saga iterate over key and call put


This way nice2 specific logic is outside of sages and not mixed

packages/entity-browser/src/modules/detailView/sagas.js
```diff
@@ -21,14 +21,18 @@ export default function* sagas() {
...
+    fork(takeEvery, actions.SUBMIT_FORM, submitForm),
+    fork(takeEvery, actions.LOAD_RELATION_ENTITIES, loadRelationEntities)
   ]
 }
 
 export function* initialize({payload}) {
   const {entityName, formBase} = payload
 
   yield put(actions.setEntityName(entityName))
+  const entityModel = yield call(fetchModel, entityName)
```
MethodenMann: Rightnow the model gets loaded in SearchForm and DetailForm, i would refactor that out to root module so its only gets called once.

packages/entity-browser/src/components/DetailView/DetailForm.js
```diff
@@ -33,16 +33,20 @@ export const DetailForm = props => {
...
-          } else if (entityField.type === 'entity-list') {
+            const store = props.stores[field.name].data
+            fieldProps.options = {store}
+          } else if (entityField.type === 'entity-list' && props.stores[field.name] !== undefined) {
             fieldProps.type = 'multi-select'
-            const store = entityField.value.map(e => ({value: e.key, label: e.display}))
+            const store = props.stores[field.name].data
             fieldProps.options = {store}
+          } else {
+            fieldProps.options = {store: []}
```
MethodenMann: i get why you have this `else` but its not really a default fallback (to an unknown type). better remove it and change line 38 / 42 to this:

  const store = props.stores[field.name] ? props.stores[field.name].data : []

packages/entity-browser/src/components/DetailView/DetailView.js
```diff
@@ -13,6 +13,8 @@ const DetailView = props => {
         submitForm={props.submitForm}
         formDefinition={props.formDefinition}
         entity={props.entity}
+        loadRelationEntities={props.loadRelationEntities}
```
MethodenMann: wouldn't "lazyLoadRelationEntites" be a better name?

packages/entity-browser/src/dev/rest-responses/form_user_detail.json
```diff
@@ -131,7 +131,7 @@
                 "displayType": "READONLY",
                 "children": [
                   {
-                    "name": "relUser_code",
+                    "name": "relUser_code1",
```
MethodenMann: why? :)

packages/entity-browser/src/modules/detailView/actions.js
ottrob: Nope this function loads all possible Entities for a specific relation. For ex.: for `relGender` it loads Gender(male) and Gender(female). This action gets fired if a select box gets focused.

packages/entity-browser/src/dev/rest-responses/form_user_detail.json
ottrob: Because I created a json response mock for the User_code1 entity to load the store of the select box :) I also could have changed the response mock instead of this form definition. But its consistent now.

packages/entity-browser/src/components/DetailView/DetailForm.js
ottrob: renamed it to selectBoxStores

packages/entity-browser/src/components/DetailView/DetailView.js
ottrob: I think the name is ok because the function itself just loads the related entities. In the DetailForm this function is passed to the onFocus function which gets called if someone clicks on the field.

PR #96, opened 2017-01-26
pr/entity-browser/rest-save

MethodenMann:


packages/entity-browser/src/util/reduxForms.js
```diff
@@ -34,6 +47,12 @@ export const entityToFormValues = entity => {
   return result
 }
 
+export const getDirtyFields = (initialValues, values) => {
+  return _reduce(initialValues, function(result, value, key) {
+    return _isEqual(value, values[key]) ? result : result.concat(key)
+  }, [])
+}
```
rzueger: IMO, this would look a bit nicer with arrow functions:

```
export const getDirtyFields = (initialValues, values) =>
  _reduce(
    initialValues,
    (result, value, key) => _isEqual(value, vaules[key]) ? result : result.concat(key)
    []
  )

File: packages/entity-browser/src/util/reduxForms.spec.js [hunk 2]
```diff
@@ -29,21 +28,45 @@ describe('entity-browser', () => {
...
-                  value: 'griffin'
+              firstname: 'peter',
+              gender: {key: '2'}
+            }
+          }
+
+          expect(result).to.eql(expectedEntity)
+        })
+
+        it('should ignore no dirty fields', () => {
```
rzueger: "should ignore pristine fields"

packages/entity-browser/src/util/reduxForms.js
```diff
@@ -1,17 +1,30 @@
...
   Object.keys(values).forEach(key => {
-    if (key !== wholeEntityField) {
-      entity.paths[key].value.value = values[key]
+    if (!ignoreField(key)) {
+      const type = entity.paths[key].type
+      if (type === 'field') {
+        result.paths[key] = values[key]
+      } else if (type === 'entity') {
+        result.paths[key] = {key: values[key]}
+      } else if (type === 'entity-list') {
```
rzueger: this will probably change. but we can have a look at it again as soon as the backend is ready.

PR #100, opened 2017-01-30
Reduce bundle size

uwolfer:


build/webpack.config.js
```diff
@@ -22,7 +22,8 @@ const webpackConfig = {
   resolve: {
     modules: [
       path.resolve(paths.client(), packageDir, 'src'),
-      'node_modules'
+      'node_modules',
```
MethodenMann: Is there a difference to:
path.resolve(paths.client(), packageDir, 'node_modules')

packages/entity-browser/package.json
```diff
@@ -4,10 +4,10 @@
   "description": "Lists entities and allows to sort and search them",
   "main": "dist/index.js",
   "scripts": {
-    "clean": "rimraf dist",
+    "clean": "../../node_modules/rimraf/bin.js dist",
     "compile": "babel-node ../../bin/compile",
```
MethodenMann: is this script necessary? if yes, where is the package name defined? compile script throws an error without a packagename provided

build/webpack.config.js
uwolfer: I assume it's the same.

packages/entity-browser/package.json
uwolfer: Still required.

PR #101, opened 2017-01-30
pr/entity-browser/sync-validation

MethodenMann:


packages/entity-browser/src/components/DetailView/DetailForm.js
```diff
@@ -89,12 +110,13 @@ DetailForm.propTypes = {
         })
       )
     })
-  })
+  }),
+  formSyncErrors: React.PropTypes.object,
```
rzueger: can we define the shape of `formSyncErrors`?

packages/entity-browser/src/components/DetailView/DetailView.js
```diff
@@ -1,33 +1,50 @@
...
+        />
+      </div>
+    )
+  }
 }
 
 export default DetailView
 
 DetailView.propTypes = {
+  syncErrors: React.PropTypes.object,
```
rzueger: this prop seems to be unused (?)

packages/entity-browser/src/util/syncValidation.js
```diff
@@ -0,0 +1,68 @@
+import _forOwn from 'lodash/forOwn'
+import _isEmpty from 'lodash/isEmpty'
+
+export default entityModel =>
+  values => {
```
rzueger: could be shortened to

```
values => valueValidator(...)

File: packages/entity-browser/src/util/syncValidation.js [hunk 4]
```diff
@@ -0,0 +1,68 @@
...
+      if (validatorValue) {
+        addErrors(key, validatorDefinition.validator(value, validatorValue))
+      }
+    })
+  })
+
+  return errors
+}
+
+const mandatoryValidator = (value, isMandatory) => {
```
rzueger: I think the type of the value should be taken into account here. For example, `0` might be a valid value for mandatory integer fields, but `!0` evaluates to `true`.

packages/entity-browser/src/components/DetailView/DetailForm.js
```diff
@@ -39,17 +37,38 @@ export const DetailForm = props => {
...
     return result
   }
 
   const handleSubmit = e => {
-    props.submitForm()
     e.preventDefault()
+    if (props.valid) {
+      props.submitForm()
+    } else {
+      if (props.formSyncErrors) {
```
rzueger: could be shortenend to `else if`

packages/entity-browser/src/components/DetailView/DetailView.js
```diff
@@ -28,6 +42,8 @@ DetailView.propTypes = {
   formDefinition: React.PropTypes.shape({
     children: React.PropTypes.array
   }).isRequired,
+  entityModel: React.PropTypes.object.isRequired,
+  formSyncErrors: React.PropTypes.object,
```
rzueger: shape?

packages/tocco-ui/src/EditableValue/typeEditors/NumberEdit.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+
+const NumberEdit = props => {
+  const convertStrintToNumber = stringValue => (stringValue ? Number(stringValue) : null)
```
rzueger: - `convertStringToNumber` (typo)
- can be defined outside of the render function
- Maybe check if string matches number pattern before trying to parse it? (input[type=number] for example still not supported in Opera Mini (global usage 4.36%; according to http://caniuse.com/#feat=input-number))

packages/entity-browser/src/components/DetailView/DetailForm.js
```diff
@@ -39,17 +39,36 @@ export const DetailForm = props => {
           fieldProps.options = {store}
         }
 
-        result.push(<div key={i} onFocus={() => props.loadRelationEntities(field.name)}>
-          <Field name={field.name} key={i} label={field.label} component={LabeledField} {...fieldProps}/></div>)
+        const isMandatoryField = fieldName => (
+          props.entityModel[fieldName]
+          && props.entityModel[fieldName].validate
+          && props.entityModel[fieldName].validate.mandatory
```
ottrob: I saw that lodash already has a function for this kind of problem.
https://lodash.com/docs/4.17.4#get
 Maybe this could be rewritten to something like this: _has(props.entityModel, fieldName + 'validate.mandatory')

PR #110, opened 2017-02-07
 add error handler which logs errors to server

uwolfer:


packages/tocco-util/src/dev/fetchMock/log.js
```diff
@@ -0,0 +1,6 @@
+export default fetchMock => {
+  fetchMock.post(new RegExp('^.*?/nice2/log'), (url, opts) => {
+    console.log('Logging to log-servlet...', opts)
```
MethodenMann: console.log('fetchMock: Logging to log-servlet', ops)

I would prefix the message to clarify where the message is coming from.

PR #113, opened 2017-02-13
Merge Update

MethodenMann:


packages/merge/src/utils/input.js
```diff
@@ -0,0 +1,33 @@
...
+    retrieveModel(input.model),
+    changeTargetEntity(input.entities[0].pk),
+    clearRelationMany(),
+    retrieveEditOptions(input.sourceEntitiesEditOptions),
+    ...dispatchToManySelections(input)
+  ]
+)
+
+const dispatchToManySelections = input => {
+  const result = []
```
rzueger: what's this `result` array for?

packages/tocco-util/src/appFactory/appFactory.js
```diff
@@ -0,0 +1,115 @@
...
+
+    return {
+      renderComponent: component,
+      store,
+      methods: {
+        setLocale: locale => intl.setLocale(store, name, locale)
+      }
+    }
+  } catch (e) {
+    console.log('Error creating react application: ', e)
```
rzueger: maybe we should add a logging utility which prevents from null pointers if console is not defined

packages/merge/src/utils/input.js
```diff
@@ -0,0 +1,33 @@
...
+  input.model.relations
+    .filter(relation => relation.toMany)
+    .map(relation => {
+      input.entities.map(entity => {
+        const userRelation = entity.relations[relation.name]
+        userRelation.values.map(userRelationValue => {
+          if (userRelationValue.checked) {
```
rzueger: Is it considered good practice to return `undefined` implicitly in functions like this (there is no `return` if the value is not checked)? (IMO, it's not)

packages/tocco-util/src/appFactory/appFactory.js
```diff
@@ -37,8 +39,14 @@ export const createApp = (name, content, reducers, sagas, input, events, actions
...
       }
     }
-  } catch (e) {
-    console.log('Error creating react application: ', e)
+  } catch (error) {
+    try {
+      if (store) {
+        store.dispatch(logError('Error', 'Error creating react application: ', error))
+      }
+    } catch (logError) {
```
rzueger: - IMO this variable shouldn't be called `logError` - there is already an import called `logError`
- shouldn't we log both errors? (`error` and `logError`)

PR #116, opened 2017-02-15
Pr/entity browser/search form options

ottrob:
add extended search form feature


packages/entity-browser/src/components/SearchForm/SearchForm.js
```diff
@@ -19,25 +19,44 @@ const SearchForm = props => {
     })
   }
 
+  const showField = formDefinition => {
```
MethodenMann: const showField = formDefinition => (props.disableSimpleSearch
       || props.simpleSearchFields.includes(formDefinition.name) || props.showExtendedSearchForm)

packages/entity-browser/src/dev/rest-responses/messages.json
```diff
@@ -5,5 +5,6 @@
   "client.entity-browser.validationError",
   "client.entity-browser.search",
   "client.entity-browser.reset",
-  "client.entity-browser.refresh"
+  "client.entity-browser.refresh",
+  "client.entity-browser.extended-search"
```
MethodenMann: camelCase

packages/entity-browser/src/components/SearchForm/SearchForm.js
MethodenMann: i would just pass the fieldName (i think formDefiniton is wrong either way since its a field definitioN)

packages/entity-browser/src/modules/entityBrowser/reducer.js
```diff
@@ -6,9 +6,28 @@ const showEntityDetail = (state, {payload}) => ({
   showDetailEntityId: payload.entityId
 })
 
+const simpleSearchFieldsToArray = simpleSearchFields => {
```
MethodenMann: const simpleSearchFieldsToArray = simpleSearchFields => (
  simpleSearchFields.split(',')
     .filter(s => s)
     .map(s => .trim())
 )

should do the job

ottrob: done

packages/entity-browser/src/dev/rest-responses/messages.json
ottrob: done

packages/entity-browser/src/components/SearchForm/SearchForm.js
ottrob: done

packages/entity-browser/src/modules/entityBrowser/reducer.js
MethodenMann: 👍  also omit  { return

packages/entity-browser/src/modules/entityBrowser/reducer.js
```diff
@@ -6,9 +6,26 @@ const showEntityDetail = (state, {payload}) => ({
...
+  return simpleSearchFields.split(',')
+    .filter(s => s)
+    .map(s => s.trim())
+}
+
+const simpleSearchFields = (state, {payload}) => {
+  const fieldArray = simpleSearchFieldsToArray(payload.simpleSearchFields)
+  return {
+    ...state,
+    simpleSearchFields: (fieldArray.length === 0) ? ['txtFulltext'] : fieldArray
```
MethodenMann: this ['txtFulltext']  is rock solid in any scenario?

packages/entity-browser/src/modules/entityBrowser/reducer.js
```diff
@@ -17,6 +34,8 @@ const ACTION_HANDLERS = {
 const initialState = {
   entityName: '',
   formBase: '',
+  showSearchForm: true,
```
MethodenMann: add intialState for simpleSearchFields

packages/entity-browser/src/modules/entityBrowser/reducer.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    describe('entityBrowser', () => {
+      describe('reducer', () => {
+        it('should create a valid initial state', () => {
+          expect(reducer(undefined, {})).to.deep.equal(EXPECTED_INITIAL_STATE)
+        })
+
+        it('should handle SET_SIMPLE_SEARCH_FIELDS', () => {
+          let state = EXPECTED_INITIAL_STATE
+          state = reducer(state, actions.setSimpleSearchFields('relRel1, field1,field2, relRel2'))
+          const simpleSearchFields = ['relRel1', 'field1', 'field2', 'relRel2']
```
MethodenMann: IMO 'expectedSimpleSearchFields' is better. (for consistency to other tests)

packages/entity-browser/src/modules/entityBrowser/sagas.js
```diff
@@ -16,7 +16,7 @@ export default function* sagas() {
 
 export function* initialize() {
   const entityBrowser = yield select(entityBrowserSelector)
-  let {entityName, formBase} = entityBrowser
+  let {entityName, formBase, showSearchForm, disableSimpleSearch, simpleSearchFields} = entityBrowser
```
MethodenMann: maybe we should set this const and add seperate var for formBase. Seams ugly that we use let for all there vars just because of one

packages/entity-browser/src/modules/entityBrowser/reducer.js
ottrob: This is the same behaviour than we have now. Errors won't be appear in any case. If simple search fields are enabled an no simple search fields are defined in the config `txtFulltext` is the default. But if the search form does not define any txtFulltext field then just no search field will be displayed. I think this is ok.

PR #117, opened 2017-02-16
pr/display-expressions

rzueger:
Support display expressions in list view of entity browser


packages/tocco-ui/src/FormattedValue/typeFormatters/HtmlFormatter.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+
+const HtmlFormatter = props => {
+  const content = props.value
+  if (!content) {
```
MethodenMann: This just recently changed and gets handled globally:
https://github.com/tocco/tocco-client/blob/master/packages/tocco-ui/src/FormattedValue/FormattedValue.js#L12

(Try to fix this in other components this "sprints-gap")

packages/entity-browser/src/modules/listView/sagas.js
```diff
@@ -59,14 +60,18 @@ const extractFields = columnDefinition => {
...
 export function* fetchEntitiesAndAddToStore(page) {
-  const entityBrowser = yield select(listViewSelector)
-  const {entityName, orderBy, limit, entityStore, columnDefinition} = entityBrowser
+  const listView = yield select(listViewSelector)
+  const {entityName, orderBy, limit, entityStore, columnDefinition} = listView
 
   if (!entityStore[page]) {
+    const entityBrowser = yield select(entityBrowserSelector)
+    const {formBase} = entityBrowser
+    const formName = formBase + '_list'
```
MethodenMann: better use template literals

packages/entity-browser/src/util/api/entities.js
```diff
@@ -87,10 +92,11 @@ export const entitiesListTransformer = json => {
 const defaultEntitiesTransformer = json => (json)
 
 export function fetchEntities(entityName, page, orderBy = {}, limit, fields = [],
-                              searchInputs = {}, transformer = defaultEntitiesTransformer) {
+                              searchInputs = {}, transformer = defaultEntitiesTransformer, formName) {
```
MethodenMann: i would move formName before transformer to make them at least a bit ordered. maybe consider a default value to make it clear by the look at the signature that param is not mandatory?

PR #118, opened 2017-02-16
pr/tocco-theme

niklaushug:


packages/entity-browser/src/components/DetailView/DetailView.js
```diff
@@ -4,7 +4,9 @@ import DetailForm from './DetailForm'
 import syncValidation from '../../util/syncValidation'
 import {asyncValidate, AsyncValidationException} from '../../util/asyncValidation'
 
-import './styles.scss'
+if (__DEV__) {
```
niklaushug: Your right, 'tocco-theme/src/ToccoTheme/theme.scss' has always to be required in main.js.

niklaushug: done

packages/entity-browser/src/components/ListView/ListView.js
```diff
@@ -1,7 +1,10 @@
 import React from 'react'
 import {intlShape} from 'react-intl'
 import {Button, FormattedValue, Pagination, Table} from 'tocco-ui'
-import './styles.scss'
+
+if (__DEV__) {
```
niklaushug: done

niklaushug: done

packages/tocco-theme/src/ToccoTheme/theme.scss
```diff
@@ -0,0 +1,32 @@
+/* tocco theme for react components
+ * CSS class .tocco-ui-theme is used to scope styles.
+ * CSS class .tocco-ui-theme is used to increase specificity compared to standard vendors implementation. Hence Tocco's Bootstrap has a higher specificity than Bootstrap.
```
niklaushug: A Stylelint rule does exist to limit amount of characters (https://stylelint.io/user-guide/rules/max-line-length/). I prefer not to limit it but to configure my editor for good readability.

packages/tocco-theme/src/ToccoTheme/variable/mutable-imports.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../../../entity-browser/src/components/DetailView/DetailView.scss';
```
niklaushug: I do not think this is prone to errors. Creating styles is always tide to check look and feel in a browser. Therefore you will always realize if something is missing. But I agree that it is necessary to know where to import it. 

We could implement such a function into Webpack (similar to Gulp task assemble_mutable_imports in Nice2). Interesting would it be to trigger "collecting all scss file" by hot reload.

packages/tocco-theme/src/main.js
```diff
@@ -0,0 +1 @@
+
```
niklaushug: done

packages/tocco-ui/src/Table/Table.scss
```diff
@@ -0,0 +1,31 @@
+.tocco-table {
+  table {
+    width: 100%;
+
+    thead tr th {
+      cursor: pointer;
+      /* stylelint-disable */
```
niklaushug: Because of Stylelint rule property-no-vendor-prefix (https://stylelint.io/user-guide/rules/property-no-vendor-prefix/#property-no-vendor-prefix).

I intend to dispense all vendor prefixes in source files to keep them small and only to rely on well established properties. I prefer to add browser prefixes automatically by appropriate gulp or webpack plugins.

packages/tocco-theme/src/ToccoTheme/theme.scss
MethodenMann: please add a max line length of 120. Then its consistent with js-linting and we dont have a dependency on any IDE

packages/tocco-ui/src/Table/Table.scss
MethodenMann: i agree. so in longterm we can remove those vendor specific lines..

packages/tocco-theme/src/ToccoTheme/variable/mutable-imports.scss
uwolfer: I'd also prefer if we could get rid of this file. This breaks IMHO module concepts. Would be nice if this file could be generated at runtime.

packages/tocco-theme/src/ToccoTheme/theme.scss
niklaushug: done

.gitignore
```diff
@@ -10,4 +10,3 @@ coverage
 .idea
 *.iml
 .eslintcache
-
```
niklaushug: I have configured my editor to add an newline to the end of each file. 

https://www.google.ch/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=file+ends+with+empty+line

.stylelintrc
```diff
@@ -0,0 +1,141 @@
+{
+  "plugins": [
+    "stylelint-scss"
+  ],
+  "rules": {
+    "color-hex-case": "lower",
```
niklaushug: https://stylelint.io/user-guide/example-config/ lists all rules and options.
https://github.com/stylelint/stylelint-config-standard is a recommendation derived of several styleguides.
Stylelint Option "configOverrides" (https://stylelint.io/user-guide/node-api/) should it allow to use several configuration in parallel. Last configured rule is adopted, if a rule is set more than once.
Current .stylelintrc was built up from scratch. It would be interesting to compare it with https://github.com/stylelint/stylelint-config-standard

packages/tocco-theme/src/ToccoTheme/variable/mutable-imports.scss
niklaushug: I will check what it means to automize this process.

niklaushug: I have written a Webpack plugin (ActualizeMutableScssImports.js).

PR #119, opened 2017-02-17
Pr/entity browser/search form default values

ottrob:
Add the possibility to preset search field values and also to set them hidden if desired

### DIFF
--- a/packages/entity-browser/src/components/SearchForm/SearchForm.js
+++ b/packages/entity-browser/src/components/SearchForm/SearchForm.js
@@ -13,18 +13,20 @@ const SearchForm = props => {
     props.setSearchInput()
   }
 
-  const msg = id => {
-    return props.intl.formatMessage({
-      id
-    })
-  }
+  const msg = id => (props.intl.formatMessage({id}))
 
   const showField = name => (
-    props.disableSimpleSearch || props.simpleSearchFields.includes(name) || props.showExtendedSearchForm
+    isHidden(name) || props.disableSimpleSearch || props.simpleSearchFields.includes(name)
+    || props.showExtendedSearchForm
   )
 
   const toggleExtendedSearchForm = () => {
     props.setShowExtendedSearchForm(!props.showExtendedSearchForm)
+  }
+
+  const isHidden = name => {
+    const field = props.preselectedSearchFields.find(f => f.id === name)
+    return field !== undefined && field.hidden
   }
 
   return (
@@ -32,7 +34,7 @@ const SearchForm = props => {
       {
       props.searchFormDefinition.map((definition, idx) => (
         showField(definition.name)
-        && <div key={idx} className="form-group row">
+        && <div key={idx} className={'form-group row' + (isHidden(definition.name) ? ' hidden' : '')}>
           <label htmlFor={definition.name} className="col-sm-2 col-form-label">{definition.label}</label>
           <div className="col-sm-10">
             <SearchField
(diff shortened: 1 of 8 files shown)


packages/entity-browser/src/components/SearchForm/SearchForm.js
```diff
@@ -31,7 +33,7 @@ const SearchForm = props => {
     <form onSubmit={handleSubmit}>
       {
       props.searchFormDefinition.map((definition, idx) => (
-        showField(definition.name)
+        renderField(definition.name)
```
ottrob: done

packages/entity-browser/src/components/SearchForm/SearchForm.js
```diff
@@ -13,14 +13,16 @@ const SearchForm = props => {
...
 
-  const msg = id => {
-    return props.intl.formatMessage({
-      id
-    })
+  const msg = id => (props.intl.formatMessage({id}))
+
+  const isHidden = name => {
+    const field = props.preselectedSearchFields.find(f => f.id === name)
+    return field !== undefined && field.hidden
```
rzueger: I think `return field && field.hidden` is generally better in cases like this, because everything would still work if line 19 was changed so that it returns `null` instead of `undefined`. We don't care if the field is `undefined` or `null`, we just want to know if "it's there".

packages/entity-browser/src/components/SearchForm/SearchForm.js
```diff
@@ -13,14 +13,16 @@ const SearchForm = props => {
...
   }
 
-  const showField = name => (
-    props.disableSimpleSearch || props.simpleSearchFields.includes(name) || props.showExtendedSearchForm
+  const shouldRenderField = name => (
+    !isHidden(name) && (props.disableSimpleSearch || props.simpleSearchFields.includes(name)
+    || props.showExtendedSearchForm)
```
rzueger: * Could be improved by swapping `simpleSearchFields.includes(name)` and `props.showExtendedSearchForm`, since checking the `showExtendedSearchForm` boolean is cheaper than the `includes()` call.
* IMO the readability of the condition would be a bit better like this:
  ```
  const shouldRenderField = name =>
    !isHidden(name) && (
      props.disableSimpleSearch || props.showExtendedSearchForm || props.simpleSearchFields.includes(name)
    )

packages/entity-browser/src/modules/searchForm/reducer.spec.js
```diff
@@ -36,6 +37,26 @@ describe('entity-browser', () => {
 
           expect(state.searchInputs).to.deep.equal({firstname: 'Dan', age: 18})
         })
+
+        it('should handle SET_PRESELECTED_SEARCH_FIELDS', () => {
+          const stateOriginal = EXPECTED_INITIAL_STATE
```
rzueger: `stateOriginal` could be removed

packages/entity-browser/src/components/SearchForm/SearchForm.js
```diff
@@ -13,14 +13,16 @@ const SearchForm = props => {
...
+  const isHidden = name => {
+    const field = props.preselectedSearchFields.find(f => f.id === name)
+    return field && field.hidden
   }
 
-  const showField = name => (
-    props.disableSimpleSearch || props.simpleSearchFields.includes(name) || props.showExtendedSearchForm
+  const shouldRenderField = name => (
+    !isHidden(name) && (
+      props.disableSimpleSearch || props.showExtendedSearchForm) || props.simpleSearchFields.includes(name)
```
rzueger: should there really be a `)` after `props.showExtendedSearchForm`?

PR #123, opened 2017-02-21
Pr/entity browser/search param fix

ottrob:


packages/entity-browser/src/modules/listView/sagas.js
```diff
@@ -46,7 +46,16 @@ export function* getSearchInputs() {
     delete searchInputs.txtFulltext
   }
 
-  return searchInputs
+  let result = {}
```
rzueger: `const` instead of `let`

PR #124, opened 2017-02-21
fix(entity-browser): fix the search params for relations

ottrob:


packages/entity-browser/src/modules/listView/sagas.js
```diff
@@ -46,7 +46,16 @@ export function* getSearchInputs() {
     delete searchInputs.txtFulltext
   }
 
-  return searchInputs
+  const result = {}
+  Object.keys(searchInputs).forEach(f => {
+    if (searchForm.entityModel[f] && searchForm.entityModel[f].type === 'relation') {
+      result[f + '.pk'] = searchInputs[f]
```
MethodenMann: please use template literals

packages/entity-browser/src/modules/listView/sagas.js
```diff
@@ -46,7 +46,16 @@ export function* getSearchInputs() {
     delete searchInputs.txtFulltext
   }
 
-  return searchInputs
+  const result = {}
+  Object.keys(searchInputs).forEach(f => {
```
MethodenMann: i would refactor this part in an util file to keep sagas clean

packages/entity-browser/src/modules/listView/sagas.js
```diff
@@ -46,7 +47,7 @@ export function* getSearchInputs() {
     delete searchInputs.txtFulltext
   }
 
-  return searchInputs
+  return getSearchInputsForRequest(searchInputs, searchForm)
```
MethodenMann: call(getSearchInputsForRequest, ....
return result

PR #126, opened 2017-02-22
pr/entity-browser/detail-features

MethodenMann:


packages/entity-browser/src/components/DetailView/DetailForm.js
```diff
@@ -86,12 +88,13 @@ export class DetailForm extends React.Component {
   }
 
   handleKeyPress = event => {
+    console.log('event.target', event.target.tagName)
```
ottrob: remove log

PR #135, opened 2017-02-24
refactor(login): update to appFactory

MethodenMann:
- add missing message
- fix deprecated sage import


packages/login/src/main.js
```diff
@@ -1,53 +1,90 @@
...
+    const setupFetchMocks = require('./dev/fetchMocks')
+    setupFetchMocks(fetchMock)
+
+    // const app = initLoginApp('id', require('./dev/login_input.json'))
+    // uncomment to develop passwordUpdate App
+    const app = initPasswordUpdateApp('id', require('./dev/password_update_input.json'))
+
+    if (module.hot) {
+      module.hot.accept('./modules/reducers', () => {
+        let reducers = require('./modules/reducers').default
```
rzueger: `const` instead of `let`

packages/login/src/main.js
```diff
@@ -1,53 +1,90 @@
...
-  setupFetchMocks(fetchMock)
+const initLoginApp = (id, input, events, publicPath) => {
+  const dispatchActions = [
+    passwordUpdate.setShowOldPasswordField(false),
+    passwordUpdate.setForcedUpdate(true),
+    passwordUpdate.setStandalone(false)
+  ]
 
-  const mountElement = document.getElementById('root')
+  const showTitle = typeof input.showTitle === 'boolean' ? input.showTitle : false
```
rzueger: what about ``!!input.showTitle` to ensure boolean value?

MethodenMann: 👍

PR #145, opened 2017-03-06
Pr/routing modules

MethodenMann:


packages/entity-browser/src/routes/detail/route.js
```diff
@@ -0,0 +1,13 @@
+import DetailViewContainer from './containers/DetailViewContainer'
+import sagas from './modules/sagas'
+import detail from './modules'
+import {reducer as form} from 'redux-form'
+
+export default {
+  container: DetailViewContainer,
+  reducer: {
```
rzueger: maybe `reducers` instead of `reducer`?

packages/tocco-util/src/route/route.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import storeFactory from '../storeFactory'
+import asyncRoute from './asyncRoute'
+import {dispatchInput} from './input'
+
+export const loadRoute = (store, input, importRouteDependencies, path) => props => {
+  const Component = asyncRoute(() =>
+    new Promise(resolve => {
+      importRouteDependencies().then(imported => {
+        const route = imported.default
```
rzueger: A route doesn't necessarily have to contain `reducers` and `sagas` (not sure about reducers, but sagas are clearly optional). I think there should be null checks to avoid null pointers.

PR #148, opened 2017-03-15
Form Refactoring

MethodenMann:


packages/entity-browser/src/components/FormField/formFieldUtil.js
```diff
@@ -0,0 +1,98 @@
...
+          util.loadRelationEntity(modelField.targetEntity)
+        }
+      }
+  }
+
+  return events
+}
+
+export const getEditableValueProps = (formField, modelField, util) => {
+  const type = fromDefinitionTypeMap[formField.type]
```
rzueger: maybe we should throw or log an error if no `type` could be found for `formField.type` to make it easier to find the issue if a new component type has been added. could be hard to find if `undefined` is returned without logging a warning.

packages/entity-browser/src/routes/detail/components/DetailView/DetailView.js
```diff
@@ -71,9 +72,9 @@ DetailView.propTypes = {
       React.PropTypes.oneOfType([React.PropTypes.string, React.PropTypes.object])))
   ),
   entity: React.PropTypes.object,
-  loadRelationEntities: React.PropTypes.func,
+  loadRelationEntity: React.PropTypes.func,
```
rzueger: IMO, `loadRelationEntity` and `relationEntities` should be declared as required props, since they're *required* in DetailForm and we simply pass them through, even when they're not defined.

packages/entity-browser/src/routes/detail/components/ReduxFormFieldAdapter/ReduxFormFieldAdapter.js
```diff
@@ -0,0 +1,59 @@
...
+  const {
+    input,
+    meta: {dirty, touched, error, submitting},
+    id,
+    formDefinitionField,
+    entityField,
+    modelField,
+    editableValueUtils
+  } = props
+
```
rzueger: not necessary to create a new function on each render. can be moved out of the function and receive `input` as parameter.

PR #151, opened 2017-03-27
Pr/fix dateformatter test saucelabs

rzueger:
Also see:

http://stackoverflow.com/questions/33908299/javascript-parse-a-string-to-date-as-local-time-zone

"Parsing of date strings using the Date constructor or Date.parse (which are essentially the same thing) is strongly recommended against."

### DIFF
--- a/packages/tocco-ui/src/FormattedValue/util/DateUtils.js
+++ b/packages/tocco-ui/src/FormattedValue/util/DateUtils.js
@@ -1,7 +1,13 @@
-/**
- * Adds the timezone offset of the current environment to the passed date to get the date for the UTC timezone.
- */
-export const convertDateToUTC = date => {
-  const timezoneOffset = new Date().getTimezoneOffset() * 60000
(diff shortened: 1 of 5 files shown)


packages/tocco-ui/src/FormattedValue/util/DateUtils.js
```diff
@@ -1,7 +1,13 @@
-/**
- * Adds the timezone offset of the current environment to the passed date to get the date for the UTC timezone.
- */
-export const convertDateToUTC = date => {
-  const timezoneOffset = new Date().getTimezoneOffset() * 60000
-  return new Date(date.getTime() + timezoneOffset)
+export const parseLocalDate = s => {
```
MethodenMann: why calling it parse**Local**Date? its it just a parsing of a date in certain format. not 100% sure but i think its the ISO standard format?

I would rename parameter 's' to something more meaningful

PR #158, opened 2017-04-04
Select refactoring

MethodenMann:


packages/entity-browser/src/routes/entity-browser/modules/actions.js
```diff
@@ -63,3 +67,26 @@ export const setRelationEntity = (entityName, entities, reset = false) => ({
   }
 })
 
+export const loadRemoteEntity = (field, entityName, searchTerm) => ({
```
ottrob: We normaly say field if its not a relation. Maybe replace `field` with `relation`? The same would be for the other new actions.

PR #162, opened 2017-04-07
Plop

MethodenMann:
- support of routes
- general refactoring and cleanup of plopfile


plop/generators/action.js
```diff
@@ -0,0 +1,41 @@
...
+  prompts: [
+    prompts.package,
+    prompts.route,
+    prompts.module,
+    {
+      type: 'input',
+      name: 'action',
+      message: 'Action Name',
+      validate: value => {
+        if ((/.+/).test(value)) {
```
uwolfer: Could be simplified.

PR #167, opened 2017-04-11
add basic support for document fields

uwolfer:


packages/tocco-ui/src/EditableValue/typeEditors/Document.js
```diff
@@ -0,0 +1,17 @@
...
+    <div className="form-control-static">
+      <a href={props.value.binaryLink} download>
+        {props.value.fileName}
+      </a>
+    </div>
+  )
+}
+
+Document.propTypes = {
+  value: React.PropTypes.object
```
rzueger: -  `shape` should be used instead of `object`
- should be marked as required (otherwise null should be handled so that there are no null pointers at line 6)?

packages/tocco-ui/src/EditableValue/typeEditors/Document.js
```diff
@@ -0,0 +1,17 @@
...
+
+const Document = props => {
+  return (
+    <div className="form-control-static">
+      <a href={props.value.binaryLink} download>
+        {props.value.fileName}
+      </a>
+    </div>
+  )
+}
```
rzueger: (`{...}` and `return` not needed:

```
... props => (
    <div>..
```
)

packages/tocco-ui/src/EditableValue/typeEditors/Document.js
MethodenMann: Right now this component is not exposed and can only be used inside <EditableValue>. a null-check is handled there. So value can not be null. 
Setting it as required is a good idea. Should be done on any types implementation though..

PR #168, opened 2017-04-11
Pr/dirty form leave confirmation

rzueger:


packages/tocco-test-util/src/context/wrapWithContext.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+
+const wrapWithContext = (context, contextTypes, children) => {
+  const wrapperWithContext = React.createClass({
```
MethodenMann: We are about to update react to 15.5.0 where createClass is deprecated.
https://github.com/facebook/react/releases

packages/entity-browser/src/routes/detail/components/DetailForm/DetailForm.spec.js
```diff
@@ -104,6 +126,74 @@ describe('entity-browser', () => {
...
+        )
+
+        mount(
+          <Provider store={store}>
+            {formComponent}
+          </Provider>
+        )
+
+        // form is untouched -> context.history.block must not be called
+        expect(block).not.to.be.calledWith('client.entity-browser.confirmTouchedFormLeave')
```
MethodenMann: i think expect(block).not.to.be.called() would be better since we dont care about the argument in this case.

PR #172, opened 2017-04-13
add thumbnail to document field

uwolfer:


packages/tocco-ui/src/EditableValue/typeEditors/Document.js
```diff
@@ -3,15 +3,17 @@ import React from 'react'
 const Document = props => (
   <div className="form-control-static">
     <a href={props.value.binaryLink} download>
+      <img src={props.value.thumbnailLink} style={{verticalAlign: 'top', marginRight: '10px'}}/>
```
MethodenMann: Temporary Styling? If not i would replace inline with a class

PR #175, opened 2017-04-13
refactor(tocco-ui): Support multiple values in <DateAbstract>

rzueger:
At the moment, we have only the date components `DateEdit` and
`DateTimeEdit`, which need only one value.
However, there are more components to come which will need support
for multiple values (for instance a date range component).


packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -22,7 +22,8 @@ class DateAbstract extends React.Component {
       altInput: true,
       onChange: this.handleOnChange.bind(this),
       clickOpens: false,
-      ...this.props.options
+      ...this.props.options,
+      defaultDate: this.props.value
```
MethodenMann: i would move this line 1 line up. so wie append any custom options at the end.

packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -45,13 +46,8 @@ class DateAbstract extends React.Component {
...
 
   handleOnChange(selectedDates) {
-    if (!selectedDates || selectedDates.length === 0) {
-      this.props.onChange(null)
-    } else if (selectedDates.length === 1) {
-      this.props.onChange(selectedDates[0].toISOString())
-    } else {
-      throw new Error('EditableValue DateTime: unexpected length of input')
-    }
+    const asIsoString = selectedDates.map(date => date.toISOString())
```
MethodenMann: Just a detail but maybe name asIsoString = isoStrings? if you call it as.. i expect a function

packages/tocco-ui/src/EditableValue/utils.js
```diff
@@ -0,0 +1,9 @@
+export const atMostOne = array => {
+  if (!array || array.length === 0) {
+    return null
+  } else if (array.length === 1) {
+    return array[0]
+  } else {
+    throw new Error('Expected at most one item in array: ' + array.join(', '))
```
MethodenMann: maybe use template literal

PR #184, opened 2017-04-19
Intl fix

MethodenMann:


packages/login/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tocco-login",
-  "version": "0.1.12",
+  "version": "0.1.14",
```
MethodenMann: before you ask, 0.1.13 somehow was already published

PR #201, opened 2017-04-27
feat(entity-list): Support external events passed via props

rzueger:
Example usage of `onRowClick` event of `EntityListApp`:

```
import EntityListApp from 'entity-list'

<EntityListApp onRowClick={id => console.log(id)} .../>
```

### DIFF
--- a/packages/entity-list/src/main.js
+++ b/packages/entity-list/src/main.js
@@ -6,6 +6,10 @@ import EntityListContainer from './containers/EntityListContainer'
 import {getDispatchActions} from './input'
 const packageName = 'entity-list'
 
+const EXTERNAL_EVENTS = [
+  'onRowClick'
+]
+
 const initApp = (id, input, events, publicPath) => {
   const content = <EntityListContainer/>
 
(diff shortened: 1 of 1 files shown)


packages/entity-list/src/main.js
```diff
@@ -6,6 +6,10 @@ import EntityListContainer from './containers/EntityListContainer'
 import {getDispatchActions} from './input'
 const packageName = 'entity-list'
 
+const EXTERNAL_EVENTS = [
```
MethodenMann: why upper-case? till now all const are written camel-case.

packages/entity-list/src/main.js
```diff
@@ -69,7 +81,11 @@ EntityListApp.propTypes = {
   showSearchForm: React.PropTypes.bool,
   preselectedSearchFields: React.PropTypes.object,
   disableSimpleSearch: React.PropTypes.bool,
-  simpleSearchFields: React.PropTypes.arrayOf(React.PropTypes.string)
+  simpleSearchFields: React.PropTypes.arrayOf(React.PropTypes.string),
+  ...EXTERNAL_EVENTS.reduce((propTypes, event) => {
```
MethodenMann: griffelspitz : ...EXTERNAL_EVENTS.reduce((propTypes, event) => ({...propTypes, [event]: React.PropTypes.func}), {}) should work as well

PR #211, opened 2017-05-08
entity list bug fixing

MethodenMann:


packages/tocco-ui/src/LoadMask/LoadMask.js
```diff
@@ -13,13 +15,20 @@ class LoadMask extends React.Component {
   }
 
   componentDidMount() {
+    this.mounted = true
     if (this.props.promises) {
       Promise.all(this.props.promises).then(() => {
-        this.setState({initialized: true})
+        if (this.mounted) {
```
ottrob: this.mounted is always `true`?

MethodenMann: its initial value is false and as soon as loadmask gets unmounted it is set to false. this checks avoids the case, where the components gets unmounted and the promise resolves after that what would cause an error because you cant call "setState" on an unmounted component

PR #214, opened 2017-05-08
Pr/init detail page

ottrob:
add the initial detail page feature

### DIFF
--- a/packages/entity-browser/src/dev/input.json
+++ b/packages/entity-browser/src/dev/input.json
@@ -23,5 +23,6 @@
       "value": ["1", "3"],
       "hidden": false
     }
-  ]
+  ],
+  "initialKey": ""
 }
--- a/packages/entity-browser/src/main.js
+++ b/packages/entity-browser/src/main.js
@@ -42,7 +42,6 @@ const navigateToDetailIfKeySet = (history, input) => {
 }
 
 const initApp = (id, input, events, publicPath) => {
-
   const store = appFactory.createStore(undefined, undefined, input, packageName)
 
   const history = createHistory(store)
--- a/packages/entity-browser/src/main.js
+++ b/packages/entity-browser/src/main.js
@@ -28,10 +28,24 @@ const createHistory = store => createHashHistory({
   }
 })
 
+const initHistory = (store, input) => {
+  const history = createHistory(store)
+  const initialDetailViewKey = input.initialKey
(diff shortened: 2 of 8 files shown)


packages/entity-browser/src/main.js
```diff
@@ -28,10 +28,24 @@ const createHistory = store => createHashHistory({
   }
 })
 
+const initHistory = (store, input) => {
+  const history = createHistory(store)
+  const initialDetailViewKey = input.initialKey
+  if (initialDetailViewKey && initialDetailViewKey !== '' && !isNaN(initialDetailViewKey)) {
```
MethodenMann: if (initialDetailViewKey) implies (initialDetailViewKey !== '')

packages/entity-browser/src/main.js
```diff
@@ -28,10 +28,24 @@ const createHistory = store => createHashHistory({
   }
 })
 
+const initHistory = (store, input) => {
+  const history = createHistory(store)
```
MethodenMann: i personally would create the store outside of initHistory and pass it instead of store and name initHistory something like "navigate to detail if initial key set" to make intentions clearer

packages/entity-browser/src/util/detailView/showBackButton.js
```diff
@@ -0,0 +1,8 @@
+const showBackButton = (initialKey, modelPaths) => {
+  if (modelPaths.length === 0) {
+    return (initialKey === '' || isNaN(initialKey))
```
MethodenMann: this requires the intialKey to be set in input? instead i would write return !initialKey || ... to make it optional

packages/entity-browser/src/util/detailView/showBackButton.spec.js
```diff
@@ -0,0 +1,31 @@
...
+
+        it('should return `false` if modelPaths is empty and initialKey is a number', () => {
+          const initialKey = '123'
+          const modelPaths = []
+          const result = showBackButton(initialKey, modelPaths)
+          expect(result).to.be.false
+          expect(result).to.be.false
+        })
+      })
+    })
```
MethodenMann: add initialKey = undefined test to check previous comment

packages/entity-browser/src/dev/input.json
```diff
@@ -23,5 +23,6 @@
       "value": ["1", "3"],
       "hidden": false
     }
-  ]
+  ],
+  "initialKey": ""
```
MethodenMann: please add documentation of input in readme of entity-browser

PR #219, opened 2017-05-09
EntityBrowser ReadOnly

MethodenMann:


packages/tocco-ui/src/FormattedValue/typeFormatters/SingleSelectFormatter.spec.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import SingleSelectFormatter from './SingleSelectFormatter'
+import {mount} from 'enzyme'
+
+describe('tocco-ui', function() {
```
ottrob: 'tocco-ui', () => {...

packages/tocco-ui/src/FormattedValue/typeFormatters/MultiSelectFormatter.spec.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import MultiSelectFormatter from './MultiSelectFormatter'
+import {mount} from 'enzyme'
+
+describe('tocco-ui', function() {
```
ottrob: 'tocco-ui', () => {

packages/entity-browser/src/routes/detail/components/DetailForm/DetailForm.js
```diff
@@ -40,6 +42,7 @@ export class DetailForm extends React.Component {
     return (
       <Field
         key={key}
+        readOnlyForm={this.isReadOnlyForm}
```
ottrob: A single field also can be readonly if the whole form is not readonly. In my opinion it should be something like {this.isReadOnlyForm || formDefinitionField.isReadOnly}

MethodenMann: yes, it can. but we only want to show "formattedValues" if the whole form is readonly. Otherwise we want "editableValues" and if a single field is readonly it is disabled.

PR #232, opened 2017-05-18
Fix Empty Records

MethodenMann:


packages/tocco-ui/src/LoadMask/LoadMask.js
```diff
@@ -76,9 +76,9 @@ LoadMask.propTypes = {
    */
   loadingText: React.PropTypes.string,
   /**
-   * As soon as all object are truthy, children will be displayed.
+   * As soon as all elements of the array are truthy, children will be displayed.
    */
-  required: React.PropTypes.arrayOf(React.PropTypes.object),
+  required: React.PropTypes.array,
```
rzueger: or `React.PropTypes.arrayOf(React.PropTypes.any)` to make it explicit that the array can contain anything

PR #253, opened 2017-05-24
feat(entity-list): integrate react-bootstrap-table

MethodenMann:
- replace tocco-ui table and pagination with react-bootstrap-table
- hide pagination if no more than one page exists
- add __key attribute to entity and story fields directly in object and not in an attribute values (entities.js).
  This change was needed to add mandatory key columns to react bootstrap table
- Columndefinition: Only regard first field in columns. To add two fields into one column, freemarker is required.
  This change made handling with react-bootstrap-table easier and does not require a hack.

### DIFF
--- a/packages/entity-list/src/components/ListView/ListView.js
+++ b/packages/entity-list/src/components/ListView/ListView.js
@@ -71,7 +71,7 @@ class ListView extends React.Component {
           loadingText={this.msg('client.entity-list.loadingText')}
         >
           <BootstrapTable
-            remote={true}
+            remote
             data={props.entities}
             pagination={ showPagination }
             fetchInfo={{dataTotalSize: props.entityCount}}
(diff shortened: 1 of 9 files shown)


packages/entity-list/src/components/ListView/ListView.js
```diff
@@ -1,75 +1,100 @@
...
-            />
-            <Button
-              onClick={props.refresh}
-              label={this.msg('client.entity-list.refresh')}
-              icon="glyphicon-refresh"
-              className="refresh-button"
-            />
-          </div>
+          <BootstrapTable
+            remote={true}
```
rzueger: could be moved at the end of the tag without `={true}` (like `striped` and `hover`)

same applies to `isKey` and `dataSort` below

MethodenMann: agree, but i won't move it to the end since these are essential props

PR #262, opened 2017-05-30
Pr/remote fields more options available

ottrob:


packages/tocco-ui/src/EditableValue/typeEditors/MultiRemoteSelect.js
```diff
@@ -8,6 +8,18 @@ class MultiRemoteSelect extends React.Component {
     }
   }
 
+  getOptions = () => {
```
MethodenMann: getOptions = () => {
    const options  =  this.props.options.options || []
     if (this.props.options.moreOptionsAvailable) {
         optionspush( {display: this.props.options.moreOptionsAvailableText, disabled: true})
       } 
     }
     return options
   }

should also work? shorter and in my optinion better regarding complexity

ottrob: correct

PR #284, opened 2017-06-14
chore(tocco-client): switch license to AGPLv3+ (adding the +)

pgerber:
Omitting the + means we can't upgrade to a new version of the
license should the copyright holder of any commit ever be any
other party than Tocco. Also, we really want to be able to
upgrade in case the license turns out to be flawed, legally
unenforceable or if legal circumstances change.

note: SPDX allows a + suffix to indicate "or any later version"

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     }
   },
   "author": "Tocco AG",
-  "license": "AGPL-3.0",
+  "license": "AGPL-3.0+",
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


pgerber:
The warning was only issued because AGPL-3.0 instead of AGPLv3 had been used. The "license should be a valid SPDX license expression" warning isn't shown after adding a plus, as long as the license name is correct.

BTW, decided to force push to fix a mistake in the commit message.

PR #292, opened 2017-06-30
Pr/rest sagas

rzueger:


packages/entity-browser/src/util/api/entities.js
```diff
@@ -17,8 +18,7 @@ export const defaultModelTransformer = json => {
   return model
 }
 
-export function fetchModel(entityName, transformer = defaultModelTransformer) {
-  return request(`entities/${entityName}/model`)
-    .then(resp => resp.body)
-    .then(json => transformer(json))
+export function* fetchModel(entityName, transformer = defaultModelTransformer) {
+  const resp = yield call(requestSaga, `entities/${entityName}/model`)
```
MethodenMann: since GET is the default (as used here), i think we dont need "getRequestSaga" and can remove it?

packages/entity-browser/src/util/api/entities.js
```diff
@@ -17,8 +18,7 @@ export const defaultModelTransformer = json => {
...
+  return transformer(resp.body)
```
MethodenMann: shouldn't it be `return call(transformer, resp.body)`?
testing would be easier

PR #297, opened 2017-07-04
fix(entity-list): relation preselectedSearchFields

MethodenMann:
allow to transform searchInputs before storing  them as searchInput.
This change was necessary to allow a key as input but set display/key pair as search value.

Signed-off-by: Daniel Keller <dkeller@tocco.ch>


packages/entity-list/src/modules/searchForm/sagas.js
```diff
@@ -27,21 +26,26 @@ export function* loadSearchForm(formDefinition, searchFormName) {
...
-    const value = searchInputs[relationField]
-    return (value instanceof Array && value.length > 0) || (!(value instanceof Array) && value)
-  })
-  const targetEntities = _uniq(
-   relationFieldsWithValue.map(relationField => (entityModel[relationField].targetEntity))
-  )
-
-  yield all(targetEntities.map(targetEntity => put(actions.loadRelationEntity(targetEntity))))
+export function* setInitialSearchInputs(entityModel, preselectedSearchFields) {
+  for (let preselectedSearchField of preselectedSearchFields) {
```
rzueger: or `const`?

packages/entity-list/src/modules/searchForm/sagas.js
```diff
@@ -27,21 +26,26 @@ export function* loadSearchForm(formDefinition, searchFormName) {
...
+    const value = preselectedSearchField.value
+
+    let transformedValue = value
+
+    if (entityModel[fieldName] && entityModel[fieldName].type === 'relation') {
+      const targetEntity = entityModel[fieldName].targetEntity
+      const entities = yield call(loadRelationEntity, actions.loadRelationEntity(targetEntity))
+
+      if (Array.isArray(value)) {
+        transformedValue = value.map(v => (entities.find(e => e.key === v)))
```
rzueger: parenthesis around `entities.find()` not required

packages/entity-list/src/modules/searchForm/sagas.js
```diff
@@ -56,11 +60,13 @@ export function* getEntityModel() {
 }
 
 export function* initialize() {
-  let {formDefinition, searchInputs, searchFormName} = yield select(searchFormSelector)
-
+  let {formDefinition, preselectedSearchFields, searchFormName} = yield select(searchFormSelector)
```
rzueger: can be `const` now

PR #300, opened 2017-07-10
refactor: Move redux-form utils from entity-detail to tocco-util

rzueger:
- This way, we can use it in other packages as well (for example for
  the search form in the entity-list package)
- formFieldMappings must still be defined in entity-detail (we can't
  move it to tocco-util, because this would result in circular
  dependencies because of the sub grid component which requires the
  entity-list package)
- Only small new feature: `formBuilder` doesn't require an `entity`
   argument anymore (first parameter). If no entity is given, the
   readability check for the entity paths is simply omitted
   (this way we can use it for the search form as well, where we
   won't have any entity data).


packages/entity-detail/src/components/DetailForm/DetailForm.js
```diff
@@ -2,10 +2,14 @@ import React from 'react'
 import {reduxForm} from 'redux-form'
 import {intlShape, FormattedRelative, FormattedMessage} from 'react-intl'
 import {Button, LayoutBox} from 'tocco-ui'
+
+import formErrorsUtil from 'tocco-util/src/form/formErrors'
+import initFormBuilder from 'tocco-util/src/form/formBuilder'
+import {getFieldId} from 'tocco-util/src/form/helpers'
```
MethodenMann: we usually import tocco-util stuffe like this

import {form} from 'tocco-util'

then

form.formErrors ....

PR #301, opened 2017-07-10
Better Mock data 

MethodenMann:


packages/tocco-util/src/mockData/entityFactory.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    describe('entityFactory', () => {
+      describe('createDummyEntities', () => {
+        it('should create right amount of dummy entities', () => {
+          const amount = 101
+          const dummyEntities = createDummyEntities(amount)
+          expect(dummyEntities.length).to.eql(amount)
+        })
+      })
+
+      describe('createDummyEntities', () => {
```
rzueger: `describe('createUsers'...`

PR #308, opened 2017-07-14
refactor(entity-list): set initial search form values

MethodenMann:


packages/entity-list/src/modules/searchForm/actions.js
```diff
@@ -8,7 +8,6 @@ export const RESET_SEARCH = 'searchForm/RESET_SEARCH'
 export const SET_SHOW_EXTENDED_SEARCH_FORM = 'searchForm/SET_SHOW_EXTENDED_SEARCH_FORM'
 export const SET_SIMPLE_SEARCH_FIELDS = 'searchForm/SET_SIMPLE_SEARCH_FIELDS'
 export const PREPARE_PRESELECTED_SEARCH_FIELDS = 'searchForm/PREPARE_PRESELECTED_SEARCH_FIELDS'
```
rzueger: I would rename `PREPARE_PRESELECTED_SEARCH_FIELDS` to `SET_PRESELECTED_SEARCH_FIELDS` now that this name is free.

PR #312, opened 2017-07-27
Pr/entity list/keep store

MethodenMann:


packages/entity-list/src/modules/list/actions.js
```diff
@@ -134,3 +134,9 @@ export const setListFormName = listFormName => ({
     listFormName
   }
 })
+export const setInitialized = (initialized = true) => ({
```
rzueger: insert empty line above?

PR #315, opened 2017-08-14
Pr/entity list initial order

ottrob:


packages/entity-list/src/util/api/forms.js
```diff
@@ -37,6 +40,16 @@ export const columnDefinitionTransformer = json => {
       child: c.children.filter(isDisplayableChild)[0]
     }
   ))
+
+  if (table.sorting) {
+    const sorting = table.sorting.reduce((acc, val, idx, arr) => {
```
MethodenMann: Have you considered using map instead of reduce? its easier to read imo

const sorting = table.sorting.map(sort => `${sort.orderItem} ${(sort.ascending) ? 'asc' : 'desc'}`).join(',')

ottrob: No I didn't but I don't think it will be easier because we cannot just join the elements by a comma. If there is only one entry in the array no comma should be printed. `(idx + 1 === arr.length) ? '' : ','`.

PR #323, opened 2017-08-22
Pr/login/error handling

MethodenMann:


packages/login/src/modules/passwordUpdate/password/sagas.js
```diff
@@ -13,46 +14,41 @@ export const usernameSelector = state => state.passwordUpdate.dialog.username
...
   }
-  return fetch(`${__BACKEND_URL__}/nice2/rest/principals/${username}/${action}`, options)
-}
 
-export function storePassword(username, data) {
-  return doRequest(data, username, 'password-update')
-    .then(resp => {
+  try {
+    const resp = yield call(requestSaga, `principals/${username}/password-update`, options)
+    if (resp) {
```
rzueger: In which case could `resp` be missing and what should we do in that case (currently, nothing happens)?

packages/tocco-util/src/rest/request.js
```diff
@@ -1,19 +1,24 @@
...
 
-const extractBody = response => (
-  response.json().then(body => {
-    const {ok, headers, status, statusText} = response
-    return {ok, headers, status, statusText, body: body || {}}
+const extractBody = response => {
+  const {ok, headers, status, statusText} = response
+
+  return response.json().then(body => {
+    return {ok, headers, status, statusText, body}
```
rzueger: (`{...}` and `return` not necessary)

packages/login/src/modules/passwordUpdate/password/sagas.js
MethodenMann: you are right, was something i forgot to delete after refactoring

PR #328, opened 2017-09-06
Create View

MethodenMann:


packages/tocco-util/src/form/formBuilder.js
```diff
@@ -42,6 +43,11 @@ export default (
   }
 
   const shouldRenderField = (formDefinitionField, entityField) => {
+    if (mode && formDefinitionField.scopes && formDefinitionField.scopes.length > 0
```
ottrob: I think the check `formDefinitionField.scopes.length > 0` can be removed.

packages/tocco-util/src/form/reduxForm.js
```diff
@@ -64,11 +72,20 @@ export const entityToFormValues = entity => {
   return result
 }
 
+export const emptyValues = modelName => {
```
ottrob: I would prefer 

```
export const emptyValue = modelName => ({
  [wholeEntityField]: {
    entity: {
       model: modelName
    }
  }
})
```

The structure of the result object is immediately clear this way. But it's also ok if we leave it like this.

packages/tocco-util/src/form/reduxForm.spec.js
```diff
@@ -25,9 +25,17 @@ describe('tocco-util', () => {
                 }
               }
             }
+          }
 
+          const entityModel = {
+            firstname: { },
+            gender: {relationName: 'relGender', multi: false},
+            status: {relationName: 'relGender', multi: true}
```
ottrob: replace `relGender` with `relStatus`

packages/entity-list/README.md
```diff
@@ -21,6 +21,7 @@ React-registry name: `entity-list`
 
 ### Events
 
-| Name          | Payload                       | Description
-|---------------|-------------------------------|-------------
-| `onRowClick`  | `id` (The id of the record)   | This event is fired when a list row is clicked
+| Name                | Payload                       | Description
+|---------------------|-------------------------------|-------------
+| `onRowClick`        | `id` (The id of the record)   | This event is fired when a list row is clicked
+| `navigateToCreate`  | -                             | This event is fired when "new" button is clicked
```
ottrob: typo: ... is fired when `the` new button ...

packages/entity-detail/README.md
```diff
@@ -11,14 +11,16 @@ React-registry name: `entity-detail`
 | `entityName`           | x         | Entity name of the record
 | `entityId`             | x         | The ID of the entity 
 | `formName`             |           | Detail-form that should be loaded
+| `mode`                 |           | Determines if the detail is in create (`create`) or update (`update`) mode
```
ottrob: i would simplify the the description to "Determines if the detail is in create or update mode"

packages/entity-browser/src/routes/detail/modules/sagas.js
```diff
@@ -22,18 +23,24 @@ export function* loadEntityDetail({payload}) {
   const {modelPaths, entityId, parentUrl} = yield call(parseUrl, payload.url)
   const {entityName, formBase} = yield select(entityBrowserSelector)
 
+  const mode = yield call(detail.getMode, entityId)
+
   let targetEntityName = entityName
-  let formName = `${formBase}_detail`
+  let formName = `${formBase}`
```
ottrob: `let formName = formBase` ?

packages/tocco-util/src/form/formBuilder.js
MethodenMann: why? by default scopes as returned by REST is an empty array. If no scope set (array empty) we dont want to regard the current mode (create or update) and display the field anyways.

packages/entity-browser/src/routes/detail/modules/sagas.js
MethodenMann: whats is the question? :)

PR #331, opened 2017-09-15
Pr/create button

MethodenMann:


packages/entity-detail/src/components/SubGrid/SubGrid.js
```diff
@@ -48,7 +49,8 @@ SubGrid.propTypes = {
     reverseRelationName: PropTypes.string
   }).isRequired,
   onRowClick: PropTypes.func,
-  onNavigateToCreate: PropTypes.func.isRequired
+  onNavigateToCreate: PropTypes.func.isRequired,
+  showSubGridsCreateButton: PropTypes.bool
```
ottrob: Also adjust the test (SubGrid.spec.js) ?

PR #348, opened 2017-10-13
Pr/scheduler/minor adjustments

ottrob:


packages/entity-list/src/main.js
```diff
@@ -116,6 +116,18 @@ EntityListApp.propTypes = {
   ),
   disableSimpleSearch: PropTypes.bool,
   simpleSearchFields: PropTypes.string,
+  /**
```
MethodenMann: to be consistent i would omit these comments since its a duplication of informations from the readme

PR #362, opened 2017-11-08
Pr/scheduler/two tab

MethodenMann:


packages/entity-browser/src/input.js
```diff
@@ -0,0 +1,7 @@
+import {
+  setAppId
```
ottrob: I would import this function on a single line

PR #365, opened 2017-11-22
Pr/scheduler/search filter field

ottrob:


packages/entity-list/src/modules/list/sagas.js
```diff
@@ -72,18 +72,19 @@ export function* fetchEntitiesAndAddToStore(page) {
   if (!entityStore[page]) {
     const {sorting, limit, columnDefinition} = list
 
-    const {searchFilters} = input
     const formName = `${formBase}_list`
 
     const searchInputs = yield call(getSearchInputs)
+    // combine search filters from input and search form
+    searchInputs._filter = [...input.searchFilters || [], ...searchInputs._filter || []]
```
MethodenMann: searchInputs._filter = call(_union, input.searchFilters, searchInputs._filter)

would also work and is a bit easier to read in my opinion (comment is omittable)

packages/entity-list/src/modules/list/sagas.js
```diff
@@ -72,18 +72,19 @@ export function* fetchEntitiesAndAddToStore(page) {
...
+
     const fields = getFieldsOfColumnDefinition(columnDefinition)
```
MethodenMann: my fault but while you're at it: use call to invoke getFieldsOfColumnDefinition

PR #369, opened 2017-11-28
Pr/actions

MethodenMann:


packages/tocco-util/src/actions/components/GroupElement.js
```diff
@@ -0,0 +1,32 @@
...
+        onClick(definition)
+      }
+    }}>
+      {definition.icon && <i className={'fa ' + definition.icon}/>} {definition.label}
+    </MenuItem>
+  )
+}
+
+GroupElement.propTypes = {
+  definition: PropTypes.object.isRequired,
```
ottrob: I would add a comment for  the `definition` property

packages/tocco-util/src/actions/components/SingleAction.js
```diff
@@ -0,0 +1,25 @@
...
+      }}
+      icon={definition.icon}
+      {...definition.label && definition.useLabel === 'NO' ? {title: definition.label} : {label: definition.label}}
+      {...definition.readOnly === true && {disabled: true}}
+    />
+  )
+}
+
+SingleAction.propTypes = {
+  definition: PropTypes.object.isRequired,
```
ottrob: I would add a comment for the `definition` property

PR #374, opened 2017-12-21
Pr/sublist search fix

MethodenMann:


packages/entity-list/src/modules/searchForm/sagas.js
```diff
@@ -90,7 +90,11 @@ export function* loadRelationEntity({payload}) {
   const {entityName} = payload
   const {relationEntities} = yield select(searchFormSelector)
   if (!relationEntities[entityName] || !relationEntities[entityName].loaded) {
-    const entities = yield call(fetchEntities, entityName, {}, selectEntitiesTransformer)
+    const searchInputs = {
+      fields: ['pk'],
+      limit: 100
```
ottrob: should the limit be fix set to 100?

PR #415, opened 2018-04-09
Chore Update

MethodenMann:


packages/entity-browser/tests/test-bundler.js
```diff
@@ -1,4 +1,8 @@
+import Enzyme from 'enzyme'
+import Adapter from 'enzyme-adapter-react-16'
```
rzueger: I think we should move this import to a separate file (maybe package `tocco-test-util`) and import this file in every test-bundler? I reckon we won't remain with `enzyme-adapter-react-16` once React 17 is here and we don't want to update each test-bundler.js every time React is updated.
```
import Adapter from 'enzyme-adapter-react-16'

export default Adapter
```

packages/tocco-ui-showcase/src/components/ShowCaseApp.spec.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
rzueger: What's that needed for? If I remove this line and run the linter locally, it doesn't report any errors.

packages/tocco-ui/src/FormattedValue/typeFormatters/DocumentFormatter.spec.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
rzueger: What's that needed for? If I remove this line and run the linter locally, it doesn't report any errors.

packages/tocco-ui/src/LoadMask/example.js
```diff
@@ -7,15 +7,24 @@ class Example extends React.Component {
...
+    return true
+  }
+
   render() {
     return (
       <div style={{height: '100px'}}>
         {/* start example */}
-        <button onClick={() => this.setState({loaded: {}})}>Load</button>
+        <button onClick={() => {
+          console.log('this.state.loaded', this.state.loaded)
```
rzueger: Do we want to keep this console log in this example?

packages/tocco-util/src/notifier/components/Notifier.spec.js
```diff
@@ -1,6 +1,7 @@
+/* eslint-disable no-unused-vars */
```
rzueger: What's that needed for? Can we instead just cleanup the imports?

MethodenMann: you're right. i added these ignores for developing purposes and forgot to remove them afterwards. tanks

packages/tocco-test-util/src/testSetup/bundle.js
```diff
@@ -0,0 +1,9 @@
+import { configure } from 'enzyme'
```
rzueger: Is there no linting rule for the spaces inside `{...}` or is this file excluded from the linting? (as far as I can tell, we prefer the style without spaces)

PR #419, opened 2018-04-17
Resource-Scheduler: Show Conflicts

MethodenMann:
- Show border on existing conflicts
- Show localized text in event popover for existing and resolved conflicts
- Add mock data
- Load scheduler mock in resource-scheduler

### DIFF
--- a/packages/scheduler/src/components/Conflict/Conflict.js
+++ b/packages/scheduler/src/components/Conflict/Conflict.js
@@ -24,7 +24,7 @@ const Conflict = ({conflictStatus, intl}) => {
 
 Conflict.propTypes = {
   intl: intlShape.isRequired,
-  conflictStatus: PropTypes.string
+  conflictStatus: PropTypes.oneOf(Object.values(conflicts))
 }
 
 export default Conflict
--- a/packages/scheduler/src/components/Conflict/Conflict.js
+++ b/packages/scheduler/src/components/Conflict/Conflict.js
@@ -1,0 +1,30 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import {intlShape} from 'react-intl'
+import {conflicts} from './utils'
+
+const Conflict = ({conflictStatus, intl}) => {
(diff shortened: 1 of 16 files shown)


packages/scheduler/src/components/Conflict/Conflict.js
```diff
@@ -0,0 +1,30 @@
...
+
+  return <span style={style}>
+    <i className={`fa ${icon}`} style={{paddingRight: '10px'}} aria-hidden="true"></i>
+    {intl.formatMessage({id: textResource})}
+  </span>
+}
+
+Conflict.propTypes = {
+  intl: intlShape.isRequired,
+  conflictStatus: PropTypes.string
```
rzueger: Maybe better to restrict the `conflictStatus` prop to one of the status defined in utils.js so that a warning/error appears in the console if an unexpected prop is defined? Otherwise it could be overlooked here easily if a status changes or a new one is added.

packages/scheduler/src/components/Conflict/Conflict.spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe('scheduler', () => {
+  describe('components', () => {
+    describe('Conflict', () => {
+      it('should render return null if conflictstatus is undefined', () => {
+        const wrapper = shallow(<Conflict intl={IntlStub} />)
+        expect(wrapper.type()).to.be.null
+      })
+
+      it('should render return null if conflictstatus is undefined nones', () => {
```
rzueger: this should probably be 'should return null...' and '... conflictstatus is none'?

packages/scheduler/src/components/Conflict/Conflict.spec.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import Conflict from './Conflict'
+import {shallow} from 'enzyme'
+import {IntlStub} from 'tocco-test-util'
+
+describe('scheduler', () => {
+  describe('components', () => {
+    describe('Conflict', () => {
+      it('should render return null if conflictstatus is undefined', () => {
```
rzueger: this should probably be 'should return null ...'

PR #422, opened 2018-04-19
Resource-Scheduler Colors

MethodenMann:


packages/scheduler/src/components/Scheduler/utils.js
```diff
@@ -10,6 +12,12 @@ export const getResources = calendars => {
   }))
 }
 
+const WHITE_RGB = {r: 255, g: 255, b: 255}
+const getTextColor = hexColor => {
```
rzueger: maybe change signature from `getTextColor(hexColor)` to

`getOptimalTextColor(backgroundHexColor)`

This way it would be clear what this function really returns and what the paramter `hexColor` represents.

packages/tocco-util/src/color/contrast.js
```diff
@@ -0,0 +1,25 @@
...
+
+/***
+ * Returns a dark or bright value depending on the given color.
+ * This could be used to choose a pleasant font color to a background color.
+ * @param rgbColor Object whit this shape: {r: 0, g: 0, b: 0}
+ * @param brightValue Bright value can be overridden. Default is the color white as hex.
+ * @param darkValue Dark value can be overridden. Default is the color black as hex.
+ * @returns {string}
+ */
+export const getContrastColor = (rgbColor, brightValue = '#FFFFFF', darkValue = '#000000') =>
```
rzueger: I personally don't like that we use two different formats for parameters of the same type. `rgbColor`, `brightValue` and `darkValue` all represent a color, but for `rgbColor` we use an object and for the other two a string and we also return a string.

What about expecting for `rgbColor` a hex string as well and parse it here instead of in the scheduler? I reckon 
 this util would be more useful in other places as well if it could handle a hex string for `rgbColor`.

PR #424, opened 2018-04-24
refactor(tocco-util): various form-util improvements

MethodenMann:
- move formDefinition util from entity-detail package to tocco-util to reuse it later in search form
- add tests for formDefintion utils
- move enum like object in separate folder and freeze object


packages/tocco-util/src/form/formDefinition.spec.js
```diff
@@ -0,0 +1,147 @@
...
+        }
+      ]
+    },
+    testField3
+  ]
+}
+
+describe('tocco-util', () => {
+  describe('form', () => {
+    describe('helper', () => {
```
rzueger: shouldn't this be 'formDefinition' instead of 'helper'?

packages/tocco-util/src/form/formDefinition.spec.js
```diff
@@ -0,0 +1,147 @@
...
+      beforeEach(() => {
+        fetchMock.reset()
+        fetchMock.restore()
+      })
+
+      describe('getFieldDefinitions', () => {
+        it('return an array of fields', () => {
```
rzueger: ('should return ...'? 😁)

packages/tocco-util/src/form/formDefinition.spec.js
```diff
@@ -0,0 +1,147 @@
...
+          }
+
+          expect(gen.next(resp).value).to.eql(call(formDefinition.defaultFormTransformer, resp.body))
+
+          const next = gen.next(resp.body.form)
+
+          expect(next.value).to.eql(resp.body.form)
+          expect(next.done).to.be.true
+        })
+      })
```
rzueger: (maybe insert empty line here)

packages/tocco-util/src/form/formDefinition.spec.js
```diff
@@ -0,0 +1,147 @@
...
+
+      describe('getFieldId', () => {
+        it('should return a string containing formName and fieldName', () => {
+          const formName = 'User_detail'
+          const fieldName = 'firstname'
+          const result = formDefinition.getFieldId(formName, fieldName)
+          expect(result).to.have.string(formName)
+          expect(result).to.have.string(fieldName)
```
rzueger: Are we using the generated ID ("input-{formName}-{fieldName}") just for simple identification - or does it matter what it looks like exactly?
For example, if the ID will be used for styling purposes as well (CSS selectors) it would in my opinion make sense to check the generated ID explicitly against the actual format ("input-{formName}-{fieldName}").
(now the test implies that it really doesn't matter how the ID looks like, as long as the form name and the field name are in there - and that we can change the actual format without having to fear any side effects)

MethodenMann: at the moment this should generate an unique id for each field on the form. We're using it to set the focus on these fields if there is a validation error.
Since we rewrite the application to styled components i think css selector compatibility can be ignored.

PR #426, opened 2018-04-26
feat(resource-scheduler): display color for each calendar type

MethodenMann:
- adjust propTypes
- enhance mock data


packages/resource-scheduler/src/components/SearchPanel/SearchPanel.specs.js
```diff
@@ -17,6 +17,37 @@ describe('resource-scheduler', () => {
         )
         expect(wrapper.find('div')).to.have.length(1)
       })
+
+      it('should show show colored icon if color is defined', () => {
```
rzueger: 'show show'

PR #437, opened 2018-05-29
Pr/resource scheduler/refresh loading

MethodenMann:


packages/scheduler/src/components/FullCalendar/FullCalendar.js
```diff
@@ -38,14 +38,20 @@ class FullCalendar extends React.Component {
     }
   }
 
-  fixOptions = {
-    customButtons: {
-      refresh: {
-        bootstrapGlyphicon: 'glyphicon-refresh',
-        click: () => {
+  getRefreshButton = isLoading => {
+    return {
```
rzueger: (`{ ... return ... }` not necessary)

PR #440, opened 2018-06-04
Document in List

MethodenMann:


packages/tocco-ui/src/FormattedValue/typeFormatters/DocumentCompactFormatter.js
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+const DocumentFormatter = props => (
+  <span className="form-control-static document-compact">
+    <a
+      download="Test.pdf"
```
rzueger: hardcoded string should be replaced with prop

PR #445, opened 2018-06-18
Pr/styled components phase 3 a

niklaushug:


packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -0,0 +1,57 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+/**
+ * <Panel/> is used to conceal and display related content alternating by interaction.
+ */
+class Panel extends React.Component {
+  state = {
```
MethodenMann: does this really work? usually the state is set with this.setState

packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -0,0 +1,57 @@
...
+    isOpen: this.props.isOpen
+  }
+
+  toggleOpenState = () => {
```
MethodenMann: toggleOpenState = () => {
   if (this.props.isToggleable) {
    this.setState(prevState => ({
      isOpen: !prevState.isOpen
...

packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -0,0 +1,57 @@
...
+
+  render() {
+    const {
+      isToggleable
+    } = this.props
+
+    return (
+      <div>
+        { // eslint-disable-next-line
+          React.Children.map(this.props.children, child => {
```
MethodenMann: Why do you disable eslint?

Is it because of missing propType for children? Please add them instead.

Please never user eslint-disable unless you are really sure.

packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -0,0 +1,57 @@
...
+            return React.cloneElement(child, {
```
MethodenMann: 'return' statement can be omitted

packages/tocco-ui/src/Panel/PanelHeaderFooter.js
```diff
@@ -0,0 +1,55 @@
...
+      <StyledPanelHeaderFooter>
+        <div>
+          {React.Children.map(children, child => React.cloneElement(child))}
+        </div>
+        {isToggleable
+          && showToggler
+          && <Button
+            icon={isOpen ? 'fa-minus' : 'fa-plus'}
+            onClick={toggleOpenState}
+            title={isOpen ? 'Weitere Informationen verbergen' : 'Weitere Informationen anzeigen'}
```
MethodenMann: Please no hard coded text resources. Use props to inject this texts (as i did with RemoteField for example).
Fallback text can be hardcoded but should be in english (e.g. "More" and "Less")

packages/tocco-ui/src/Panel/Panel.js
niklaushug: Yes, because of babel. I do that according a react course I took.

* https://stackoverflow.com/questions/37788342/is-it-better-to-define-state-in-constructor-or-using-property-initializers
* https://stackoverflow.com/questions/43546575/react-defining-state-without-a-constructor?rq=1
* https://stackoverflow.com/questions/42993989/init-state-without-constructor-in-react

niklaushug: I am open for coding style.

packages/tocco-ui/src/Panel/Panel.js
niklaushug: I will use conditional statements over short circuit to respect ESLint.

packages/tocco-ui/src/Panel/Panel.js
niklaushug: Yes and ok.

packages/tocco-ui/src/Panel/Panel.js
niklaushug: ok

packages/tocco-ui/src/Panel/Panel.js
MethodenMann: ok. just didnt know this notation

packages/tocco-ui/src/Panel/Panel.js
niklaushug: see 88e5dbe45b2124693323e37e12e97f459bcf44ec

packages/tocco-ui/src/Panel/Panel.js
niklaushug: see 88e5dbe45b2124693323e37e12e97f459bcf44ec

packages/tocco-ui/src/Panel/Panel.js
niklaushug: see 88e5dbe45b2124693323e37e12e97f459bcf44ec

packages/tocco-ui/src/Panel/PanelHeaderFooter.js
niklaushug: see 88e5dbe45b2124693323e37e12e97f459bcf44ec

packages/tocco-ui/src/Panel/PanelBody.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  describe('PanelBody', function() {
+    it('should render parent and children', () => {
+      const wrapper = mount(<PanelBody><span>text-1</span><span>text-2</span></PanelBody>)
+      expect(wrapper.find('div')).to.have.length(1)
+      expect(wrapper.find('span')).to.have.length(2)
+      expect(wrapper.find('span').first().text()).to.equal('text-1')
+      expect(wrapper.find('span').last().text()).to.equal('text-2')
+    })
+
+    // it('body should be hidden', () => {
```
MethodenMann: Either uncomment or delete these tests

PR #446, opened 2018-06-18
feat(tocco-ui): display unicode character as Icon

niklaushug:
- allow to pass unicode charcter as prop
- export StyledIcon

### DIFF
--- a/packages/tocco-ui/src/Icon/Icon.js
+++ b/packages/tocco-ui/src/Icon/Icon.js
@@ -11,16 +11,15 @@ import {
 } from '../utilStyles'
 
 const getClassName = (icon, animation) => {
-  let cls = classNames({
-    'glyphicon': icon.startsWith('glyphicon-'),
-    'fa': icon.startsWith('fa-')
-  }, icon, 'icon')
-
-  if (animation === stylingAnimation.SPIN) {
-    cls = classNames(cls, 'fa-spin')
+  if (icon) {
+    return classNames({
+      'glyphicon': icon.startsWith('glyphicon-'),
+      'fa': icon.startsWith('fa-'),
+      'fa-spin': animation === stylingAnimation.SPIN
+    }, icon, 'icon')
+  } else {
+    return 'fa'
   }
-
-  return cls
 }
 
 /**
@@ -37,7 +36,9 @@ const Icon = props => {
       onMouseLeave={props.onMouseLeaveFunction}
       onMouseDown={props.onMouseDownFunction}
       position={props.position}
-    />
+    >
+      {props.unicode}
+    </StyledIcon>
   )
 }
 
@@ -56,7 +57,7 @@ Icon.propTypes = {
    * specific classname (e.g. "bars")
    * https://getbootstrap.com/docs/3.3/components/#glyphicons or https://fontawesome.com/v4.7.0/icons/
    */
-  icon: PropTypes.string.isRequired,
+  icon: PropTypes.string,
   /**
    * If true, button occupies less space. It should only used for crowded areas like tables and only if necessary.
    */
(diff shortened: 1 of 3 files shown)


packages/tocco-ui/src/Icon/Icon.spec.js
```diff
@@ -32,7 +32,7 @@ describe('tocco-ui', function() {
         onMouseDown,
         position
       } = props
-      expect(Object.keys(props)).to.have.lengthOf(7)
+      expect(Object.keys(props)).to.have.lengthOf(8) // children is eighth prop
```
MethodenMann: what is the value of this comment? or even the line. to test for an amount of props does not seam to test anything?

packages/tocco-ui/src/Icon/Icon.js
```diff
@@ -68,7 +74,11 @@ Icon.propTypes = {
   onClickFunction: PropTypes.func,
   onMouseEnterFunction: PropTypes.func,
   onMouseLeaveFunction: PropTypes.func,
-  onMouseDownFunction: PropTypes.func
+  onMouseDownFunction: PropTypes.func,
+  unicode: PropTypes.string
```
MethodenMann: Description is necessary here. And if i got that right, its only unicode OR icon that can be set?

packages/tocco-ui/src/Icon/Icon.js
```diff
@@ -11,15 +11,19 @@ import {
 } from '../utilStyles'
 
 const getClassName = (icon, animation) => {
-  let cls = classNames({
-    'glyphicon': icon.startsWith('glyphicon-'),
-    'fa': icon.startsWith('fa-')
-  }, icon, 'icon')
+  let cls
```
MethodenMann: const getClassName = (icon, animation) => {
  if (icon) {
    return classNames({
      'glyphicon': icon.startsWith('glyphicon-'),
      'fa': icon.startsWith('fa-'),
      'fa-spin': animation === stylingAnimation.SPIN
  } else {
    return 'fa'
  }
}

or probably better with the ternary operator instead of the if

packages/tocco-ui/src/Icon/Icon.spec.js
niklaushug: It should be passed 7 props but it is tested for 8 props. To clarify what makes the difference I add a comment. "props.children" is added by react what has confused me. I explain the pitfall to a former me or a forgetting me.

packages/tocco-ui/src/Icon/Icon.js
niklaushug: Yes, it is an exclusive or. Do you have an approach to test an XOR by propTypes. I have only found https://stackoverflow.com/questions/35892278/how-to-declare-react-proptypes-xor but it feels not right.

packages/tocco-ui/src/Icon/Icon.js
```diff
@@ -68,7 +69,15 @@ Icon.propTypes = {
   onClickFunction: PropTypes.func,
   onMouseEnterFunction: PropTypes.func,
   onMouseLeaveFunction: PropTypes.func,
-  onMouseDownFunction: PropTypes.func
+  onMouseDownFunction: PropTypes.func,
+  /*
+   * Display one or more unicode character. Use unicode escape string (e.g. \u2022).
+   * Font Awesome styling is used. If prop icon and unicode is used together, icon prepend unicode charaters.
```
MethodenMann: typo charaters

packages/tocco-ui/src/Icon/Icon.spec.js
MethodenMann: please remove this line entirely. its confusing me more than there is a value to it

niklaushug: ok

packages/tocco-ui/src/Icon/Icon.js
niklaushug: thx

packages/tocco-ui/src/Icon/Icon.js
MethodenMann: Never used it before but i think custom proptype validators should do the job. But i would no use the generic method you mentioned and hard code instead

PR #447, opened 2018-06-18
feat(tocco-ui): establish Typography

niklaushug:
- establish React components for commonly used html tags
- styling inspired by Bootstrap 3.7 and Google Material Design
- extend DefaultTheme by fontFamily.sansSerif, fontFamily.monospace and info signal color
- inject default font into tocco-ui-showcase and appFactory
- establish function to declare font
- describe font size, spacing and headings
- hyphenate or truncate typographic tags
- establish prop breakWords to control word wrap
  - if true words break with hyphens
  - if false text is forced into a single truncated line
  - if false tag has a title attribute with all content
- create helper function to get text from all nested nodes
- create helper function to declare css for word wrap
- use tocco standard font styles


packages/tocco-util/src/appFactory/appFactory.js
```diff
@@ -15,6 +15,7 @@ import it from 'react-intl/locale-data/it'
 
 import {LoadMask} from 'tocco-ui'
 import {ToccoTheme} from 'tocco-theme'
+import InjectFontRoboto from '../../../tocco-ui/src/Typography/Font'
```
MethodenMann: dont use a relative import here

packages/tocco-ui/src/Typography/Font.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import {injectGlobal} from 'styled-components'
+import {themeGet} from 'styled-system'
+
+const InjectFontRoboto = props => {
+  const fontFamilySansSerif = themeGet('fontFamily.sansSerif', 'false')(props)
+  const fontFamilyMonospace = themeGet('fontFamily.monospace', 'false')(props)
+
+  if (fontFamilySansSerif.includes('Roboto')) {
+    // eslint-disable-next-line
```
MethodenMann: why is the eslint-disable needed? As mentioned before please try to fix it or edit eslint config. just ignoring a line should be the last option

packages/tocco-ui/src/Typography/Heading.js
```diff
@@ -0,0 +1,146 @@
...
+  )
+}
+
+H5.defaultProps = {
+  breakWords: true,
+  styledLike: 'H5'
+}
+
+const H6 = props => {
+  return (
```
MethodenMann: returns can be omitted

packages/tocco-ui/src/Typography/Misc.js
```diff
@@ -0,0 +1,271 @@
...
+  return (
+    <StyledPre
+      breakWords={props.breakWords}
+      title={props.breakWords ? undefined : getTextOfChildren(props.children)}
+    >{props.children}</StyledPre>
+  )
+}
+
+const S = props => {
+  return (
```
MethodenMann: returns can be omitted

packages/tocco-ui/src/Typography/StyledList.js
```diff
@@ -0,0 +1,69 @@
...
+`
+
+const StyledList = styled.ol`
+  && {
+    display: block;
+    list-style-position: outside;
+    margin: 0 0 ${props => theme('space.5')} 1.6rem;
+    padding: 0;
+
+    & & {
```
MethodenMann: same as && ?

packages/tocco-ui/src/Typography/Typography.js
```diff
@@ -0,0 +1,36 @@
...
+import {StyledSpan} from './StyledTypography'
+import {getTextOfChildren} from '../utilStyles'
+
+/**
+ * Utilize only React components (e.g. <Span>) instead of pure html tags (e.g. <span>)
+ * ensure coherent typography and prevent css leaking. Default Fonts are loaded
+ * automatically. When other fonts are defined it is required to load font files
+ * manually and to define corresponding @font-face.
+ */
+const Span = props => {
```
MethodenMann: Component name does no match file name

packages/tocco-ui/src/Typography/Font.js
MethodenMann: or at lease disable specific rules

packages/tocco-ui/src/Typography/StyledList.js
niklaushug: '&&' is not equal '& &'
* '&&' => control specificity (https://www.styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity)
* '& &' => control nesting  (not exactly https://www.styled-components.com/docs/advanced#referring-to-other-components but has the same goal)

packages/tocco-ui/src/Typography/Typography.js
niklaushug: This is a workaround. 

My first approach was to declare Span in Misc.js but I need a file which has the same name like the package to be able to load example file.

My second approach was to declare everything from Misc.js in Typography.js but it is only allowed to have one component in main file (if I remember right).

Both approaches should be related to tocco-ui-showcase implementation. 

I am open for other solutions. Maybe it is the way to code currently.

packages/tocco-ui/src/Typography/Font.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import {injectGlobal} from 'styled-components'
+import {themeGet} from 'styled-system'
+
+const InjectFontRoboto = props => {
+  const fontFamilySansSerif = themeGet('fontFamily.sansSerif', 'false')(props)
+  const fontFamilyMonospace = themeGet('fontFamily.monospace', 'false')(props)
+
+  if (fontFamilySansSerif.includes('Roboto')) {
+    // eslint-disable-next-line chai-friendly/no-unused-expressions
```
MethodenMann: https://eslint.org/docs/rules/no-unused-expressions

take a look at allowTaggedTemplates. should fix this without eslint-disable

niklaushug: Declaring "chai-friendly/no-unused-expressions": [2, { "allowTaggedTemplates": true }] in /.eslintrc would make disabling this specific rule dispensable. Whereas currently linting is disabled for only one line, we are about to allow Tagged Template Literals globally without assignment.

injectGlobal is used rarely. Tagged Template Literals are normally nested in functions or are assigned to variable, which does not trigger any linting rule. Hence I prefer the current implementation (being explicit).

packages/tocco-ui/src/Typography/Heading.js
```diff
@@ -0,0 +1,128 @@
...
+            /**
+             * If true words break with hyphens. If false text is forced into a single truncated line.
+             */
+            breakWords: PropTypes.bool,
+            children: PropTypes.node.isRequired,
+            /**
+             * Control size and space independently from semantic meaning.
+             */
+            styledLike: PropTypes.oneOf(['H1', 'H2', 'H3', 'H4', 'H5', 'H6'])
+          }
```
rzueger: (maybe a little less indentation?)

packages/tocco-ui/src/Typography/Misc.js
```diff
@@ -0,0 +1,223 @@
...
+= Pre.defaultProps
+= Q.defaultProps
+= S.defaultProps
+= Strong.defaultProps
+= Sub.defaultProps
+= Sup.defaultProps
+= U.defaultProps
+= Var.defaultProps = {
+                                    breakWords: true
+                                  }
```
rzueger: (doesn't look very nice. lines 185-191 neither)

packages/tocco-ui-showcase/src/components/ShowCaseApp.js
```diff
@@ -38,6 +39,7 @@ export default class ShowCaseApp extends React.Component {
       <ThemeProvider theme={ToccoTheme}>
         <IntlProvider locale={this.state.locale}>
           <div className="show-case-app tocco-ui-theme">
+            <InjectFontRoboto theme={ToccoTheme}/>
```
rzueger: Is there a more generic way to inject the required font files? maybe by creating a higher order component for ThemeProvider, do everything there, and just use this extended ThemeProvider everywhere instead of the original one.
Inserting this `InjectFontRoboto` tag everywhere is something that tends to be forgotten easily. Furthermore, what happens if we decide to use a different font? Right now we would have to replace this element everywhere.

I don't mind whether we refactor this in this pull request or in a following change, but I think we should rethink this approach. What do you think?

packages/tocco-ui/src/Typography/Heading.js
niklaushug: This is unwanted and caused by `yarn lint:fix`. Was checking linting rules but could not find a configuration in https://eslint.org/docs/rules/indent or https://eslint.org/docs/rules/no-multi-assign. 

Possible solutions are:
* Assignment in one line which reduces readability.  --> `H1.propTypes = H2.propTypes = H3.propTypes = H4.propTypes = H5.propTypes = H6.propTypes = {`  
* `/* eslint-disable indent */` and `/* eslint-enable indent */`

packages/tocco-ui/src/Typography/Misc.js
niklaushug: see above

packages/tocco-ui-showcase/src/components/ShowCaseApp.js
niklaushug: I agree, this solution is not flexible enough and adds unnecessarily tags. It does not provide options to add a dev key for paid font service or font files. 

Adding two times the component itself seems okay to me. I am sure better solution would exist. I'd like to discuss this issue and I'm happy about some advice.

@MethodenMann Is it okay to add a task to your icebox?

niklaushug: Pivotal Story ID 158457342

PR #451, opened 2018-06-19
feat(tocco-ui): establish SignalList

niklaushug:
- signalize conditions by icon, color and label in a list
- define singal colors
- define styling conditions
- utilize typographic components for SignalList

### DIFF
--- a/packages/tocco-ui/src/SignalList/SignalListItem.js
+++ b/packages/tocco-ui/src/SignalList/SignalListItem.js
@@ -1,0 +1,59 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import Icon from '../Icon'
+import StyledSignalListItem from './StyledSignalListItem'
+import {
+  conditionPropTypes,
+  stylingPosition,
+  stylingCondition
+} from '../utilStyles'
+
+const getIcon = props => {
+  switch (props.condition) {
+    case stylingCondition.DANGER:
+      return {icon: 'fa-times'}
+    case stylingCondition.SUCCESS:
+      return {icon: 'fa-check'}
+    case stylingCondition.WARNING:
+      return {icon: 'fa-exclamation-triangle'}
+    case stylingCondition.PRIMARY:
+    case stylingCondition.Base:
+      return {unicode: '\u2022'} // unicode bullet point
+  }
+}
+
+/**
+ * Signalize single condition by icon and color. It must be wrapped by <SignalList/>
+ */
+const SignalListItem = props => {
+  return (
+    <StyledSignalListItem condition={props.condition}>
+      <Icon {...getIcon(props)} position={stylingPosition.BEFORE}/>
+      {props.label}
+      { // eslint-disable-next-line
+        React.Children.map(props.children, child => {
+          return React.cloneElement(child)
+        })}
+    </StyledSignalListItem>
+  )
+}
+
+SignalListItem.defaultProps = {
+  label: '',
+  condition: stylingCondition.Base
+}
+
+SignalListItem.propTypes = {
+  /**
+   * Visible text. Default is an empty string.
+   */
+  label: PropTypes.string.isRequired,
+  /**
+   * Color and icon is set according condition. Default value is 'base'.
+   * Possible values: base|danger|primary|success|warning
+   */
+  condition: conditionPropTypes
+}
+
+export default SignalListItem
--- a/packages/tocco-ui/src/SignalList/SignalListItem.js
+++ b/packages/tocco-ui/src/SignalList/SignalListItem.js
@@ -20,6 +20,10 @@ const getIcon = props => {
     case stylingCondition.PRIMARY:
     case stylingCondition.BASE:
       return {unicode: '\u2022'} // unicode bullet point
+    default:
+      // eslint-disable-next-line no-console
+      console.warn('Be explicit by adding a case. You may want display a different character or an icon.')
+      return {unicode: '\u2022'} // unicode bullet point
   }
 }
 
--- a/packages/tocco-ui/src/SignalList/SignalListItem.js
+++ b/packages/tocco-ui/src/SignalList/SignalListItem.js
@@ -18,7 +18,7 @@ const getIcon = props => {
     case stylingCondition.WARNING:
       return {icon: 'fa-exclamation-triangle'}
     case stylingCondition.PRIMARY:
-    case stylingCondition.Base:
+    case stylingCondition.BASE:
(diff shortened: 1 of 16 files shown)


packages/tocco-ui/src/SignalList/SignalListItem.js
```diff
@@ -0,0 +1,53 @@
...
+  switch (props.condition) {
+    case stylingCondition.DANGER:
+      return {icon: 'fa-times'}
+    case stylingCondition.SUCCESS:
+      return {icon: 'fa-check'}
+    case stylingCondition.WARNING:
+      return {icon: 'fa-exclamation-triangle'}
+    case stylingCondition.PRIMARY:
+    case stylingCondition.BASE:
+      return {unicode: '\u2022'} // unicode bullet point
```
rzueger: what if a new styling condition item is added at some point? maybe it would make sense to return the unicode bullet point as default?

packages/tocco-ui/src/SignalList/StyledSignalListItem.js
```diff
@@ -0,0 +1,35 @@
...
+  switch (props.condition) {
+    case stylingCondition.DANGER:
+      return theme('colors.signal.danger')(props)
+    case stylingCondition.SUCCESS:
+      return theme('colors.signal.success')(props)
+    case stylingCondition.WARNING:
+      return theme('colors.signal.warning')(props)
+    case stylingCondition.PRIMARY:
+      return theme('colors.primary.line.1')(props)
+    case stylingCondition.Base:
```
rzueger: - `stylingCondition.BASE`?
- and maybe return `'inherit'` also as default?

packages/tocco-ui/src/SignalList/StyledSignalListItem.js
```diff
@@ -0,0 +1,35 @@
+import {theme} from 'styled-system'
+
+import {StyledIcon} from '../Icon'
+import {StyledLi} from '../Typography'
+import {stylingCondition} from '../utilStyles'
+
+const StyledSignalListItem = StyledLi.extend`
+  && {
+  color: ${props => getColor(props)};
+  position: relative;
```
rzueger: (indentation missing for two lines above)

packages/tocco-ui/src/SignalList/SignalListItem.js
rzueger: I share your point of view that it would make sense to consider it here if a new condition is added. However, in my opinion, it's more important to make our code as stable as possible - chances are that we won't notice that a new styling condition has been added until this component looks broken in the production environment, which is certainly not what we want to achieve.

Instead of replacing the `PRIMARY` and `BASE` cases with a `default` block, we could keep them there and add a `default` block where we return the unicode bullet point but also log a warning in the console. This way, the risk of a broken production environment would be reduced, but we would still notice that a new styling condition has been added which should be considered in this component.

packages/tocco-ui/src/SignalList/StyledSignalListItem.js
```diff
@@ -27,7 +27,7 @@ const getColor = props => {
       return theme('colors.signal.warning')(props)
     case stylingCondition.PRIMARY:
       return theme('colors.primary.line.1')(props)
-    case stylingCondition.Base:
+    default:
```
rzueger: Please consider my comment in SignalListItem.js here as well

packages/tocco-ui/src/SignalList/SignalListItem.js
rzueger: I've thought about this a little bit more and I've come up with another idea.

What if we would define the icons in a map instead of returning them in a switch/case. This map could be compared with the prop types when the file is loaded. This way we could log the warning when the file is loaded instead of when the component is rendered (which could be discovered rather late, maybe already in production).

```
const ICONS = {
  [stylingCondition.DANGER]: {icon: 'fa-times'},
  [stylingCondition.SUCCESS]: {icon: 'fa-check'},
  ...
}

requireAllKeys(ICONS, conditionPropTypes)

// can be somewhere in the utils module
function requireAllKeys(map, propType) {
  // compare map with prop type items and log if not the same
}

const getIcon = props => {
  const icon = ICONS[props.stylingCondition]
  return icon || {unicode: '\u2022'}
}

File: packages/tocco-ui/src/utilStyles/doesReuseAllValuesAsKeys.spec.js [hunk 5]
```diff
@@ -0,0 +1,53 @@
...
+        HEAD: 'head'
+      }
+
+      const correct = {
+        shoulder: 's',
+        head: 'hd',
+        heart: 'ht'
+      }
+
+      const toShort = {
```
rzueger: `tooShort`?

packages/tocco-ui/src/utilStyles/doesReuseAllValuesAsKeys.spec.js
```diff
@@ -0,0 +1,53 @@
...
+        shoulder: 's',
+        heart: 'ht'
+      }
+
+      const toLong = {
```
rzueger: `tooLong`?

packages/tocco-ui/src/utilStyles/doesReuseAllValuesAsKeys.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Enforce that all values of a flat object are used as keys in another object.
+ * @param  {object} completeMap            A flat object with contains a full list of
+ *                                         all allowed keys defined as values.
+ * @param  {object} potentialIncompleteMap A flat object which is does not reuse all values as keys.
+ * @return {boolean}                       Return true if all values are reused as keys.
+ */
+
+const doesReuseAllValuesAsKeys = (completeMap, potentialIncompleteMap) => {
```
rzueger: Is there a better name for this function? Maybe `assertObjectKeysMatch` or something like that?

packages/tocco-ui/src/utilStyles/doesReuseAllValuesAsKeys.js
```diff
@@ -0,0 +1,32 @@
...
+      return value === keys[index]
+    }
+  )) {
+    // eslint-disable-next-line no-console
+    console.warn(`At least on identifier does not match (${values.toString()} !== ${keys.toString()}).`)
+    return false
+  }
+
+  return true
+}
```
rzueger: Could we log the specific keys which are missing or should be removed? I reckon it's much more likely that I notice a log warning if a key is mentioned that I removed or added somewhere.

PR #457, opened 2018-06-26
feat(entity-detail): Allow sub grid page limit to be set in model

rzueger:
The previously hardcoded page limit of 5 now acts as a default value.


packages/entity-detail/src/util/detailView/fromFieldFactories/subGrid.js
```diff
@@ -9,5 +9,6 @@ export default type =>
       relationName={formField.id}
       modelField={modelField}
       showSearchForm={formField.showSearchForm}
+      limit={formField.limit || undefined}
```
ottrob: just `limit={formField.limit}`? if limit is undefined its the same.

PR #458, opened 2018-06-27
chore(scheduler): use raw html in calendar event popover

ottrob:
the description property of a calendar event now will return html
instead of text. to correctly apply the html to the popover the
`dangerouslySetInnerHtml` property for React components is used.

the popover content has to be split because in raw html code the react
component `Conflict` will not be recognized otherwise.

### DIFF
--- a/packages/scheduler/src/components/FullCalendar/FullCalendar.js
+++ b/packages/scheduler/src/components/FullCalendar/FullCalendar.js
@@ -68,16 +68,15 @@ class FullCalendar extends React.Component {
     },
     eventClick: event => this.props.onEventClick(event),
     eventRender: (event, element) => {
-      const content = <div>
-        <p>{event.description}</p>
+      const reservation = event.start.twix(event.end).format({monthFormat: 'MMMM', dayFormat: 'Do'})
+      const tooltipContent1 = <p dangerouslySetInnerHTML={{__html: `
+        <p>${event.description}</p>
         <p>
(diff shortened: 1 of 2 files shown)


packages/scheduler/src/components/FullCalendar/FullCalendar.scss
```diff
@@ -40,3 +40,14 @@
     text-overflow: ellipsis;
   }
 }
+
+.tocco-resource-scheduler {
+  .popover{
```
rzueger: (space before `{`)

packages/scheduler/src/components/FullCalendar/FullCalendar.js
```diff
@@ -68,16 +68,15 @@ class FullCalendar extends React.Component {
...
-        </p>
-        <p>
-          <Conflict conflictStatus={event.conflict} intl={this.props.intl}/>
-        </p>
-      </div>
+          <i class="fa fa-clock-o" style="padding-right:10px" aria-hidden="true"></i>${reservation}
+        </p>`
+      }} />
+
+      const tooltipContent2 = <p><Conflict conflictStatus={event.conflict} intl={this.props.intl}/></p>
```
rzueger: are there better names for these variables than `tooltipContent1` and `tooltipContent2`? or maybe put all elements in an array and then build the content like this:

```
elements.map(element => ReactDOMServer.renderToString(element)).join('')
```

PR #462, opened 2018-07-04
Pr/styled components FormattedValues

niklaushug:
@MethodenMann Have a close look on changes in DocumentFormatter.js and DocumentCompactFormatter of commit f087c90. You have adapted those formatters in #440.

Since formatted values are readable only `<Preview>` is a better choice than `<Upload>` and plain html. Probably it makes sense to remove prop.readOnly from `<Upload>` to enforce usage of `<Preview>`. Note that `<Preview>` provides a link for download instead of a button, which might be a usability issue.

**Solution:** ButtonLink is used 76543c98fb56199c0db13301aded0f7c9910a4e5

### DIFF
--- a/packages/tocco-ui/src/FormattedValue/typeFormatters/TimeFormatter.js
+++ b/packages/tocco-ui/src/FormattedValue/typeFormatters/TimeFormatter.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import {FormattedTime, injectIntl, intlShape} from 'react-intl'
+import moment from 'moment'
 
 import {Time} from '../../Typography'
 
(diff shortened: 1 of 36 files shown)


packages/tocco-ui/src/FormattedValue/typeFormatters/TimeFormatter.js
```diff
@@ -10,15 +12,24 @@ const TimeFormatter = props => {
 
   const date = new Date(2000, 1, 1, hours, minutes, seconds, milliSeconds)
 
+  const twoDigits = n => { return String('00' + n).slice(-2) }
```
MethodenMann: Please do the following: 
- Add the newest version of moment js (yarn add moment)
- use moment js format to get iso format
    const timeIso = moment(date).format(moment.HTML5_FMT.TIME_MS)

PR #467, opened 2018-07-10
style(tocco-ui): limit maximal dimension of Preview image

niklaushug:
I have implemented maximal dimension as parameter because not always is such a limitation desirable or should be used equally. Unit 'vw' and 'vh' enables responsiveness to some limitations. 

Current implementation corresponds with scale only but not with a crop and scale. A crop and scale effect would be possible with css property background-image and an additional element but needs a lot more logic. 

Personally I question an automatic crop function in an editor centric view anyway. Crop is a good choice to assures a nice aligned layout in a multi image read only view but not to administer raw data.

### DIFF
--- a/packages/tocco-ui/src/Preview/Preview.js
+++ b/packages/tocco-ui/src/Preview/Preview.js
@@ -5,6 +5,7 @@ import Icon from '../Icon'
 import Link from '../Link'
 import {Figcaption} from '../Typography'
 import StyledPreview from './StyledPreview'
+import {validateCssDimension} from '../utilStyles'
 
 /**
  * Use <Preview> to display a preview of any kind of file. Provide URLs to thumbnail and file.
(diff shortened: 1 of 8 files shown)


packages/tocco-ui/src/Preview/Preview.js
```diff
@@ -72,6 +76,20 @@ Preview.propTypes = {
    * thumbnailLink as arguments. onClick overrules downloadOnClick.
    */
   onClick: PropTypes.func,
+  maxDimensionX: function(props, propName, componentName) {
+    if (props[propName] !== undefined && !/^[1-9][0-9]*(px|em|rem|fr|%|vh|vw|vmin|vmax)$/.test(props[propName])) {
```
MethodenMann: i would create a validate css dimension propType function and use it here.

packages/tocco-ui/src/Preview/Preview.js
```diff
@@ -72,6 +76,20 @@ Preview.propTypes = {
    * thumbnailLink as arguments. onClick overrules downloadOnClick.
    */
   onClick: PropTypes.func,
+  maxDimensionX: function(props, propName, componentName) {
```
MethodenMann: comment missing

packages/tocco-ui/src/utilStyles/propTypesValidator.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Validate if valid css dimension as PropTypes.
+ * @param  {object} props
+ * @param  {string} propName
+ * @param  {string} componentName
+ * @return {object or null}
+ */
+const validateCssDimension = (props, propName, componentName) => {
```
MethodenMann: export const validateCssDimension...
and remove export at file bottom

packages/tocco-ui/src/utilStyles/propTypesValidator.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      ]
+
+      it('should be valid dimension', () => {
+        valid.map(value => {
+          props['dimension'] = value
+          expect(validateCssDimension(props, 'dimension')).to.be.null
+        })
+      })
+
+      it('should be invalid dimension', () => {
```
MethodenMann: nice tests 👍

PR #476, opened 2018-07-24
Pr/entity list selection

rzueger:


packages/entity-list/src/modules/list/sagas.spec.js
```diff
@@ -305,6 +305,38 @@ describe('entity-list', () => {
           })
         })
 
+        describe('onRowClick saga', () => {
+          it('should not change selection if selectOnRowClick not true', () => {
+            const gen = sagas.onRowClick(actions.onRowClick('1'))
```
MethodenMann: you could take a look at https://github.com/jfairbank/redux-saga-test-plan
i used it a couple of times and i think it enables better redux-saga testing.

packages/entity-list/README.md
```diff
@@ -29,3 +29,4 @@ React-registry name: `entity-list`
 |---------------------|-------------------------------|-------------
 | `onRowClick`        | `id` (The id of the record)   | This event is fired when a list row is clicked
 | `navigateToCreate`  | -                             | This event is fired when the "new" button is clicked
+| `onSelectChange`    | `keys` (the ids of the records which got selected or deselected), `isSelected` (`true` if the records got selected, `false if they got deselected) | This event is fired when the selection changes
```
MethodenMann: not sure if the params are correct. isnt it only the array of selected ids?

PR #479, opened 2018-07-27
Pr/styled components tocco logos as loading icon

niklaushug:


packages/tocco-ui/src/utilStyles/declareInteractionColors.js
```diff
@@ -1,21 +1,21 @@
 import {theme} from 'styled-system'
 
-const declareInteractionColors = colors => {
+const declareInteractionColors = (colors, format = 'html') => {
```
MethodenMann: why dont you create a const with this
   ${format === 'html' ? 'background-color' : 'fill'}:
and use it instead of using the ternary operator on every line. the code should get better readable.


also ..can you think of something better than this magic string "html"? Maybe move to a enum like object oder just rename the parameter to isHtml?

packages/tocco-ui/src/IconTocco/IconTocco.js
```diff
@@ -0,0 +1,60 @@
...
+import {
+  inkPropTypes,
+  positionPropTypes,
+  stylingInk,
+  stylingLook,
+  stylingPosition
+} from '../utilStyles'
+
+const IconTocco = props => {
+  return (
```
MethodenMann:
```
const IconTocco = props =>
<StyledIconToccoWrapper
....
```

packages/tocco-ui/src/IconTocco/StyledIconTocco.js
```diff
@@ -0,0 +1,132 @@
+import styled, {keyframes} from 'styled-components'
```
MethodenMann: the main problem i see with this new loading image is that sometimes we use the loading image multiple times in a row on the same position (loading of entity browser and then loading of list , which is a part of entity browser).
When it gets shown the logo looks like one loading mask with a laggy animation. But what really is shown are two independent loading masks and the second one starts the animation from the beginning... But i think there is nothing we can do about that?

packages/tocco-ui/src/utilStyles/declareInteractionColors.js
niklaushug: Possible formats are "html" and "svg". I moved format definition to stylingConstants and introduced tenary operator fillProperty and strokeProperty.

packages/tocco-ui/src/IconTocco/IconTocco.js
niklaushug: What is the issue?

PR #481, opened 2018-07-30
feat(tocco-util): allow multiple modals

MethodenMann:
- first simple solution to show a modal over a modal. this is used
  for example to prompt an advancedsearch modal over an action form data modal.


packages/tocco-util/src/notifier/modules/modalComponents/ModalDisplay.js
```diff
@@ -5,20 +5,22 @@ import ModalContent from './ModalContent'
...
+          close={props.close}
+        />
+      )}
     </div>
   )
 }
 
 ModalDisplay.propTypes = {
-  modal: propTypes.object,
+  modals: propTypes.array,
```
rzueger: arrayOf(shape(...))?

PR #486, opened 2018-08-03
Pr/fulltext search form

MethodenMann:


packages/entity-list/src/modules/searchForm/sagas.js
```diff
@@ -89,7 +89,11 @@ export function* submitSearchFrom() {
 
 export function* loadSearchForm(searchFormName) {
   const formDefinition = yield call(fetchForm, searchFormName, searchFormTransformer)
-  yield put(actions.setFormDefinition(formDefinition))
+  if (formDefinition !== null) {
```
rzueger: I think `if (formDefinition) { ...` would be more stable (cover undefined as well)

PR #496, opened 2018-08-21
Pr/travis fix

MethodenMann:


packages/tocco-ui/src/SearchBox/SearchBox.spec.js
```diff
@@ -5,7 +5,7 @@ import {shallow} from 'enzyme'
 
 describe('tocco-ui', function() {
   describe('SearchBox', function() {
-    const DEFAULT_DEBOUNCE = 200
+    const DEFAULT_DEBOUNCE = 200 * 2
```
rzueger: why not "400" instead of "200 * 2"?

PR #497, opened 2018-08-23
Pr/styled components popover

niklaushug:


packages/tocco-ui/src/Popover/StyledPopover.js
```diff
@@ -0,0 +1,66 @@
+import styled from 'styled-components'
+
+import {declareTypograhpy} from '../Typography'
+import {theme} from 'styled-system'
+
+const borderWidth = 1
```
MethodenMann: constants like this could be written with constant-case:
BORDER_WIDTH

its a little bit hard to draw the line there since almost every var is a const but i usually use constant-case for setting const

packages/tocco-ui/src/Popover/Popover.js
```diff
@@ -65,22 +92,38 @@ class Popover extends React.Component {
...
+  isPlainHtml: PropTypes.boolean,
+  /**
+   * Minimal distance between popper and viewport edge in pixels. Default is '10'.
+   */
+  spacer: PropTypes.number,
   /**
    * Content of the popover
    */
-  placement: PropTypes.string
+  placement: PropTypes.oneOf(Object.values(placements))
```
MethodenMann: 👍

packages/tocco-ui/src/Popover/Popover.js
```diff
@@ -65,22 +92,38 @@ class Popover extends React.Component {
...
   content: PropTypes.node,
   /**
-   * Content of the popover
+   * Reference to the popover.
    */
   children: PropTypes.node,
+  /**
+   * Remove space between content and border if useful (e.g. content is an image only). Default is {false}.
+   */
+  isRimless: PropTypes.boolean,
```
MethodenMann: would it be more consistent if you name this just "rimless"

dense of <Button> also is not called isDense

PR #508, opened 2018-09-03
feat(tocco-ui): fix redux form

khanlifts:


packages/tocco-util/src/form/formBuilder.js
```diff
@@ -125,6 +125,7 @@ export default (
           formFieldUtils={formFieldUtils}
           formFieldMapping={formFieldMapping}
           readOnlyFormFieldMapping={readOnlyFormFieldMapping}
+          format={null}
```
MethodenMann: As mentioned in https://github.com/facebook/react/releases/tag/v15.0.0:

> React 15 warns you on a null input value and offers you to clarify your intention. To fix the warning, you may explicitly pass an empty string to clear a controlled input, or pass undefined to make the input uncontrolled.

So we have to make sure, that null is never passed to the input fields value. To test this you could change the editableValues example.js and set all values to null an see if any warnings are shown in the console.

PR #515, opened 2018-09-12
Pr/styled components toastr phase 1

niklaushug:
Defined goals
- migrate third-party-integration.scss to styled-components 
- remove third-party-integration.scss
- do not touch import ‘!style-loader!css-loader!react-redux-toastr/lib/css/react-redux-toastr.min.css’ 
- utilize DefaultTheme.js
- bounce in and fade out modal component
- bounce in blocker Info

Things for consideration but not relevant for this PR
- Error logger register ReduxToastr separately which leads to overlapping Toastr boxes and rise the need to import StyledNotifier in showcase app for proper styling. 
- By using toastrActionsr.remove(id) directly fade out animation is omited. This leads to inconsistent behaviour. Commit b36b8ff42d6cd1ae26430b9dbdc0fb1cc594df4b introduces a work around which could be used for blocker info as well.


packages/tocco-ui-showcase/src/toccoUtilExamples/notifier/example.js
```diff
@@ -19,11 +19,12 @@ class Example extends React.Component {
...
 
   info = () => {
-    this.store.dispatch(notifier.info('info', 'client.title', 'client.description', 'info', 2000))
+    this.store.dispatch(notifier.info(
+      'info', <Typography.H4>Typography component</Typography.H4>, 'string contains <b>html</b>', 'info', 2000))
   }
 
   success = () => {
-    this.store.dispatch(notifier.info('success', 'client.title', 'client.description', 'thumbs-up', 0))
+    this.store.dispatch(notifier.info('success', 'client.title', 'Lorem Ipsum', 'thumbs-up', 2000))
```
MethodenMann: Why did you change the timeout to 2000? With 0 the toastr stays which could be used in some cases. If you think this should be the same every time we should rather remove that property.

packages/tocco-ui-showcase/src/toccoUtilExamples/notifier/example.js
```diff
@@ -77,21 +78,24 @@ class Example extends React.Component {
...
-            label="primary action"
-            onClick={props.close}
-          />
-          <Button
-            label="secondary action"
-            onClick={props.close}
-          />
-        </div>
-      )
+        <React.Fragment>
```
MethodenMann: Have you tried if  the Fragment shorthand works in our setup:

    <>
      <ChildA />
    </>

packages/tocco-util/src/notifier/components/Notifier.js
```diff
@@ -4,13 +4,16 @@ import ReduxToastr from 'react-redux-toastr'
...
+import {StyledNotifier} from './StyledNotifier'
 
 const Notifier = props => {
   return (
-    <div className="tocco-notifier">
-      <ReduxToastr {...props.toastrOptions} />
-      <ModalDisplayContainer/>
-    </div>
+    <StyledNotifier>
+      <div className="tocco-notifier">
```
MethodenMann: Why is tocco-notifier still needed with styled-components?

packages/tocco-util/src/notifier/components/TitleMessage.js
```diff
@@ -1,32 +1,41 @@
...
 
+const containsHtml = s => s && /<\/?[a-zA-Z0-9]*\/?>/.test(s)
 const isKey = s => s && s.startsWith('client.')
 
-const Content = ({content}) =>
-  _isString(content)
-    ? (isKey(content) ? <FormattedMessage id={content}/> : <FormattedValue type="html" value={content}/>)
-    : content
+/**
+ * Format content correctly
```
MethodenMann: We use propTypes to describe props

packages/tocco-util/src/notifier/components/TitleMessage.js
```diff
@@ -1,32 +1,41 @@
...
-    <div className={className}>
-      <header className="tocco-notifier__title">
-        <Content content={title}/>
-      </header>
-      {message
-      && <div className="tocco-notifier__message">
-        <Content content={message}/>
-      </div>}
+    <React.Fragment>
+      {title && <Content content={title} isTitle={true} />}
```
MethodenMann: Content "is Title"? Rather confusing. IMO it would be better to introduce e Title component than creating a component that is kind of both.

I think is considered bad practice to reasigne a property (in this case content)

packages/tocco-util/src/notifier/notificationActionFactory.js
```diff
@@ -61,16 +60,12 @@ export function getBlockingInfo(id, title, message, icon) {
...
-    transitionOut: 'bounceOutUp',
     attention: true,
     onAttentionClick: () => {},
     component: () => <TitleMessage title={title} message={message}/>
   }
 
-  if (icon) {
-    options.icon = <Icon icon={icon} size="2x" />
-  }
+  options.icon = _isString(icon) ? <Icon icon={`${icon}`} size="3x" /> : icon
```
MethodenMann: icon={icon} ?

MethodenMann: What if icon is undefined? Does that work?

packages/tocco-ui-showcase/src/toccoUtilExamples/notifier/example.js
niklaushug: So far I know that babel-loader throws a SyntaxError and Sublime's syntax highlighting fails.

packages/tocco-util/src/notifier/components/TitleMessage.js
niklaushug: I have replaced `props.isTitle` by `props.tag` to make component `Content` more versatile. Having two components which shares complete codebase next to a single tag (e.g. <h4> or <p>) seems awkward to me.

I have introduced variable output.

packages/tocco-util/src/notifier/notificationActionFactory.js
niklaushug: If icon is undefined Toastr's default icons are used. Instead of leaving options.icon undefined I set a default icon.

packages/tocco-util/src/notifier/components/TitleMessage.js
```diff
@@ -1,40 +1,67 @@
...
+
+  if (_isString(content)) {
+    if (containsHtml(content)) {
+      output = <FormattedValue type="html" value={content}/>
+    } else {
+      output = isKey(content) ? <FormattedMessage id={content}/> : content
+      output = <props.tag>{output}</props.tag>
+    }
+  }
+  return output
```
MethodenMann: as mentioned in another PR please do rather create a new const than reassign a let var and give proper names.

packages/tocco-util/src/notifier/modules/modalComponents/ModalContent.js
```diff
@@ -1,35 +1,43 @@
...
+    const {
+      closable,
+      message,
+      title
+    } = this.props
+
+    return (
+      <div className="rrt-confirm-holder">
+        <StyledModelContent isClosing={this.state.isClosing}>
+          {title && <Content content={title} tag={Typography.H4} />}
```
MethodenMann: Why are you not using `<TitleMessage>` directly?

packages/tocco-util/src/notifier/notificationActionFactory.js
```diff
@@ -61,16 +60,12 @@ export function getBlockingInfo(id, title, message, icon) {
...
-    transitionOut: 'bounceOutUp',
     attention: true,
     onAttentionClick: () => {},
     component: () => <TitleMessage title={title} message={message}/>
   }
 
-  if (icon) {
-    options.icon = <Icon icon={icon} size="2x" />
-  }
+  options.icon = _isString(icon) ? <Icon icon={`${icon}`} size="3x" /> : <Icon icon="exclamation-triangle" size="3x" />
```
MethodenMann: why do you use a template strings for icon?

and why are you using a ternary operator like this in the first place and not:

  options.icon =<Icon icon={ _isString(icon) ? icon : 'excelmation-triangle'} size="3x" />

packages/tocco-util/src/notifier/notificationActionFactory.js
```diff
@@ -11,9 +12,7 @@ export function getInfoAction(type, title, message, icon, timeOut) {
...
```
MethodenMann: same as below applies here

PR #523, opened 2018-09-19
Pr/report

MethodenMann:


packages/tocco-util/src/actions/utils/report.js
```diff
@@ -0,0 +1,115 @@
+import PropTypes from 'prop-types'
+
+const GROUP_GENERAL = 'generalSettings'
+const GROUP_RECIPIENT = 'recipientSettings'
+
+export const getGroupedValues = (settingsDefinition, values) => {
+  const groupValues = name => {
+    return {
```
rzueger: superfluous return. line 10 again.

but that's fine with me - I guess readability is better this way?

packages/tocco-util/src/actions/utils/report.js
```diff
@@ -0,0 +1,115 @@
...
+        'id': 'recipientSettings',
+        'label': msg('client.common.report.recipientSettings'),
+        'layoutType': 'vertical-box',
+        children: extractFields(GROUP_RECIPIENT)
+      }
+    ]
+  }
+}
+
+export const reportSettingsDefinitionPropType = PropTypes.shape({
```
rzueger: generalSettings, recipient or description required?

MethodenMann: isn't it your definition that they can be null?

rzueger: (Well, generalSettings will never be null probably.)
I just wanted to make sure that you're aware that they're not marked as required and that you're code is prepared to handle null values. Seems like that's the case. 👍

packages/tocco-util/src/actions/modules/actionHandlers/report.js
```diff
@@ -0,0 +1,36 @@
...
+        settingsDefinition={settingsDefinition}
+      />
+    ),
+    true
+  ))
+
+  const response = yield take(answerChannel)
+
+  // eslint-disable-next-line no-console
+  console.log('>>>>>response', response, entity, ids)
```
rzueger: can be removed?

MethodenMann: you are right, not really good to push this into master. I will implement report handling in this spring and merge it into this branch. Then we can merge everything together into master.
Will close this PR for now. thanks for reviewing

PR #527, opened 2018-09-21
refactor(scheduler): add StyledFullCalendar

khanlifts:


packages/scheduler/src/components/FullCalendar/FullCalendar.js
```diff
@@ -147,7 +148,9 @@ class FullCalendar extends React.Component {
   }
 
   render = () =>
-    <div className={this.state.wrapperId}><div ref={ref => { this.calendarContainer = ref }}></div></div>
```
niklaushug: What was the purpose of `className={this.state.wrapperId}`? There is no variable className in CSS, hence it can not be related with styling. Removing this code fragment does potential break the application. May jQuery plugin FullCalendar relys on it.

niklaushug: The `className={this.state.wrapperId}` is the reference for the popovers. Those popovers show details while hovering a single event (the colored bars in the calendar). See line 85 in FullCalendar.js.

Do not remove `className={this.state.wrapperId}`.

PR #531, opened 2018-09-25
refactor(tocco-util): use modalComponent for customized notifier

niklaushug:
- use modalComponent for getConfirmationAction and getYesNoAction instead of toastrActions.showConfirm
- show usage of plain string, html string and node in examples of confirmQuestion and yesNoQuestion
- establish modalComponentsTemplates OkCancelButtons and YesNoCancelButtons


packages/tocco-ui-showcase/src/toccoUtilExamples/notifier/example.js
```diff
@@ -38,24 +38,29 @@ class Example extends React.Component {
...
     this.store.dispatch(notifier.yesNoQuestion(
-      'Title',
-      'Message Line 1<br/>Message Line 2',
-      'Yes Text',
-      'No Text',
-      'Cancel Text',
-      () => console.log('yes pressed'),
-      () => console.log('no pressed'),
-      () => console.log('Cancel pressed')
+      <Typography.H5>Title</Typography.H5>,
```
MethodenMann: Does it really make sense to pass a react node here? Shouldn't a title like the same all the time, hence only string (or resource id ) is allowed?

packages/tocco-util/src/notifier/modules/modalComponentsTemplates/YesNoCancelButtons.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
```
MethodenMann: Why are YesNoCancel and OkCancelButtons two components? 
Cant you make it more generic. Maybe even call if ButtonBar that accepts a list of buttons containing label and callback (and primary)?

packages/tocco-ui-showcase/src/toccoUtilExamples/notifier/example.js
niklaushug: TitleMessage.js handles four types of inputs (resource id, plain text, text containing html and nodes) and process props.title and props.message for all kind of toasters. To ensure identical presentation it would make sense to force resource id and wrap text automatically in certain tags (e.g. <h4> and <p>). I believe this is to tight and can not cope reality. By allowing all four types for all kind of toasters it is flexibel and comprehensible. May example yesNoQuestion should be improved to convey the purpose better (e.g. <Typography.H4>Title <Typography.Small>Byline</Typography.Small></Typography.H4>).

packages/tocco-util/src/notifier/modules/modalComponentsTemplates/YesNoCancelButtons.js
niklaushug: By creating one template for each type it is more comprehensible but prone to bad implementation of corporate design. As long we have only two templates it is easy to handle it. May it make sense to refactor out single buttons (e.g. YesButton, NoButton, CancelButton) and use them in both templates?

ButtonGroup is the better choice than ButtonBar because it is a flat list of buttons and not a hierarchically list.

PR #535, opened 2018-09-25
chore(tocco-client): add intl polyfill to jest

khanlifts:


MethodenMann:
please rephrase the commit message. its not "to the tocco-client" but "to jest".
and this is "chore" not "feat".

PR #538, opened 2018-09-26
refactor(tocco-ui-showcase): convert scss to styled-component

niklaushug:


packages/tocco-ui-showcase/src/components/StyledShowCaseApp.js
```diff
@@ -0,0 +1,105 @@
+import styled, {injectGlobal} from 'styled-components'
+
+// eslint-disable-next-line chai-friendly/no-unused-expressions
```
MethodenMann: Please consider every eslint-disable very careful. In this case its a valid expression and no need to disable linting for this line if eslint is configured correctly.

.eslintrc

    "chai-friendly/no-unused-expressions": [2, {"allowTaggedTemplates": true}],

would solve it in this case

packages/tocco-ui-showcase/src/components/StyledShowCaseApp.js
```diff
@@ -0,0 +1,105 @@
...
+injectGlobal`
```
MethodenMann: are you sure about injectGlobal? 
The official website says: 
> We do not encourage the use of this. Try to use it once per app at most, if you must, contained in a single file. This is an escape hatch. Only use it for the rare @font-face definition or body styling.

niklaushug: > definition or **body** styling

`injectGlobal` is the only way to declare styles on `<body>`

Current body styles are obsolete. Hence I have removed it.

MethodenMann:
Could you please commit the eslint configuration proposal of me in another PR.

PR #546, opened 2018-09-27
refactor(tocco-ui): convert EditableValue.scss to styled-component

niklaushug:


packages/tocco-ui/src/EditableValue/StyledEditableValue.js
```diff
@@ -0,0 +1,117 @@
+import styled, {injectGlobal} from 'styled-components'
+import {theme} from 'styled-system'
+
+// eslint-disable-next-line chai-friendly/no-unused-expressions
```
MethodenMann: this PR would be good to adjust eslint config

PR #552, opened 2018-10-02
feat(tocco-util): add FormValidation maxNumber postPoint 

khanlifts:


packages/tocco-util/src/form/validators/model.js
```diff
@@ -56,8 +56,49 @@ export const maxLengthValidator = (value, maxLength) => {
...
+            <FormattedMessage
+              key="syncValidationMaxLength"
+              id="client.component.form.syncValidationPostPoint"
+              values={{limit}}
+            />
+          ]
+        }
+      }
+    }
+    return null
```
MethodenMann: what is the point of this return null

packages/tocco-util/src/form/validators/model.js
```diff
@@ -56,8 +56,48 @@ export const maxLengthValidator = (value, maxLength) => {
   return null
 }
 
+export const maxNumberValidator = (value, maxValue) => {
+  if (value && value > maxValue) {
+    return {
+      maxNumber: [
+        <FormattedMessage
+          key="syncValidationMaxNumber"
+          id="client.component.form.syncValidationMaxNumber"
```
MethodenMann: you need to add these keys to fetchMock textResource mock

PR #561, opened 2018-10-09
fix(tocco-ui): fix phoneformatter async bug

khanlifts:
PhoneFormatter could get empty object and threw an
error on async loading

### DIFF
--- a/packages/tocco-ui/src/FormattedValue/typeFormatters/PhoneFormatter.js
+++ b/packages/tocco-ui/src/FormattedValue/typeFormatters/PhoneFormatter.js
@@ -14,11 +14,19 @@ class PhoneFormatter extends React.Component {
     this.setState({libPhoneImport})
   }
 
+  getFormattedInput = () => {
+    const {libPhoneImport} = this.state
+    const {value} = this.props
+    if (libPhoneImport) {
+      const parsed = libPhoneImport.parseNumber(value)
+      return _isEmpty(parsed) ? value : libPhoneImport.formatNumber(parsed, 'International')
+    } else {
+      return value
+    }
+  }
+
   render() {
-    const libPhoneImport = this.state.libPhoneImport
-    const parsed = libPhoneImport ? libPhoneImport.parseNumber(this.props.value) : this.props.value
-    const formattedNumber = libPhoneImport ? libPhoneImport.formatNumber(parsed, 'International') : parsed
-    return (<span>{_isEmpty(parsed) ? parsed : formattedNumber}</span>)
+    return <span>{this.getFormattedInput()}</span>
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/tocco-ui/src/FormattedValue/typeFormatters/PhoneFormatter.spec.js
```diff
@@ -13,8 +13,8 @@ describe('tocco-ui', () => {
             .then(() => expect(wrapper.text()).to.equal('+41 44 388 60 00'))
         })
 
-        test('should show original string if its not a valid phonenumber', () => {
-          const invalidPhoneNumber = '+123 456'
+        test('should show original string if its not a valid phone number', () => {
+          const invalidPhoneNumber = '+41 4438860011111110'
```
MethodenMann: just curious , what's the motivation behind this change?

packages/tocco-ui/src/FormattedValue/typeFormatters/PhoneFormatter.js
```diff
@@ -16,9 +16,11 @@ class PhoneFormatter extends React.Component {
...
   render() {
     const libPhoneImport = this.state.libPhoneImport
-    const parsed = libPhoneImport ? libPhoneImport.parseNumber(this.props.value) : this.props.value
-    const formattedNumber = libPhoneImport ? libPhoneImport.formatNumber(parsed, 'International') : parsed
-    return (<span>{_isEmpty(parsed) ? parsed : formattedNumber}</span>)
+    const parsed = libPhoneImport ? libPhoneImport.parseNumber(this.props.value, {extended: true}) : this.props.value
+    const validityParsed = parsed.valid ? parsed : this.props.value
+    const notEmptyValue = _isEmpty(validityParsed) ? this.props.value : validityParsed
+    const formattedNumber = libPhoneImport ? libPhoneImport.formatNumber(notEmptyValue, 'International') : notEmptyValue
+    return (<span>{formattedNumber}</span>)
```
MethodenMann: You check 2 times if libPhoneImport is defined. Does it have to be that complicated? Would the old version work with a simple check?

something like:
```
if (libPhoneImport) {
const parsed = libPhoneImport.parseNumber(props.value)
return <span>{_isEmpty(parsed) ? props.value : libPhoneImport.formatNumber(parsed, 'International')}</span>
} ellse {
 return <span>{props.value}</span>
}
```

you can also move it to a function

`return <span>{libPhoneImport ? formatWithLib(...) : props.value}</span>`

packages/tocco-ui/src/FormattedValue/typeFormatters/PhoneFormatter.spec.js
khanlifts: The library formats the first 2-3 signs as prefix for all the possible areas (+1, +31, +3).

MethodenMann: As far as i know both are invalid phone numbers. the effect is the same, the component shows the original value unformatted?

PR #569, opened 2018-10-17
feat(tocco-ui): focus empty LoginForm input

khanlifts:
set focus on empty LoginForm input if username is provided

### DIFF
--- a/packages/login/src/components/LoginForm/LoginForm.js
+++ b/packages/login/src/components/LoginForm/LoginForm.js
@@ -26,6 +26,9 @@ export class LoginForm extends Component {
   }
 
   render() {
+    const passwordFocus = !!this.props.username
+    const usernameFocus = !passwordFocus
+
     return (
       <div className="login-form">
         {
@@ -47,7 +50,7 @@ export class LoginForm extends Component {
                 placeholder={this.msg('client.login.form.userPlaceholder')}
                 value={this.props.username}
                 required
-                autoFocus
+                autoFocus={usernameFocus}
               />
             </div>
           </div>
@@ -61,6 +64,7 @@ export class LoginForm extends Component {
                 onChange={this.handlePasswordChange.bind(this)}
                 placeholder={this.msg('client.login.form.passwordPlaceholder')}
                 required
+                autoFocus={passwordFocus}
               />
             </div>
           </div>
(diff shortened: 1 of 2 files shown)


packages/login/src/components/LoginForm/LoginForm.js
```diff
@@ -12,6 +12,18 @@ import {
 import {Pages} from '../../types/Pages'
 
 export class LoginForm extends Component {
+  constructor(props) {
+    super(props)
+
+    this.passwordRef = React.createRef()
```
rzueger: looks quite complicated :)

couldn't we simply set the `autoFocus` prop on the password field instead of the username field if `props.username` is set?

packages/login/src/components/LoginForm/LoginForm.js
```diff
@@ -12,6 +12,22 @@ import {
...
+
+  setAutoFocus() {
+    if (this.props.username) {
+      this.usernameFocus = false
+      this.passwordFocus = true
+    } else {
+      this.usernameFocus = true
+      this.passwordFocus = false
+    }
+  }
```
rzueger: still quite complicated and I believe it won't work in some cases, because `usernameFocus` and `passwordFocus` won't be updated when the props change. :)

you don't need a constructor or any other additional method in order to solve it. :) simply build the required conditions inside the render method.

packages/login/src/components/LoginForm/LoginForm.js
```diff
@@ -26,6 +26,8 @@ export class LoginForm extends Component {
   }
 
   render() {
+    this.passwordFocus = !!this.props.username
```
rzueger: should be a local variable. check if prop has been set on the correct component in the test.

packages/login/src/components/LoginForm/LoginForm.js
```diff
@@ -61,6 +63,7 @@ export class LoginForm extends Component {
                 onChange={this.handlePasswordChange.bind(this)}
                 placeholder={this.msg('client.login.form.passwordPlaceholder')}
                 required
+                autoFocus={passwordFocus}
```
rzueger: `autoFocus` is still set on username field. Maybe it would be better to set it explicitly only if the password field shouldn't be focused.

PR #575, opened 2018-10-23
feat(toco-ui): add MultiDate Component

khanlifts:
component to set one date or a range of dates

### DIFF
--- a/packages/tocco-ui/src/EditableValue/typeEditors/MultiDateEdit.js
+++ b/packages/tocco-ui/src/EditableValue/typeEditors/MultiDateEdit.js
@@ -1,0 +1,165 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import moment from 'moment'
+import {injectIntl, intlShape} from 'react-intl'
+
+import DateAbstract from './DateAbstract'
+import {atMostOne, momentJStoToFlatpickrFormat, toLocalDateString} from '../utils'
+
+class MultiDateEdit extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      showCalendar: true,
+      showDoubleCalendar: false,
+      counter: 0,
+      label: this.RANGE_CALENDAR_LABEL
+    }
+    this.onChangeObject = {
+      exactValue: null,
+      fromValue: null,
+      toValue: null
+    }
+  }
+  DATE_FORMAT = 'YYYY-MM-DD'
+  RANGE_CALENDAR_LABEL = 'Range Calendar'
+  SINGLE_CALENDAR_LABEL = 'Exact Calendar'
+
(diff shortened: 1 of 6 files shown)


packages/tocco-ui/src/EditableValue/typeEditors/MultiDateEdit.js
```diff
@@ -0,0 +1,165 @@
...
+class MultiDateEdit extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      showCalendar: true,
+      showDoubleCalendar: false,
+      counter: 0,
+      label: this.RANGE_CALENDAR_LABEL
+    }
+    this.onChangeObject = {
```
rzueger: why don't you handle the values in the state? I guess that's what React State is for? :)

packages/tocco-ui/src/EditableValue/typeEditors/MultiDateEdit.js
```diff
@@ -0,0 +1,165 @@
...
+      showCalendar: !prevState.showCalendar,
+      showDoubleCalendar: !prevState.showDoubleCalendar,
+      counter: prevState.counter + 1
+    }))
+
+    this.switchValues()
+  }
+
+  switchValues = () => {
+    if (this.state.counter % 2 === 0) {
```
MethodenMann: why do you use a counter for a this true/false state?


Wouldn't it be way more intuitiv to actually use a bool? If i look at the state i would ask myself what this counter is for.

packages/tocco-ui/src/EditableValue/typeEditors/MultiDateEdit.js
khanlifts: When switching `setState()` wasn't the right fit because it's async.
From https://reactjs.org/docs/faq-state.html#when-is-setstate-asynchronous:

> Currently, setState is asynchronous inside event handlers.

packages/tocco-ui/src/EditableValue/typeEditors/MultiDateEdit.js
```diff
@@ -0,0 +1,165 @@
...
+
+import DateAbstract from './DateAbstract'
+import {atMostOne, momentJStoToFlatpickrFormat, toLocalDateString} from '../utils'
+
+class MultiDateEdit extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      showCalendar: true,
+      showDoubleCalendar: false,
```
rzueger: If I understand the code correctly, `showCalendar` and `showDoubleCalendar` are connected. If `showCalendar` is `false`, `showDoubleCalendar` **must** be `true` and vice versa.

Hence, in my option, it would make sense to reduce it to one property. It could be called `view` or `mode` or something like that and accept the values `exact` or `range`.

PR #577, opened 2018-10-24
feat(tocco-ui): upgrade DateRangeEdit component

khanlifts:
- add exact and range selection
- add chronological selection validation

### DIFF
--- a/packages/entity-list/src/components/SearchForm/SearchForm.js
+++ b/packages/entity-list/src/components/SearchForm/SearchForm.js
@@ -38,7 +38,8 @@ class SearchForm extends React.Component {
       formFieldUtils,
       {
         ...formField.defaultMapping,
-        'fulltext-search': formField.editableValueFactory('string')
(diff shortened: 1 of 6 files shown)


khanlifts:
I wrote your requirements into the `switchValue` method on Line 75.

I left the names as they are. I switched to your recommendation and found that my names 
made the code a bit clearer when skimming the code.

rzueger:
switching between range and exact mode doesn't work as expected yet:

Change `dateRangeObject` for example in showcase to:

```
dateRangeObject: {
  exactValue: '2018-06-24'
}
```

Expected behavior:
- Component is in _exact_ mode and shows the given value
- When expanding to range view: exact value inserted in both the `from` and `to` field

Actual behavior:
- Component is in _exact_ mode and shows the given value
- When expanding to range view: `from` and `to` field empty

---

Other example:

If the initial value consists of a range (without exact value), I think the component should to range mode automatically. What's your opinion?

---

Furthermore, in my opinion, an additional prop to specify the initial mode would be useful

PR #578, opened 2018-10-25
Client Actions

MethodenMann:


packages/entity-list/src/modules/list/sagas.js
```diff
@@ -185,6 +193,28 @@ export function* onRowClick({payload}) {
   yield put(externalEvents.fireExternalEvent('onRowClick', {id: payload.id}))
 }
 
+export function* deleteEntities({payload}) {
+  const blockingInfoId = uuid()
+  yield put(notifier.blockingInfo(blockingInfoId, 'client.entity-detail.deleteInProgress', null))
+  const {entity, ids} = payload
+  const deleteRequests = yield all(ids.map(id => {
+    return call(rest.deleteEntity, entity, id)
+  }))
```
rzueger: (could be reduced to `id => call(...)`)

PR #580, opened 2018-10-30
style(tocco-ui): scale LoadMask icon down (70px to 30px)

niklaushug:


rzueger:
@MethodenMann This is primarily for the incoming payment action. What would be the correct target branch? nice-releases/218?

PR #584, opened 2018-11-07
Pr/consolidate theme

niklaushug:


package.json
```diff
@@ -127,5 +127,8 @@
     "webpack-hot-middleware": "^2.22.2",
     "whatwg-fetch": "2.0.4",
     "yargs": "12.0.1"
+  },
```
MethodenMann: add as dev dependency

packages/tocco-theme/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "devDependencies": {
     "bootstrap-sass": "^3.3.7",
-    "react-select": "^1.0.0-rc.10"
+    "react-select": "^1.0.0-rc.10",
+    "tocco-ui": "^0.1.14"
```
MethodenMann: are you sure about this? I think this adds a circular dependency because ui should depend on theme and not the other way around?

niklaushug: tocco-theme is not imported in tocco-ui. Hence I do not see any circular dependency.

tocco-ui is imported in tocco-theme to reuse utility function trimDecimalPlaces in modularScale.js. Probably it is better to remove the dependency an replace trimDecimalPlaces by Lodash [_.round](https://lodash.com/docs/4.17.10#round). What do you think?

MethodenMann: sound better yes

packages/tocco-ui/src/utilStyles/declareInteractionColors.js
```diff
@@ -23,57 +31,150 @@ const declareInteractionColors = (colors, format = stylingFormat.HTML) => {
...
+    focusColor: fg[1],
+    focusBackground: bg[1],
+    activeColor: fg[2],
+    activeBackground: bg[2]
   }
 }
 
-const declareFlatPrimaryColors = props => {
+const generateFlatPrimaryColors = props => {
+  const primary = themeGet('colors.primary', '#010101')(props)
```
MethodenMann: Move fallbacks to const?

packages/tocco-ui/src/utilStyles/declareInteractionColors.js
```diff
@@ -23,57 +31,150 @@ const declareInteractionColors = (colors, format = stylingFormat.HTML) => {
...
-    focusColor: theme('colors.primary.fillContrast.1')(props),
-    focusBackground: theme('colors.primary.fill.1')(props),
-    activeColor: theme('colors.primary.fillContrast.2')(props),
-    activeBackground: theme('colors.primary.fill.2')(props)
+    defaultColor: fg[0],
+    defaultBackground: bg[0],
+    focusColor: fg[1],
+    focusBackground: bg[1],
+    activeColor: fg[2],
+    activeBackground: bg[2]
```
MethodenMann: loads of repetitive code in this file.

Example why not make something like

```
const idk = fg => ({
   defaultColor: fg[0],
    defaultBackground: bg[0],
    focusColor: fg[1],
    focusBackground: bg[1],
    activeColor: fg[2],
    activeBackground: bg[2]
})
```

and then this 3 times:

`return idk(fg)  `


what does fg stand for?

packages/tocco-ui/src/utilStyles/modularScale.js
```diff
@@ -0,0 +1,18 @@
...
+  return Math.pow(scale, exponent) * base
+}
+
+export const fontScale = (props, exponent, unit = 'rem') => {
+  const base = theme('fontSizeBase')(props)
+  const scale = theme('fontSizeScale')(props)
+  return `${_round(scaleExponential(base, exponent, scale), 3)}${unit}`
+}
+
+export const spaceScale = (props, exponent, unit = 'rem') => {
```
MethodenMann: same function as above expect the "fontSizeBase" and "spaceBase" selector? why no move it to one and make the selector a parameter

packages/tocco-ui/src/utilStyles/declareInteractionColors.js
niklaushug: fg = foreground
bg = background

packages/tocco-theme/src/ToccoTheme/DefaultTheme.js
```diff
@@ -1,141 +1,93 @@
...
-const space = [
-  0, '.0625rem', '.125rem', '.25rem', '.5rem', '1rem', '2rem', '4rem', '8rem', '16rem', '32rem'
-]
+const spaceBase = _round(fontSizeBase * lineHeights.regular, 3)
+const spaceScale = 2
 
 const defaultTheme = {
   colors,
-  fontSizes,
+  fontSizeBase,
```
MethodenMann: IMO this file would be easier to read if you directly export an object rather than building an object with consts and then export it

packages/tocco-theme/src/ToccoTheme/DefaultTheme.js
```diff
@@ -1,141 +1,66 @@
...
-      '#BDBDBD' // Grey 400
-    ],
-    line: [
-      '#424242', // Grey 800
-      '#303030', // Grey 850
-      '#212121' // Grey 900
-    ],
-    paper: '#FFFFFF',
-    text: '#212121' // Grey 900
+export default {
```
MethodenMann: file should be called defaultTheme i guess. only react components are pascal-case

PR #586, opened 2018-11-08
refactor(entity-list): add support for multi fields

MethodenMann:
- add mock data


packages/entity-list/src/util/api/entities.js
```diff
@@ -36,42 +37,43 @@ export function* fetchModel(entityName, transformer = defaultModelTransformer) {
...
+      value: v.display
+    }))
+  } else if (type === 'display-expression') {
+    return {
+      type: 'html',
+      value: path.value
+    }
+  } else if (type === 'multi') {
+    return path.value.map(v => getFieldValues(v))
+  } else {
```
rzueger: I'm sure we've already talked about my following question earlier:
would it make sense to log a warning in the fallback case?

packages/tocco-util/src/mockData/data/user_list_form.json
```diff
@@ -390,6 +411,7 @@
             "hidden": false,
             "sortable": true
           },
+
```
rzueger: (blank line inserted intentionally?)

PR #589, opened 2018-11-14
feat(tocco-ui): add MoneyEdit component

khanlifts:


packages/app-extensions/src/formField/editableValueFactory.js
```diff
@@ -95,8 +95,8 @@ const getOptions = (type, formField, modelField, utils) => {
       if (utils.intl) {
         options.intl = utils.intl
       }
-      options.postPointDigits = modelField.postPointDigits
-      options.maxValue = modelField.maxValue
+      options.postPointDigits = modelField.validation.decimalDigits.postPointDigits
```
MethodenMann: i would use lodash get so make this a bit safer. Otherwise an undefined validation object would throw en exception.

packages/tocco-ui/src/EditableValue/typeEditors/DecimalEdit.js
```diff
@@ -14,6 +14,10 @@ const DecimalEdit = props => {
   const {thousandSeparator, decimalSeparator} = parseLocalePlaceholder(props.options.intl.locale)
   const value = props.value === null ? '' : props.value
 
+  const maxInteger = (10 ** props.options.prePointDigits)
```
MethodenMann: use better const name and move this calculation to a function

packages/tocco-ui/src/EditableValue/typeEditors/DecimalEdit.js
```diff
@@ -32,11 +36,18 @@ const DecimalEdit = props => {
...
       decimalSeparator={decimalSeparator}
-      isAllowed={limitValue(props.options.maxValue)}
+      isAllowed={limitValue(maxValue)}
     />
   )
 }
 
+DecimalEdit.defaultProps = {
+  options: {
+    prePointDigits: 12,
```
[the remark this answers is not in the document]
khanlifts: I thought about making it more flexible in terms of the case that no pre- and postPointDigits were passed to the component. Considering this, should I still remove defaultProps?

PR #590, opened 2018-11-14
feat(tocco-ui): establish SomeOfButton

niklaushug:
- Numbers greater than 999 are shortened and affixed by system unit.


packages/tocco-ui/src/SomeOfButton/SomeOfButton.js
```diff
@@ -0,0 +1,88 @@
...
+import _round from 'lodash/round'
+
+import Button from '../Button'
+import {
+  inkPropTypes,
+  stylingLook
+} from '../utilStyles'
+
+const affixSystemUnit = value => {
+  const log10 = Math.log10(value)
```
MethodenMann: Isnt it possible to create an object with the units and loop over it? This would increase readability.
im thinking about something like this:

`
[{value: 3, symbol: 'k'}, {value: 9, symbol: 'M'}]

`

niklaushug: Thanks for inspiration.

niklaushug:
Should SomeOf be integrated into FormattedValue?

packages/tocco-ui/src/SomeOf/SomeOf.spec.js
```diff
@@ -0,0 +1,17 @@
+import {mount} from 'enzyme'
+import React from 'react'
+
+import SomeOf from './SomeOf'
+
+describe('tocco-ui', () => {
+  describe('SomeOf', () => {
+    test('should display values', () => {
+      const wrapper = mount(<SomeOf some={12} of={345}/>)
+      expect(wrapper.find('span').text()).to.equal('12 / 345')
```
MethodenMann: in general it would be the better approach to check whether the output contains the some and of value then to check for a very specific format.
e.g. we could replace "/" with a ":" and the test should not fail IMO

packages/tocco-util/src/units/addSystemUnit.js
```diff
@@ -0,0 +1,15 @@
+import _round from 'lodash/round'
+
+export const addSystemUnit = value => {
```
MethodenMann: i dont quite like the name. Wouldn't "shorten something" be a better name?

packages/tocco-util/src/units/addSystemUnit.js
```diff
@@ -0,0 +1,15 @@
...
+  const ranges = [
+    {base: 15, symbol: 'P'},
+    {base: 12, symbol: 'T'},
+    {base: 9, symbol: 'G'},
+    {base: 6, symbol: 'M'},
+    {base: 3, symbol: 'k'},
+    {base: 0, symbol: ''}
```
MethodenMann: 👍

packages/tocco-util/src/units/addSystemUnit.spec.js
```diff
@@ -0,0 +1,33 @@
...
+    describe('addSystemUnit', () => {
+      test('shorten numbers correctly', () => {
+        const scenarios = [
+          [1, '1'],
+          [12, '12'],
+          [123, '123'],
+          [1234, '1k'],
+          [12345, '12k'],
+          [123456, '123k'],
+          [1234567, '1M'],
```
MethodenMann: I would like an edge case test. e.g. 1999999 should get 1.9M or maybe 2M but not 1M

packages/tocco-util/src/units/addSystemUnit.js
niklaushug: I change to "condenseNumber".

packages/tocco-ui/src/SomeOf/SomeOf.spec.js
niklaushug: https://www.chaijs.com/plugins/chai-string/ or https://www.chaijs.com/plugins/chai-match/ would be helpful. Please consider to add one of those.

packages/tocco-util/src/units/addSystemUnit.spec.js
niklaushug: - Unwanted outputs like 1000k is not possible anymore. 999.5K is rounded up to 1M. 
- Complex math were removed (log10).
- Needed loops were reduced by reversing array ranges by the assumption most numbers are less than 1'000'000.

niklaushug: Edge Case 0 does cause an error.

PR #595, opened 2018-11-19
feat(tocco-ui): add format fields of type long to Entity-Browser

khanlifts:


packages/tocco-util/src/mockData/data/user_template.json
```diff
@@ -47,6 +47,15 @@
         "writable": false
       }
     },
+    "licence_plate": {
```
MethodenMann: not needed because gets overwritten by "entityFactory" anyways?

packages/tocco-util/src/mockData/data/user_search_form.json
```diff
@@ -115,7 +115,22 @@
             "hidden": false,
             "scopes": [ ]
           },
-
+          {
```
MethodenMann: good that you tested it but i think you dont need to commit that field in the search form.

packages/tocco-util/src/mockData/data/report_settings.json
```diff
@@ -282,6 +282,15 @@
       ],
       "dataType": "single-select-box"
     },
+    {
```
MethodenMann: not really needed here.

packages/tocco-util/src/mockData/data/user_list_small_form.json
```diff
@@ -35,6 +35,23 @@
             "label": "Nr.",
             "sortable": true
           },
+          {
+            "id": "licence_plate",
```
MethodenMann: not really needed here

PR #597, opened 2018-11-26
Pr/display relevant datatypes

khanlifts:


packages/tocco-ui/src/EditableValue/example.js
```diff
@@ -65,6 +65,7 @@ class Example extends React.Component {
         datetime: '2017-01-25T15:15:00.000Z',
         decimal: 123456.78,
         duration: 3660000,
+        time: 0,
```
MethodenMann: why do you set 0 as example value and not a "real" value like 18:10

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,57 @@
...
+    hoursOfDay: PropTypes.number,
+    minutesOfHour: PropTypes.number
+  }),
+  name: PropTypes.string,
+  id: PropTypes.string,
+  readOnly: PropTypes.bool,
+  options: PropTypes.shape({
+    hoursLabel: PropTypes.string,
+    minutesLabel: PropTypes.string,
+    maxHours: PropTypes.number
```
MethodenMann: this does not seem right. Why should i use a time component with a maxHours not 24?

packages/tocco-ui/src/EditableValue/example.js
MethodenMann: i thought the time value is an object?

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,57 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import DurationEdit from './DurationEdit'
+import {calculateMilliseconds} from '../utils'
+
+const TimeEdit = props => {
```
MethodenMann: There is a bug in the component when i enter 18 into hour field and then 33 in minutes field the components calls onChange with value "{hoursOfDay: 6, minutesOfHour: 33}" instead of "{hoursOfDay: 18, minutesOfHour: 33}"

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,57 @@
...
+  onChange: PropTypes.func.isRequired,
+  value: PropTypes.shape({
+    hoursOfDay: PropTypes.number,
+    minutesOfHour: PropTypes.number
+  }),
+  name: PropTypes.string,
+  id: PropTypes.string,
+  readOnly: PropTypes.bool,
+  options: PropTypes.shape({
+    hoursLabel: PropTypes.string,
```
MethodenMann: hoursLabel should be removed since its not used?

packages/tocco-ui/src/FormattedValue/typeFormatterProvider.js
```diff
@@ -64,6 +64,7 @@ export const map = {
   'phone': PhoneFormatter,
   'postcode': StringFormatter,
   'remote': SingleSelectFormatter,
+  'remote-field': SingleSelectFormatter,
```
MethodenMann: did you change this line intentionally?  doesnt seem to be related to the commit msg

packages/app-extensions/src/formField/defaultMapping.js
```diff
@@ -32,8 +36,10 @@ export default {
     'date': editableValueFactory('date-range'),
     'birthdate': editableValueFactory('date-range')
   },
+  'remote-field': editableValueFactory('remote'),
```
MethodenMann: haven't we found out that this is not needed?

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
```diff
@@ -23,7 +24,14 @@ class DurationEdit extends React.Component {
   }
 
   handleHourChange = e => {
-    const hours = e.target.value.replace(/[^\d]/g, '')
+    let hours = e.target.value.replace(/[^\d]/g, '')
+    if (this.props.options.maxHours) {
+      if (hours > 23) {
```
MethodenMann: why 23?

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import DurationEdit from './DurationEdit'
+import {calculateMilliseconds} from '../utils'
+
+const TimeEdit = props => {
+  const propHours = (props.value === null || props.value === undefined) ? 0 : props.value.hoursOfDay
```
MethodenMann: Maybe move (props.value === null || props.value === undefined) into a var. Or create a helper function "isNullOrUndefined" i think this boilerplate code makes the code hard to read

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,55 @@
...
+  const propMinutes = (props.value === null || props.value === undefined) ? 0 : props.value.minutesOfHour
```
MethodenMann: i question the name "propMinutes". why not only minutes, there is no extra value provided by the prop prefix

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,55 @@
...
+
+TimeEdit.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  value: PropTypes.shape({
+    hoursOfDay: PropTypes.number,
+    minutesOfHour: PropTypes.number
+  }),
+  name: PropTypes.string,
+  id: PropTypes.string,
+  readOnly: PropTypes.bool
```
MethodenMann: popTypes for options?

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <DurationEdit
+      onChange={handleChange}
+      value={value}
+      name={props.name}
+      id={props.id}
+      readOnly={props.readOnly}
+      options={{
+        maxHours: true,
```
MethodenMann: i dont get "maxHours". why is it a boolean?

packages/tocco-ui/src/EditableValue/utils.js
```diff
@@ -70,3 +70,12 @@ export const parseLocalePlaceholder = countryCode => {
 export const convertStringToNumber = stringValue => (
   !stringValue || isNaN(stringValue) ? null : parseFloat(stringValue)
 )
+
+/*
+ * Convert two numbers as hours and minutes to milliseconds
+ */
+export const calculateMilliseconds = (hoursValue, minutesValue) => {
```
MethodenMann: hoursValue = hours
minutesValue = minutes 

would be better IMO

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,55 @@
...
+import PropTypes from 'prop-types'
+
+import DurationEdit from './DurationEdit'
+import {calculateMilliseconds} from '../utils'
+
+const TimeEdit = props => {
+  const propHours = (props.value === null || props.value === undefined) ? 0 : props.value.hoursOfDay
+  const propMinutes = (props.value === null || props.value === undefined) ? 0 : props.value.minutesOfHour
+
+  const millisecondsToDuration = milliSeconds => {
```
MethodenMann: since "calculateMilliseconds" is move to util, this could also? Better for unit testing as well

packages/tocco-ui/src/FormattedValue/typeFormatterProvider.js
MethodenMann: and is it needed?

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
```diff
@@ -23,7 +25,14 @@ class DurationEdit extends React.Component {
   }
 
   handleHourChange = e => {
-    const hours = e.target.value.replace(/[^\d]/g, '')
+    let hours = e.target.value.replace(/[^\d]/g, '')
+    if (this.maxHours) {
+      if (hours > this.maxHours) {
+        hours = 0
+      } else if (hours < 0) {
+        hours = this.maxHours
```
MethodenMann: is it possible to enter a negative value in case of no maxHours defined?

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
MethodenMann: to that's a no on this comment?

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
MethodenMann: Shower thoughts: can a duration be negative?

packages/tocco-ui/src/EditableValue/utils.js
```diff
@@ -70,3 +70,17 @@ export const parseLocalePlaceholder = countryCode => {
 export const convertStringToNumber = stringValue => (
   !stringValue || isNaN(stringValue) ? null : parseFloat(stringValue)
 )
+
+/*
+ * Convert two numbers as hours and minutes to milliseconds
+ */
+export const calculateMilliseconds = (hours, minutes) => {
+  const hoursMilliseconds = (hours || 0) * 60 * 60000
```
MethodenMann: This method will break with negative values since -1 is false. does that matter?

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.spec.js
```diff
@@ -0,0 +1,24 @@
...
+    describe('typeEditors', () => {
+      describe('TimeEdit ', () => {
+        const valueObject = {
+          value: {hoursOfDay: 8, minutesOfHour: 45}}
+
+        test('should render DurationEdit', () => {
+          const wrapper = mount(
+            <TimeEdit value={valueObject} onChange={EMPTY_FUNC}/>
+          )
+          expect(wrapper.find('DurationEdit')).to.have.length(1)
```
MethodenMann: you could check if the output contains an "8" and the number "45". this would be a more or less useful unit test

packages/tocco-ui/src/EditableValue/example.js
khanlifts: That's right.

packages/tocco-ui/src/FormattedValue/typeFormatterProvider.js
khanlifts: It is not. All related fields should be either of type single-remote-field or multi-remote-field.

packages/app-extensions/src/formField/defaultMapping.js
khanlifts: Yes.

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
khanlifts: Moved it to utils.js.

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
khanlifts: Duration should not be negative. Lines 32/33 were redundant anyway because the regex above and the prop on the input element didn't allow negative values.

packages/tocco-ui/src/EditableValue/utils.js
khanlifts: It shouldn't because Time and Duration components prevent negative inputs already.

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.spec.js
khanlifts: IMO this kind of test belongs to DurationEdit, as there already is.

MethodenMann: if you do white box tests you are right.

packages/tocco-ui/src/EditableValue/example.js
```diff
@@ -65,6 +65,10 @@ class Example extends React.Component {
         datetime: '2017-01-25T15:15:00.000Z',
         decimal: 123456.78,
         duration: 3660000,
+        time: {
+          minutesOfHour: 33,
```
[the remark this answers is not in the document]
khanlifts: Isn't it inappropriate to provide two entries for milliseconds (millisOfDay & millisOfSecond)?

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import StyledTimeEdit from './StyledTimeEdit'
+import Button from '../../Button'
+import {stringToDuration, numbersToTimeFormat} from '../utils'
+
+const TimeEdit = props => {
+  const hours = props.value.hourOfDay
```
MethodenMann: this does crash if the value is null

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,58 @@
...
+  const handleChange = e => {
+    if (props.onChange) {
+      props.onChange(stringToDuration(e.target.value))
+    }
+  }
+
+  const clearInput = () => {
+    const emptyTimeObject = {hoursOfDay: undefined, minuteOfHour: undefined}
+    if (props.onChange) {
+      props.onChange(emptyTimeObject)
```
MethodenMann: component should return en empty value not an empty "timeobject"

packages/tocco-ui/src/EditableValue/utils.js
```diff
@@ -70,3 +70,65 @@ export const parseLocalePlaceholder = countryCode => {
...
+  const minuteOfHour = parseInt(timeString.split(':')[1])
+  return {hourOfDay, minuteOfHour}
+}
+
+/*
+ * Convert hour and minute numbers to time string of type HH:MM
+ */
+export const numbersToTimeFormat = (hours, minutes) => {
+  if (!hours && !minutes) {
+    return '--:--'
```
MethodenMann: i wouldn't do that. why not return an empty string? The input will show this string (--:--) without help

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,58 @@
...
+        name={props.name}
+        id={props.id}
+        readOnly={props.readOnly}
+      />
+      <Button
+        icon="times"
+        look="raised"
+        onClick={clearInput}
+      />
+    </StyledTimeEdit>
```
MethodenMann: TimeEdit works well, but when i enter 12 in the hours field and press tab twice to leave the field i expect it to set 00 into minutes field. or at least that its validated properly. rightnow i get de fault html validation if i try to submit the form with 12:--

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,58 @@
...
+      <input
+        className="form-control"
+        type="time"
+        value={timeString}
+        onChange={handleChange}
+        name={props.name}
+        id={props.id}
+        readOnly={props.readOnly}
+      />
+      <Button
```
MethodenMann: tabIndex = 0 ?

packages/tocco-ui/src/EditableValue/utils.js
```diff
@@ -70,3 +70,65 @@ export const parseLocalePlaceholder = countryCode => {
...
+  }
+  const hourString = hours ? hours.toString() : '00'
+  const minuteString = minutes ? minutes.toString() : '00'
+  const timeArray = [hourString, minuteString]
```
MethodenMann: why do you push this into an array? you know that there are always 2 values (hours and minutes)

packages/tocco-ui/src/EditableValue/utils.js
```diff
@@ -70,3 +70,65 @@ export const parseLocalePlaceholder = countryCode => {
...
+  const checkArray = []
+  timeArray.forEach(item => {
```
MethodenMann: take a look at padStart

packages/tocco-ui/src/EditableValue/utils.js
khanlifts: This is needed to output --:-- when the value is e.g. --:12 or 23:--.
The target value of input[type="time"] does not differentiate between an empty string and the two examples above and outputs an empty string in each case.

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,58 @@
...
+      />
+    </StyledTimeEdit>
+  )
+}
+
+TimeEdit.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  value: PropTypes.shape({
+    hourOfDay: PropTypes.number,
+    minuteOfHour: PropTypes.number
```
rzueger: There can be two more properties in the value object

packages/tocco-ui/src/EditableValue/utils.js
```diff
@@ -70,3 +70,70 @@ export const parseLocalePlaceholder = countryCode => {
...
+
+/*
+ * Return true if the value is null or undefined
+ */
+export const isNullOrUndefined = value => value === null || value === undefined
+
+/*
+ * Return true if the value is null or undefined
+ */
+export const millisecondsToDuration = milliSeconds => {
```
rzueger: `milliseconds` is one word --> no camel case

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,58 @@
...
+import {stringToDuration, numbersToTimeFormat} from '../utils'
+
+const TimeEdit = props => {
+  const hours = props.value ? props.value.hourOfDay : null
+  const minutes = props.value ? props.value.minuteOfHour : null
+
+  const timeString = numbersToTimeFormat(hours, minutes)
+
+  const handleChange = e => {
+    if (props.onChange) {
```
rzueger: is `onChange` a required prop? the prop types definition says so. in this case, the check here isn't necessary

packages/tocco-ui/src/EditableValue/typeEditors/TimeEdit.js
```diff
@@ -0,0 +1,58 @@
...
+      props.onChange(stringToDuration(e.target.value))
+    }
+  }
+
+  const clearInput = () => {
+    if (props.onChange) {
```
rzueger: is `onChange` a required prop? the prop types definition says so. in this case, the check here isn't necessary

packages/tocco-ui/src/EditableValue/utils.js
```diff
@@ -70,3 +70,70 @@ export const parseLocalePlaceholder = countryCode => {
...
+  const minuteOfHour = parseInt(timeString.split(':')[1])
+  return {hourOfDay, minuteOfHour}
+}
+
+/*
+ * Pad a string with leading zeros
+ */
+export const padLeadingZeros = (string, numberOfZeros) => {
+  return string.length < numberOfZeros ? string.padStart(numberOfZeros, '0') : string
+}
```
rzueger: could be simplified to

```
... (string, numberOfZeros) => string.length < ...
```

(without `return`)

packages/tocco-ui/src/FormattedValue/typeFormatters/TimeFormatter.js
```diff
@@ -27,7 +27,12 @@ const TimeFormatter = props => {
 
 TimeFormatter.propTypes = {
   intl: intlShape.isRequired,
-  value: PropTypes.object.isRequired
+  value: PropTypes.shape({
+    hourOfDay: PropTypes.number,
+    minuteOfHour: PropTypes.number,
+    secondOfMinute: PropTypes.number,
+    millisOfSecond: PropTypes.number
```
rzueger: The naming of the properties is not consistent. why are hours, minutes and seconds named in singular and millis in plural?

packages/tocco-ui/src/EditableValue/typeEditors/StyledTimeEdit.js
```diff
@@ -0,0 +1,33 @@
...
+          
+    > button {
+      border: #CCCCCC 1px solid;
+      border-left: 0;
+      padding: 9px 14px;
+      background-color: #EEEEEE;
+      border-top-left-radius: 0;
+      border-bottom-left-radius: 0;
+    }
+
```
rzueger: (blank line could be removed)

packages/tocco-ui/src/EditableValue/utils.js
rzueger: can be formatted nicely, you don't have to write it on one single line :)

PR #598, opened 2018-11-28
Pr/icon expansion

MethodenMann:


packages/tocco-ui/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "devDependencies": {
     "@fortawesome/fontawesome-svg-core": "^1.2.8",
+    "@fortawesome/free-brands-svg-icons": "^5.5.0",
```
niklaushug: Though icons are chunked and delivered separately, it has an impact on loading time. As I added fortawesome I have decided not to include brands. Are brand icons requested now? How many brand icons are used? We still have the possibility to assemble an own set of brand icons.

packages/tocco-ui/src/Icon/Icon.js
```diff
@@ -40,9 +40,12 @@ class Icon extends React.Component {
   }
 
   render() {
-    const filteredProps = _omit(this.props, ['dense', 'onLoaded', 'position', 'theme'])
```
niklaushug: Why was filteredProps removed? It ensures that only props are passed down to react-fontawesome which are consumed. All props which are only consumed by Icon are removed by _omit.

packages/tocco-ui/src/Icon/Icon.js
```diff
@@ -40,9 +40,12 @@ class Icon extends React.Component {
...
+    const icon = (typeof this.props.icon === 'string' && this.props.icon.includes(','))
```
niklaushug: When prop.icon is defined correctly this code is not needed. May extend PropTypes validation by allowing only strings without comma and remove this code.

packages/tocco-ui/src/Icon/example.js
```diff
@@ -9,8 +9,8 @@ export default () => {
   return (
     <div>
       {/* start example */}
-      <Icon icon="address-card" /><br />
-      <Icon icon={['far', 'address-card']} /><br />
+      <Icon icon="address-card" /><Icon icon="fab, microsoft"/><br />
```
niklaushug: Define as array and use curly braces like in the following line.

packages/tocco-ui/src/Icon/Icon.js
MethodenMann: I dont quite get your comment. That you can use a comma separated string is a feature. I think in most cases it's friendlier this way.

packages/tocco-ui/src/Icon/Icon.js
MethodenMann: Yes, i know. So what is there in "props" except the icon attribute which i handle separately?

packages/tocco-ui/src/Icon/example.js
MethodenMann: As i said above this is something i explicitly want to support so i want to show it in the example

PR #600, opened 2018-12-07
Pr/map residual number datatypes

khanlifts:


packages/tocco-util/src/mockData/data/user_model.json
```diff
@@ -172,6 +172,13 @@
         }
       }
     },
+    {
+      "fieldName": "rank",
```
MethodenMann: do we need two integer fields in the mock data?

packages/tocco-ui/src/FormattedValue/typeFormatters/NumberFormatter.js
```diff
@@ -4,18 +4,20 @@ import {FormattedNumber} from 'react-intl'
...
+        style="decimal"
+        minimumFractionDigits={2}
+      />
+    </Typography.Span>
+  )
+}
 
 NumberFormatter.propTypes = {
-  value: PropTypes.number
+  value: PropTypes.number.isRequired
```
MethodenMann: i dont think we sould set a value as required. a value should and can be empty/null

packages/app-extensions/src/formField/defaultMapping.js
```diff
@@ -17,25 +17,30 @@ export default {
...
-    'number': editableValueFactory('number'),
-    'counter': editableValueFactory('number'),
+    'number': editableValueFactory('decimal'),
+    'counter': editableValueFactory('decimal'),
     'date': editableValueFactory('date-range'),
     'birthdate': editableValueFactory('date-range')
   },
   'single-remote-field': editableValueFactory('remote'),
   'search-filter': editableValueFactory('search-filter'),
+  'sorting': editableValueFactory('number'),
```
MethodenMann: i'm a little confused. why do we have 3 types (decimal, number and integer)? Shouldn't two be sufficient ? Integer for whole numbers and decimal for those with a floating point?

packages/entity-detail/src/util/detailView/readOnlyFormFieldMapping.js
```diff
@@ -22,8 +22,8 @@ export default {
   'multi-remote-field': formField.formattedValueFactory('multi-remote'),
   'multi-select-box': formField.formattedValueFactory('multi-select'),
   'named-upload': formField.formattedValueFactory('document'),
-  'integer': formField.formattedValueFactory('number'),
-  'percent': formField.formattedValueFactory('number'),
+  'integer': formField.formattedValueFactory('decimal'),
```
MethodenMann: Shouldn't integer be mapped to integer?

packages/tocco-ui/src/EditableValue/typeEditors/IntegerEdit.js
```diff
@@ -0,0 +1,41 @@
...
+
+const IntegerEdit = props => {
+  const handleChange = value => {
+    if (props.onChange) {
+      props.onChange(value)
+    }
+  }
+
+  return (
+    <NumberEdit
```
MethodenMann: Hm so the IntegerEdit is exactly the same as Number Edit? Does it make sense to set PostPoint Digits according to the prop options.postPointDigits. With integer is always 0.

packages/tocco-ui/src/EditableValue/typeEditors/NumberEdit.js
```diff
@@ -1,37 +1,95 @@
...
-  const handleChange = e => {
+
+  const prePointDigits = props.options.prePointDigits
+  const postPointDigits = props.options.postPointDigits
+  const prePointMarker = prePointDigits === 0 ? true : prePointDigits
+  const postPointMarker = postPointDigits === 0 ? true : postPointDigits
+
+  const minValue = props.options.minValue
+  const maxValue = props.options.maxValue
+  const minValueMarker = minValue === 0 ? true : minValue
```
MethodenMann: minValue = -1 would be a valid value    that means minValueMarker = -1 is possible

Problems come on line 49 where a valid number results in false since -1 = false

packages/tocco-ui/src/EditableValue/typeEditors/MoneyEdit.js
```diff
@@ -12,7 +12,7 @@ const MoneyEdit = props => {
   }
 
   return (
-    <DecimalEdit
+    <NumberEdit
```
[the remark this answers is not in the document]
khanlifts: Algeria, Bahrain, Iraq, Jordan, Kuwait, Libya and Tunisia use the Dinar and have floating points of 3.

packages/app-extensions/src/formField/editableValueFactory.js
```diff
@@ -93,11 +93,14 @@ const getOptions = (type, formField, modelField, utils) => {
       break
     case 'decimal':
     case 'moneyamount':
+    case 'integer':
```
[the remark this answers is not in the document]
khanlifts: My thoughts were that if I would receive them as props in IntegerEdit I have to specify them there.

packages/tocco-ui/src/EditableValue/typeEditors/IntegerEdit.js
khanlifts: The props already set this behaviour. If minValue and maxValue are set, as they are in the model for integers, the decimalScale is set to zero.

packages/tocco-ui/src/EditableValue/typeEditors/NumberEdit.js
khanlifts: -1 doesn't trigger false in an if condition.
`
Number | The result is false if the argument is +0, −0, or NaN, otherwise the result is true.
`

packages/tocco-ui/src/EditableValue/typeEditors/IntegerEdit.spec.js
```diff
@@ -0,0 +1,24 @@
...
+describe('tocco-ui', () => {
+  describe('EditableValue', () => {
+    describe('typeEditors', () => {
+      describe('MoneyEdit ', () => {
+        test('should render NumberEdit', () => {
+          const optionsObject = {intl: {...IntlStub, locale: 'en'}}
+          const wrapper = mount(
+            <IntegerEdit value={1234567} options={optionsObject} onChange={EMPTY_FUNC} />
+          )
+          expect(wrapper.find(NumberEdit)).to.have.length(1)
```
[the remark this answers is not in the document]
khanlifts: I did write an additional test.

packages/tocco-ui/src/EditableValue/typeEditors/MoneyEdit.js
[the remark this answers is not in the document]
MethodenMann: so with a high chance not a currency we want to support within the next few year. i would set it to 2

packages/tocco-ui/src/EditableValue/typeEditors/NumberEdit.js
```diff
@@ -1,37 +1,95 @@
...
+}
 
 const NumberEdit = props => {
+  const {thousandSeparator, decimalSeparator} = parseLocalePlaceholder(props.options.intl.locale)
   const value = props.value === null ? '' : props.value
-  const handleChange = e => {
+
+  const prePointDigits = props.options.prePointDigits
+  const postPointDigits = props.options.postPointDigits
+  const prePointMarker = prePointDigits === 0 ? true : prePointDigits
```
MethodenMann: i think there are better solution to check whether prePointDigits is a number >= 0 than this. I thank thats what you are trying to achieve?

packages/tocco-ui/src/EditableValue/typeEditorFactory.js
```diff
@@ -45,8 +45,9 @@ export const map = {
   'createuser': StringEdit,
   'createts': DateTimeEdit,
   'duration': DurationEdit,
-  'decimal': DecimalEdit,
-  'moneyamount': MoneyEdit
+  'decimal': NumberEdit,
```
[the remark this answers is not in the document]
khanlifts: `number: NumberEdit` is already defined on line 26.

packages/tocco-ui/src/EditableValue/typeEditors/MoneyEdit.spec.js
```diff
@@ -2,20 +2,20 @@ import React from 'react'
...
 
-import DecimalEdit from './DecimalEdit'
+import NumberEdit from './NumberEdit'
 
 const EMPTY_FUNC = () => {}
 
 describe('tocco-ui', () => {
   describe('EditableValue', () => {
     describe('typeEditors', () => {
       describe('MoneyEdit ', () => {
```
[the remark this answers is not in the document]
khanlifts: I mounted the wrong import NumberEdit instead of MoneyEdit.

PR #601, opened 2018-12-07
Pr/react selectv2 styling

niklaushug:


packages/tocco-ui/src/Select/DropdownIndicator.js
```diff
@@ -1,12 +1,22 @@
...
 import Button from '../Button'
 
-const DropdownIndicator = () =>
-  <span>
-    <Button
-      icon="chevron-down"
-      look="ball" />
-  </span>
+const DropdownIndicator = props =>
+  props.isDisabled
```
MethodenMann: Would this also work?

```
const DropdownIndicator = props =>
   props.isDisabed === false && <Button ..../>
```

a tenary that returns null in case of true seams a bit odd.

packages/tocco-ui/src/Select/IndicatorsContainer.js
```diff
@@ -16,11 +16,12 @@ const IndicatorsContainer = props => {
     <components.IndicatorsContainer {...restProps}>
       {children}
       {openAdvancedSearch
+      && !readOnly
```
MethodenMann: i like the icon (openadvancedsearch) to distinguish remote and select fields. This wouldn't be the possible in case of readonly. The question is, does if have to be?

packages/tocco-ui/src/Select/IndicatorsContainer.js
```diff
@@ -16,11 +16,12 @@ const IndicatorsContainer = props => {
...
       && <span
         onTouchEnd={e => e.stopPropagation()}
         onMouseDown={e => e.stopPropagation()}>
         <Button
-          disabled={readOnly || !openAdvancedSearch}
+          disabled={!openAdvancedSearch}
```
MethodenMann: select fields never have an advanced search button but also use the select component (as remote fields).  If no "openadvancedSearch" is defined the icon should not be displayed.

packages/tocco-ui/src/Select/LoadingIndicator.js
```diff
@@ -2,6 +2,6 @@ import React from 'react'
 
 import IconTocco from '../IconTocco'
 
-const LoadingIndicator = () => <IconTocco size="1.2em"/>
+const LoadingIndicator = () => <span><IconTocco size="1.8rem"/></span>
```
MethodenMann: Why is the span needed?

packages/tocco-ui/src/utilStyles/declareDensity.js
```diff
@@ -1,9 +1,14 @@
 import objectToCss from './objectToCss'
 import {spaceScale} from './modularScale'
+import {stylingLook} from '../utilStyles'
 
 const declareDensity = props => objectToCss({
   'line-height': [props.dense ? 'lineHeights.dense' : 'lineHeights.regular'],
-  'padding': props.dense ? spaceScale(props, -4) : `${spaceScale(props, -3)} ${spaceScale(props, -2)}`
+  'padding': props.look === stylingLook.BALL
```
MethodenMann: I dont really have a advice ready right now but we have to be careful that these components dont tend to be swiss army knives with multiple if then cases inside.

packages/tocco-ui/src/Select/StyledReactSelect.js
```diff
@@ -21,8 +24,21 @@ const reactSelectStyles = outerTheme => {
...
   })
-
+  const infoText = outerTheme.colors.signal.info.text
   return {
+    control: (base, state) => ({
+      ...base,
+      borderColor: state.isFocused
+        ? infoText
+        : state.theme.colors.neutral20,
+      boxShadow: state.isFocused
```
MethodenMann: you can add optional attributes with the spread operator like this:
{
 ...(condition? {attributeName : 'value'} : {})
        }

if condition is false, no attribute is added.

packages/tocco-ui/src/Select/DropdownIndicator.js
niklaushug: see 112dbc4cce34971f5044ca0e8b94d063de1726aa

packages/tocco-ui/src/Select/IndicatorsContainer.js
niklaushug: If readonly, only the current selection is relevant. Selected values defines the characteristic, not the none selected items. Listing none accessible actions are from minor interest or disturbing. 

In case that identifying a field's type is a real need than I would suggest to display an icon.

packages/tocco-ui/src/Select/IndicatorsContainer.js
niklaushug: see 83b567c3cf24252169282ef0b3bac6882ba198d9

packages/tocco-ui/src/Select/LoadingIndicator.js
niklaushug: I've introduced a new wrapping element for buttons to attach events and props (see ClearIndicator.js and IndicatorsContainer.js). LoadingIndicator.js and DropdownIndicator.js does contain that wrapping span as well in order to use the very same css.

packages/tocco-ui/src/Button/StyledButton.js
```diff
@@ -45,15 +45,15 @@ const meltButtons = props => {
...
   const {ink, look} = props
-  const {FLAT, RAISED} = stylingLook
+  const {BALL, FLAT, RAISED} = stylingLook
   const {BASE, PRIMARY} = stylingInk
   if (look === FLAT && ink === BASE) {
     declareColor = generateFlatBaseColors
   } else if (look === FLAT && ink === PRIMARY) {
     declareColor = generateFlatPrimaryColors
-  } else if (look === RAISED && ink === BASE) {
+  } else if ((look === BALL || look === RAISED) && ink === BASE) {
```
MethodenMann: if would define the (look === FLAT && ink === BASE) case as default? Because i think declareColor should always be defined otherwise the app will crash on line 59

PR #602, opened 2018-12-10
fix: Use field property `path` instead of `id` to get the value

rzueger:


packages/app-extensions/src/form/formBuilder.spec.js
```diff
@@ -79,6 +79,7 @@ const testData = {
                       {
                         'id': 'firstname',
                         'componentType': 'field',
+                        'path': 'firstname',
```
MethodenMann: wouldn't a path that does not match the id be a better test data?

PR #609, opened 2018-12-27
Pr/list retrieve post

MethodenMann:
Main goal of this PR:
- Add helper to fetch entities via /search POST request with possibility to overwrite endpoint and method
- Add entity count helper 
- Use these helpers in entity list
- clean up fetch options and add better comments.

some of these changes serve as preparation to send selections as query to the server.

Also done:
- fix some problems introduced with type mapping refactoring (#600)
- fix bug where invisible search form fields where not initialized correctly
- transform search form values by type and not by object analysis (and keep fallback for values that are not in the form)

### DIFF
--- a/packages/entity-list/src/util/api/forms.js
+++ b/packages/entity-list/src/util/api/forms.js
@@ -60,19 +60,16 @@ export const getFields = formDefinition => {
   return _uniq(fields)
 }
 
-export const getFormFieldFlat = formDefinition => {
-  return searchChildren(formDefinition)
-}
+export const getFormFieldFlat = formDefinition =>
+  searchChildren(formDefinition)
 
 const searchChildren = node => {
   if (node.componentType === 'field-set') {
     return node.children.reduce((acc, value) => ({...acc, [value.id]: value.dataType}), {})
-  } else {
-    if (node.children) {
-      return node.children.reduce((acc, value) => ({
-        ...acc,
-        ...searchChildren(value)
-      }), {})
-    }
+  } else if (node.children) {
+    return node.children.reduce((acc, value) => ({
+      ...acc,
+      ...searchChildren(value)
+    }), {})
   }
 }


packages/tocco-util/src/mockData/entities.js
```diff
@@ -160,19 +185,33 @@ const createEntityResponse = (entityName, entityStore) => {
   }
 }
 
+const extractUrlParams = url => () => {
+  return {
```
rzueger: can be written without `return`

packages/entity-list/src/util/api/fetchOptions.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+  } else {
+    const objectBased = objectTransformFallback(value, key)
+    if (objectBased) return objectBased
+  }
+
+  return {[key]: {value}}
+}
+
+export const getFetchOptionsFromSearchForm = (searchFormValues, formFields = {}) => {
```
rzueger: can be written without `return`

packages/entity-list/src/modules/list/sagas.js
```diff
@@ -112,26 +119,27 @@ export function* setSorting() {
 }
 
 export function* getSearchFilter(inputSearchFilters, searchInputsFilters) {
-  const filters = yield call(_union, inputSearchFilters, searchInputsFilters)
-  return yield call(_join, filters, ',')
+  return yield call(_union, inputSearchFilters, searchInputsFilters)
```
rzueger: can be written without `return`

packages/app-extensions/src/rest/helpers.js
```diff
@@ -85,34 +106,66 @@ export const defaultFormTransformer = json => (json.form)
...
+    ...(Array.isArray(relations) ? {'relations': relations.length === 0 ? '!' : relations} : {}),
+    ...(filters ? {'filter': filters} : {})
   }
+)
 
-  if (limit) {
-    params._limit = limit
-    if (page) {
-      params._offset = (page - 1) * limit
+export const queryObjectToUrlParams = queryObject => (
```
rzueger: can be written without `(`...

packages/app-extensions/src/rest/helpers.js
```diff
@@ -15,31 +17,50 @@ import {call} from 'redux-saga/effects'
...
- * - query {String} TQL query
+ * @param options {Object} see 'buildEntityQueryObject' function
  * @param transformer {function} Function to directly manipulate the result. By default returns data attribute
+ * @param endpoint {String} To overwrite default endpoint entities/entityName/search
+ * @param method {String} To overwrite default http method 'POST'
  */
-export function* fetchEntities(entityName, options, transformer = defaultEntityTransformer) {
-  const queryParams = yield call(buildParams, options)
-  const resp = yield call(requestSaga, `entities/${entityName}`, {queryParams})
+export function* fetchEntities(entityName, options, transformer = defaultEntityTransformer, endpoint, method = 'POST') {
```
rzueger: - quite a lot method parameters. shouldn't become any more than 5, IMO
- or is there a way to reduce the number of parameters?
- can we rename `options` to something a little bit more specific?

packages/entity-list/src/modules/list/sagas.js
MethodenMann: generator function....

packages/entity-list/src/util/api/forms.js
```diff
@@ -59,3 +59,20 @@ export const getFields = formDefinition => {
 
   return _uniq(fields)
 }
+
+export const getFormFieldFlat = formDefinition => {
+  return searchChildren(formDefinition)
```
rzueger: (superfluous `return`)

packages/entity-list/src/util/api/forms.js
```diff
@@ -59,3 +59,20 @@ export const getFields = formDefinition => {
...
+}
+
+const searchChildren = node => {
+  if (node.componentType === 'field-set') {
+    return node.children.reduce((acc, value) => ({...acc, [value.id]: value.dataType}), {})
+  } else {
+    if (node.children) {
```
rzueger: (could be simplified to `else if ...`)

PR #610, opened 2019-01-04
refactor(entity-list): toggle to show only selected records

MethodenMann:


packages/entity-list/src/components/SelectionController/SelectionController.js
```diff
@@ -29,7 +34,9 @@ SelectionController.propTypes = {
   clearSelection: PropTypes.func.isRequired,
   selection: PropTypes.array.isRequired,
   setSelectionMode: PropTypes.func.isRequired,
-  selectionMode: PropTypes.string
+  toggleShowSelectedRecords: PropTypes.func.isRequired,
+  selectionMode: PropTypes.string,
```
rzueger: I know that the `selectionMode` already existed before, but I personally don't really like that it accepts any string. Can we change this to `oneOf` (apparently, the allowed modes are defined in `selectionModes`)? Furthermore, I'd replace the `'selection'` literal with `selectionModes.SELECTION` on line 14.

PR #613, opened 2019-01-11
Pr/select story

MethodenMann:


packages/tocco-ui/src/Select/Select.stories.js
```diff
@@ -0,0 +1,93 @@
...
+          loadTooltip={this.loadToolTip}
+        />
+      </span>
+    )
+  }
+}
+
+Story.propTypes = {
+  isMulti: PropTypes.bool
+
```
khanlifts: Empty line can be removed.

packages/tocco-ui/src/Select/Select.stories.js
```diff
@@ -0,0 +1,93 @@
...
+  {key: 5, display: 'Option 5'}
+]
+
+class Story extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      isLoading: false,
+      options: null,
+      value: [{key: 3, display: 'Option 3.'}],
```
khanlifts: I think the initial option string should be "Option 3" without the dot.

PR #615, opened 2019-01-14
Util stories

MethodenMann:
https://tocco.github.io/tocco-client/pr_util-stories

### DIFF
--- a/packages/app-extensions/src/notifier/README.md
+++ b/packages/app-extensions/src/notifier/README.md
@@ -1,0 +1,109 @@
+# Notifier
+
+
+## Usage
+### Import
+```javascript
+import {notifier} from  'tocco-app-extensions'
+```
+
+### Setup
+Firstly the notifier util has be added to the store:
+```javascript
+notifier.addToStore(store, accept)
+```
+
+The first parameter is the redux store. The store is needed to adding its reducers and add saga middleware.
+The second parameter (accept) determines whether the app is handling the notifier action itself or just emits the
+actions to the parent app. The latter also needs the `emmitAction` util added to the store.
+
+Secondly the Notifier Components must be added to dom:
+
+```javascript
+ render() {
+    return (
+      <Provider store={store}>
+        <div>
+          <notifier.Notifier/>
+          <...>
+        </div>
+     </Provider>
+     )
+}
+```
+
+### Show notifications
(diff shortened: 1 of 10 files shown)


packages/app-extensions/src/notifier/notifier.stories.js
```diff
@@ -0,0 +1,180 @@
...
+import {Provider} from 'react-redux'
+import {Button, ButtonGroup, Typography} from 'tocco-ui'
+import {withReadme} from 'storybook-readme'
+
+import Readme from './README.md'
+import notifier from './'
+import appFactory from '../appFactory'
+
+const longText = `Lorem ipsum dolor sit amet, at sed inermis intellegam scriptorem, usu facete apeirian ad.
+Sit et meliore intellegam. Mel cu maluisset philosophia, pri et habeo oportere. Vis in purto verear luptatum, has
```
khanlifts: I think you should indent line 14/15.

packages/app-extensions/src/notifier/notifier.stories.js
```diff
@@ -0,0 +1,180 @@
...
+
+class Story extends React.Component {
+  constructor(props) {
+    super(props)
+    this.store = appFactory.createStore({}, undefined, {}, 'notifier')
+    notifier.addToStore(this.store, true)
+  }
+
+  info1 = () => {
+    this.store.dispatch(notifier.info(
```
khanlifts: Can be written in one line like info2.

packages/app-extensions/src/notifier/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Setup
+Firstly the notifier util has be added to the store:
+```javascript
+notifier.addToStore(store, accept)
+```
+
+The first parameter is the redux store. The store is needed to adding it reducers and add saga middleware.
+The second parameter (accept) determines whether the app is handling the notifier action itself or just emits the
+actions to the parent app. Latter also needs the `emmitAction` util added to the store.
```
khanlifts: Same as in errorLogging README.

packages/app-extensions/src/notifier/notifier.stories.js
MethodenMann: Thats a multi line string.

PR #620, opened 2019-01-17
refactor(tocco-ui): replace bootstraps form-control on DurationEdit

niklaushug:


packages/tocco-ui/src/EditableValue/typeEditors/StyledDurationEdit.js
```diff
@@ -1,28 +1,44 @@
 import styled from 'styled-components'
 
+import {spaceScale} from '../../utilStyles'
+import {StyledSpan} from '../../Typography'
 import {
-  declareFont,
-  spaceScale
-} from '../../utilStyles'
+  StyledEditableWrapper,
+  StyledInputCss
```
khanlifts: Can be written in one line.

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
```diff
@@ -3,7 +3,11 @@ import React from 'react'
 
 import {calculateMilliseconds} from '../utils'
 import Typography from '../../Typography'
-import StyledDurationEdit from './StyledDurationEdit'
+import {
+  StyledDurationEditWrapper,
+  StyledDurationEdit
```
khanlifts: Can be written in one line.

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
```diff
@@ -80,33 +84,33 @@ class DurationEdit extends React.Component {
...
-          value={this.state.hours}
-          onInput={this.handleHourChange}
-          onKeyPress={this.preventNonNumeric}
-          disabled={this.props.readOnly}
-          pattern="\d+"
-          min={0}
-        />
+      <StyledDurationEditWrapper>
+        <StyledEditableWrapper>
+          <StyledDurationEdit
```
khanlifts: Would it be simpler to write the styling for DurationEdit in one instead of two wrappers? We would then have StyledEditableWrappers and then StyledDurationEdit without the StyledDurationEditWrapper.

packages/tocco-ui/src/EditableValue/typeEditors/StyledDurationEdit.js
niklaushug: I prefer readable over compact code. Code Diffs are line-based, hence it is easier to recognize what have changed if each import occupies one line. Single statements per line are easily ordered alphabetically, which can be scanned better.

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
niklaushug: Good point to reason about. It would be simpler but I could not achieve what I have intended to do. 
- StyledDurationEditWrapper is responsible for the horizontal layout and unique for DurationEdit. 
- StyledEditableWrapper is very generic and used in almost all editable fields. It is responsible to place buttons next to input fields and to emphasize a strong connection.  
- StyledDurationEdit is generic and used in a similar way in almost all editable fields. Only an additional attribute is defined. It is responsible to harmonize and to reset input field declarations.
It would be possible to remove one wrapper if only one Input field must have been wrapped.

PR #624, opened 2019-01-25
Pr/sso styling

niklaushug:


packages/tocco-ui/src/utilStyles/declareInteractionColors.js
```diff
@@ -1,3 +1,4 @@
+
```
MethodenMann: remove

packages/tocco-ui/src/utilStyles/declareInteractionColors.js
```diff
@@ -114,20 +115,32 @@ const generateRaisedPrimaryColors = props => {
...
-  const higherContrast = getContrast(text, primary) > getContrast(paper, primary) ? text : paper
+  const higherContrast = getHigherContrast(primary, text, paper)
   const fg = generateInteractionColor(higherContrast, {
     action: getLuminance(primary) > 0.5 ? 'darken' : 'lighten'
   })
   const bg = generateInteractionColor(primary)
   return mapColors(bg, fg)
 }
 
+const generateCustomColors = (base, colorA, colorB) => {
```
MethodenMann: i had a hard time understanding this code. I think it's mainly because if i look at this method signature:

`const generateCustomColors = (base, colorA, colorB) `

i would have absolutely no clue what to pass in and what to expect as return value. Even if i look at the implementation couldn't guess what `mapColors` is doing.
Maye you could improve naming?

PR #627, opened 2019-01-29
Pr/storybook

MethodenMann:


deployment/deploy_storybook.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+set -e # Exit with nonzero exit code if anything fails
+PAGES_BRANCH="gh-pages"
+
+if [ "$TRAVIS_PULL_REQUEST" != "false"  ]
```
khanlifts: Is the double negation required? Why not `if [ "$TRAVIS_PULL_REQUEST" == "true"  ]`?

packages/app-extensions/src/errorLogging/README.md
```diff
@@ -0,0 +1,42 @@
...
+Console: Logs the error to the console
+Remote: Sends the error to the nice2 error logger endpoint (nice2/log)
+Notifier: Displays error in UI. Notifier needs to be added to the store if handler is active.
+
+### Log Error
+```javascript
+store.dispatch(errorLogging.logError('client.errorTitle', 'client.errorDescription', new Error('Some Error')))
+```
+
+
```
khanlifts: Are twelve empty lines needed?

packages/app-extensions/src/notifier/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+## Usage
+### Import
+```javascript
+import {notifier} from  'tocco-app-extensions'
+```
+
+### Setup
+Firstly the notifier util has be added to the store:
```
khanlifts: Firstly the notifier util has **to** be added to the store:

packages/app-extensions/src/notifier/README.md
```diff
@@ -0,0 +1,109 @@
...
+```javascript
+notifier.addToStore(store, accept)
+```
+
+The first parameter is the redux store. The store is needed to adding its reducers and add saga middleware.
+The second parameter (accept) determines whether the app is handling the notifier action itself or just emits the
+actions to the parent app. The latter also needs the `emmitAction` util added to the store.
+
+Secondly the Notifier Components must be added to dom:
```
khanlifts: Notifier should start with N or n throughout the file.

PR #632, opened 2019-02-01
fix: rename specs files to spec

MethodenMann:
- fix failing tests


packages/scheduler/src/components/FullCalendar/FullCalendar.spec.js
```diff
@@ -7,13 +7,15 @@ import FullCalendar from './FullCalendar'
 describe('scheduler', () => {
   describe('components', () => {
     describe('Fullcalendar', () => {
+      const baseProps = {onRefresh: () => {}}
+
       it('should render wrapping div with id and calendar div', () => {
```
khanlifts: `it` can be replaced with `test` in all tests

PR #641, opened 2019-02-15
style(tocco-ui): enhance visualization of disabled state for button

niklaushug:
- do not change color on hover
- show cursor not-allowed on hover
- add knobs to button showcase
- remove obsolete buttons from showcase

### DIFF
--- a/packages/tocco-ui/src/Button/Button.stories.js
+++ b/packages/tocco-ui/src/Button/Button.stories.js
@@ -11,98 +11,93 @@ storiesOf('Button', module)
   .addDecorator(withKnobs)
   .add(
     'Knobs',
-    () =>
-      (
-        <Button
-          label={text('label', 'My Button')}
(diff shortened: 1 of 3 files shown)


packages/tocco-ui/src/Button/Button.stories.js
```diff
@@ -41,61 +41,62 @@ storiesOf('Button', module)
     () => {
       return [
         <Button
+          dense={boolean('dense', false) || undefined}
```
MethodenMann:
```
const knobs = {dense: boolean('dense', false) || undefined, disabled: boolean('disabled', false) || undefined}

<Button
  ...knobs
  key="1"
```

So you dont have to repeat yourself that often

niklaushug: see 88008f1

PR #648, opened 2019-03-04
refactor: move upload to app-extensions

MethodenMann:
- This refactoring helps to eliminate duplicate code in the packages and is the groundwork for an upcoming refactoring.
  The goal is to move all "form utils" to formData so that they can be provided with a container.


packages/app-extensions/src/formData/upload/sagas.spec.js
```diff
@@ -0,0 +1,46 @@
...
+
+import {documentToFormValueTransformer, uploadRequest} from './documents'
+import * as actions from './actions'
+import * as sagas from './sagas'
+
+import {call, fork, put, takeEvery} from 'redux-saga/effects'
+
+describe('app-extensions', () => {
+  describe('formData', () => {
+    describe('values', () => {
```
rzueger: `upload` instead of `values`

PR #651, opened 2019-03-05
feat(tocco-ui): helper method to bypass groups of attributes

niklaushug:


packages/tocco-ui/src/utilStyles/filterObjectByKeysStartingWith.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Enforce that all values of a flat object are used as keys in another object.
+ * May insert {...filterObjectByKeysStartingWith(props)} function in JSX.
+ *
+ * @param  {object} props       Pass in React props
+ * @param  {array} identifier   Define which key/value pairs are returned
+ * @return {object}             Filtered props
+ */
+
+const filterObjectByKeysStartingWith = (props = {}, identifiers = ['data-', 'aria-']) => {
```
MethodenMann: very intransparent default value. At least add to documentation

niklaushug: The default was picked according the demands of a single use case. Probably is an empty array for identifiers the better default?

In which documentation I have to describe the defaults (packages/tocco-ui/README.md or packages/tocco-ui/src/utilStyles/filterObjectByKeysStartingWith.js)?

MethodenMann: Yes probably add an empty array as default value. A Bit of an anti pattern to set default values according to single usage. 

No with documentation i meant on line 8 in method description

PR #661, opened 2019-03-15
feat(app-extensions): map longitude and latitude to string

khanlifts:


packages/app-extensions/src/formField/editableValueFactory.js
```diff
@@ -9,11 +9,17 @@ const settings = {
...
+  'coordinate': value => value.value
+}
+
 export default type =>
-  (formField, modelField, formName, props, events, utils) => {
+  (formField, modelField, formName, data, events, utils) => {
     const options = getOptions(type, formField, modelField, utils, formName)
 
-    return <EditableValue type={type} events={events} {...props} options={options}/>
+    data.value = valueOverwriter[type] ? valueOverwriter[type](data.value) : data.value
```
MethodenMann: add an if. to set `data.value = data.value` does not make sense.

PR #662, opened 2019-03-19
refactor(tocco-ui): use theme for form labels

niklaushug:


MethodenMann:
Is the mandatory asterisk black on purpose?

All the dirty fields are colored. Before this change it was green-ish. Now it's the warning orange i guess. Apart from that this orange hurts my eyes i think  "warning" orange sends a wrong message. Dirty fields are a not negative in any way.

Before:
image


After:
image


While you're at it i think you can add a small indent to the error list.

packages/tocco-ui/src/FormField/StyledFormField.js
```diff
@@ -0,0 +1,35 @@
+import styled, {css} from 'styled-components'
+import {theme} from 'styled-system'
+import classNames from 'classnames'
+
+import {spaceScale} from '../utilStyles'
+import {StyledSignalList} from '../SignalList'
+
+const markAsMandatory = () => css`
+  label:after {
```
MethodenMann: see comment in PR

PR #664, opened 2019-03-20
refactor(app-extensions): specify ToccoTheme as default

niklaushug:
Constant mergedTheme guarantees already that ToccoTheme is available. 

By adding theme as default prop only the console warning is eliminated.


packages/app-extensions/src/appFactory/ThemeWrapper.js
```diff
@@ -19,6 +19,10 @@ class ThemeWrapper extends React.PureComponent {
   }
 }
 
+ThemeWrapper.defaultProps = {
```
MethodenMann: good. but you should adjust line 11:

`    const mergedTheme = _merge({}, theme, appTheme)`

PR #665, opened 2019-03-21
Pr/refactor declare interaction color

niklaushug:


packages/sso-login/src/components/ProviderButton/StyledProviderButton.js
```diff
@@ -3,18 +3,18 @@ import {theme} from 'styled-system'
...
     display: inline-block;
     width: 100%;
     ${props => declareInteractionColors(
-    generateCustomColors(
-      props.primaryColor,
-      props.secondaryColor || theme('colors.paper')(props),
-      props.secondaryColor || theme('colors.text')(props)))}
+    generateInteractionColors(props, {
+      bg: props.primaryColor,
+      fg: props.secondaryColor || `${theme('colors.paper')(props)}, ${theme('colors.text')(props)}`
```
MethodenMann: dont forget to edit this line as soon as you rebased. (you removed theme dependency)

packages/tocco-ui/src/Button/StyledButton.js
```diff
@@ -43,22 +40,22 @@ const meltButtons = props => {
 }
 
 const declareButtonColor = props => {
-  let declareColor
+  let scheme
   const {ink, look} = props
   const {BALL, FLAT, RAISED} = stylingLook
```
MethodenMann: could be done outside of method

packages/tocco-ui/src/utilStyles/declareInteractionColors.spec.js
```diff
@@ -169,7 +108,75 @@ describe('tocco-ui', () => {
...
+          }
+        }, {
+          scheme: 'raisedPrimary',
+          expectation: {
+            bg: ['#600', '#900', '#c00'],
+            fg: ['#900', '#c00', '#f00']
+          }
+        }, {
+          scheme: {
+            bg: 'colors.primary',
```
MethodenMann: IMO it would be more intuitive to prefix everything that's not a color?
`fg: '#234, $colors.text, black'`

what do you think? You could remove dependency and just check for the $

niklaushug: Using `color.startsWith('colors.')` instead of prefixing theme paths for colors by $ prevents to introduce a special case for a few color paths.

PR #667, opened 2019-03-22
Pr/floating label

niklaushug:
Current implementation requires a polyfill for css pseudo selector `:focus-within` like https://github.com/jonathantneal/focus-within. Microsoft browser does not support it yet.

`:focus-within` makes code much cleaner and lighter than using a react library like https://github.com/ferdaber/react-focus-within.

Do we already other polyfills included in tocco-client? What do you think about polyfills in general?

### DIFF
--- a/packages/tocco-ui/src/StatedValue/StatedValue.js
+++ b/packages/tocco-ui/src/StatedValue/StatedValue.js
@@ -25,7 +25,7 @@ const detectSignal = (dirty, hasError) => {
  */
 const StatedValue = props => {
   const {
-    content,
+    children,
     description,
     dirty,
     error,
@@ -44,7 +44,7 @@ const StatedValue = props => {
       <StyledStatedValueBox
         hasValue={hasValue}
         signal={signal}>
-        {content}
+        {children}
         <StyledStatedValueLabel
           hasValue={hasValue}
           signal={signal}
@@ -66,7 +66,7 @@ StatedValue.propTypes = {
   /**
    * A component to enter or display data.
    */
-  content: PropTypes.node,
+  children: PropTypes.node,
   /**
    * A helper text to instruct users.
    */
--- a/packages/tocco-ui/src/StatedValue/StatedValue.js
+++ b/packages/tocco-ui/src/StatedValue/StatedValue.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types'
 import React from 'react'
+import FocusWithin from 'react-simple-focus-within'
 
 import ErrorList from '../FormField/ErrorList'
 import {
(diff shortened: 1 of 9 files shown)


packages/tocco-ui/src/StatedValue/StatedValue.js
```diff
@@ -0,0 +1,106 @@
...
+  } else if (dirty) {
+    return stylingCondition.WARNING
+  }
+}
+
+/**
+ * Wrap <EditableValue> and <FormattedVaule> in <StatedValue> to describe it
+ * and signal conditions.
+ */
+const StatedValue = props => {
```
MethodenMann: I dont know if i should like the name "StatedValue". Maybe we can discuss it. But isnt it just a new made "FormField"?

packages/tocco-ui/src/StatedValue/StyledStatedValue.js
```diff
@@ -0,0 +1,130 @@
...
+import {
+  declareFocus,
+  declareFont,
+  fontScale,
+  shadeColor,
+  spaceScale
+} from '../utilStyles'
+
+/*
+TODO
```
MethodenMann: i also prefer the polifilly over the react componnent. But shouldn't we load the polyfill with webpack.

MethodenMann: https://github.com/jonathantneal/focus-within
it this a "real" polyfill? Its important is only overwrites browser behavior if the the browser not supports the feature.

packages/tocco-ui/src/StatedValue/StatedValue.js
niklaushug: FormFieldFloat

packages/tocco-ui/src/StatedValue/StatedValue.js
```diff
@@ -0,0 +1,106 @@
+import PropTypes from 'prop-types'
```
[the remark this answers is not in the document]
niklaushug: see 70ae957362a4d0539ab616c90d55eefeb7a55c94

packages/tocco-ui/src/StatedValue/StatedValue.js
```diff
@@ -0,0 +1,106 @@
...
+      }
+    </StyledStatedValueWrapper>
+  )
+}
+
+StatedValue.propTypes = {
+  /**
+   * A component to enter or display data.
+   */
+  content: PropTypes.node,
```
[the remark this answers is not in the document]
niklaushug: see 70ae957362a4d0539ab616c90d55eefeb7a55c94

packages/tocco-ui/src/StatedValue/StyledStatedValue.js
niklaushug: see 2e35cdf0404838189dd0b804937b6ede184ed8e4

PR #669, opened 2019-03-26
List Fixes

MethodenMann:


packages/entity-list/src/modules/list/sagas.js
```diff
@@ -166,15 +166,30 @@ export function* fetchEntitiesAndAddToStore(page) {
...
+  yield spawn(preloadNextPage, page)
+}
+
+export function* preloadNextPage(currentPage) {
+  const list = yield select(listSelector)
+  const {entityStore, limit} = list
+  let {entityCount} = list
+  const nextPage = currentPage + 1
+
+  if (entityCount === null) {
```
rzueger: what about undefined? maybe this should be considered as well to make that line a little bit more stable?

PR #671, opened 2019-03-29
Pr/pass props automatically to styling function

niklaushug:


packages/tocco-ui/src/Button/StyledButton.js
```diff
@@ -79,9 +79,9 @@ const declareBall = props => {
 
       // ensure that width has at least the size of height
       min-width: calc(1rem
-        * ${_get(props.theme, 'fontSize.base')}
-        * ${_get(props.theme, 'lineHeights.regular')}
-        + 2 * ${scale.space(props.theme, -3)});
+        * ${theme.fontSize('base')(props)}
```
MethodenMann: I think this notation could be optimized if we use tagged template literals like styled components does.

https://celsiusnotes.com/javascript-template-literals/

Attached is a patch where i prototyped it.
[theme_with_tagged_template_literals.patch.zip](https://github.com/tocco/tocco-client/files/3028101/theme_with_tagged_template_literals.patch.zip)

niklaushug: see ffbc67e

PR #673, opened 2019-04-10
fix(tocco-ui): hide DateAbstract delete Button

khanlifts:
- replace es5 with es6 functions in DateAbstract Component
- fix editable value coordinate valueOverwriter

### DIFF
--- a/packages/app-extensions/src/formField/editableValueFactory.js
+++ b/packages/app-extensions/src/formField/editableValueFactory.js
@@ -10,7 +10,7 @@ const settings = {
 }
 
 const valueOverwriter = {
-  'coordinate': value => value.value
+  'coordinate': value => _get(value, 'value.value', '')
 }
 
 export default type =>
--- a/packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
+++ b/packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
@@ -111,28 +111,30 @@ class DateAbstract extends React.Component {
   }
 
   handleButtonVisibility = () => {
-    if (this.hasValue() && !this.props.readOnly) {
-      this.setState({hideButton: false})
-    } else {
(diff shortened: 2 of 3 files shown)


packages/app-extensions/src/formField/editableValueFactory.js
```diff
@@ -10,7 +10,7 @@ const settings = {
 }
 
 const valueOverwriter = {
-  'coordinate': value => value.value
+  'coordinate': value => _get(value, 'value.value', '')
```
MethodenMann: wrong commit

packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -84,32 +86,42 @@ class DateAbstract extends React.Component {
...
+  handleOnBlur = () => {
     if (this.props.onBlur) {
       const altValue = this.flatpickr.altInput.value
       this.props.onBlur(altValue, this.flatpickr.selectedDates, r => this.flatpickr.setDate(r, true))
     }
   }
 
-  render() {
-    const showClearButton = this.hasValue && !this.props.readOnly
+  handleButtonVisibility = () => {
```
MethodenMann:
```
 handleButtonVisibility = () => {
      this.setState({hideButton: this.hasValue() && !this.props.readOnly})
}
```

packages/tocco-ui/src/EditableValue/typeEditors/StyledDateAbstract.js
```diff
@@ -10,6 +10,9 @@ const StyledDateAbstractControl = styled.div.attrs({
   'data-clear': true
```
MethodenMann: since you are at it i would move attributes like data-clear out of the styled components. is has nothing to do with styling. instead add those to the elements itself.

PR #674, opened 2019-04-11
fix(entity-detail): fix field touched bug in DetailForm

khanlifts:


packages/entity-detail/src/components/DetailForm/DetailForm.js
```diff
@@ -12,8 +12,10 @@ import readOnlyFormFieldMapping from '../../util/detailView/readOnlyFormFieldMap
 import StyledDetailForm from './StyledDetailForm'
 
 export class DetailForm extends React.Component {
-  componentWillReceiveProps(nextProps) {
-    this.props.fireTouched(nextProps.anyTouched)
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    if (prevProps !== this.props) {
```
MethodenMann: what is the point of this `if`?

packages/entity-detail/src/components/DetailForm/DetailForm.js
```diff
@@ -12,8 +12,10 @@ import readOnlyFormFieldMapping from '../../util/detailView/readOnlyFormFieldMap
 import StyledDetailForm from './StyledDetailForm'
 
 export class DetailForm extends React.Component {
-  componentWillReceiveProps(nextProps) {
-    this.props.fireTouched(nextProps.anyTouched)
+  componentDidUpdate(prevProps, prevState, snapshot) {
```
MethodenMann: snapshot can be omitted

packages/entity-detail/src/components/DetailForm/DetailForm.js
```diff
@@ -128,6 +130,7 @@ DetailForm.propTypes = {
   anyTouched: PropTypes.bool,
   formErrors: PropTypes.object,
   valid: PropTypes.bool,
+  isDirty: PropTypes.bool,
```
MethodenMann: the save button uses "anyTouched" as well. `isDirty` could there be more suitable as well. Remove any unused properties at the end.

packages/entity-detail/src/components/DetailForm/DetailForm.js
khanlifts: standard componendDidUpdate check to control the calls, prevents infinite loops as well.

packages/entity-detail/src/components/DetailForm/DetailForm.js
```diff
@@ -91,9 +93,9 @@ export class DetailForm extends React.Component {
           />
           {!this.isReadOnlyForm()
           && <div>
-            {!props.valid && props.anyTouched && <ErrorBox formErrors={props.formErrors} showErrors={this.showErrors}/>}
+            {!props.valid && props.isDirty && <ErrorBox formErrors={props.formErrors} showErrors={this.showErrors}/>}
```
MethodenMann: Not quite perfect. 
Reproduce problem:
In the mocked form touch, but not edit, the "Callname" Field. Save button is not disabled but error on field is shown.

image

PR #677, opened 2019-04-12
refactor(tocco-ui): extend story of EditableValue

niklaushug:
- add knob to toggle readOnly
- order props alphabetically


packages/tocco-ui/src/EditableValue/EditableValue.stories.js
```diff
@@ -56,118 +58,151 @@ storiesOf('EditableValue', module)
...
+  ).add(
     'Single-Select',
-    () => <SelectStory
-      isMulti={false}
-      delay={0}
-    />
-  )
-  .add(
+    () =>
+      <SelectStory
```
khanlifts: `<SelectStory />` needs `readOnly` set to its `<Select />` child component too. Right now readOnly on all SelectStories don't work.

PR #688, opened 2019-04-26
feat(tocco-ui): add hide links on readOnly

khanlifts:
- add tests to location, phone, url

### DIFF
--- a/packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
+++ b/packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
@@ -133,7 +133,7 @@ class DateAbstract extends React.Component {
         />
         <StyledDateAbstractControl
           data-clear
-          hideButton={this.state.hideButton}
+          hideButton={this.state.hideButton || this.props.readOnly}
         >
           <Button
             icon="times"


packages/tocco-ui/src/EditableValue/typeEditors/LocationEdit.js
```diff
@@ -54,7 +54,7 @@ class LocationEdit extends React.Component {
 
   onSuggestionsClearRequested = () => {}
 
-  showGoogleMaps = value => value && Object.values(value).some(val => val)
+  showGoogleMaps = value => !this.props.readOnly && value && Object.values(value).some(val => val)
```
MethodenMann: Why? Opening google maps ist a read action and should be possible in read-only.

packages/tocco-ui/src/EditableValue/typeEditors/UrlEdit.js
```diff
@@ -36,7 +36,7 @@ const StringEdit = props => {
         id={props.id}
         disabled={props.readOnly}
       />
-      {value && <StyledEditableControl>
+      {!props.readOnly && value && <StyledEditableControl>
```
MethodenMann: Same here. Calling a number should be possible in read only mode.

PR #689, opened 2019-04-26
Pr/prepare duration edit for stated value

niklaushug:


packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
```diff
@@ -82,10 +85,28 @@ class DurationEdit extends React.Component {
     }
   }
 
+  componentDidMount = () => {
```
khanlifts: Move lifecycle methods below the constructor.

khanlifts: Move lifecycle methods below the constructor.

PR #690, opened 2019-05-02
Pr/storybook update

MethodenMann:


packages/login/src/login.stories.js
```diff
@@ -1,37 +1,57 @@
...
   )
 
-class LoginAppWrapper extends React.Component {
+class LoginAppWrapperStory extends React.Component {
   constructor(props) {
     super(props)
     this.childKey = 0
   }
 
+  confirmed = false
```
rzueger: why is that done this way? and what about `childKey`? looks a bit hacky to me.

MethodenMann: Sure looks hacky. that's because it is ;)
First of all we talking about a story. So none productive code. I think hacky is ok in this case.

We need to ensure that the component is entirely rerendered on knob (input) changes and changing the key prop a easy solution for this problem. Also see:
https://medium.com/@albertogasparin/forcing-state-reset-on-a-react-component-by-using-the-key-prop-14b36cd7448e 

I think the whole discussion about how to properly change input params and reload the app needs to be done as soon as we ship our apps to be used independently (CMS widgets and so on).

PR #696, opened 2019-05-08
Pr/editable fixes

MethodenMann:


packages/tocco-ui/src/EditableValue/EditableValue.stories.js
```diff
@@ -31,251 +17,306 @@ export class EditableValueStory extends React.Component {
...
-    'Url',
+    'Date-Range',
     () =>
       <EditableValueStory
-        ChildComponent={UrlEdit}
-        readOnly={boolean('readOnly', false) || undefined}
-        value={'http://www.tocco.ch'}
+        type="date-range"
+        knobType={object}
+        defaultValue={{from: '2019-12-21', to: '2019-12-23'}}
```
khanlifts: `defaultValue={{from: '2019-12-05T23:00:00.000Z', to: '2019-12-06T23:00:00.000Z'}}`
This matches the actual value formats when using date-range.

MethodenMann: i will leave it as it is. We need to change the backend first before we can fully integrate the date-range field

PR #698, opened 2019-05-09
feat(tocco-ui): emphasize if StatedValue contains uneditable

niklaushug:
- StatedValue can contain conten which is edtiable, temporary uneditable or never editable.
- visualize temporary state by lighter colors
- visualize mutuables by border
- visualize immutables by borderless

### DIFF
--- a/packages/tocco-ui/src/StatedValue/StatedValue.js
+++ b/packages/tocco-ui/src/StatedValue/StatedValue.js
@@ -35,27 +35,33 @@ const StatedValue = props => {
     label,
     mandatory,
     immutable,
+    type,
     touched
   } = props
 
   const hasError = touched && props.error && Object.keys(props.error).length > 0
   const signal = detectSignal(dirty, hasError)
+  const look = type === 'display'
+    ? 'display'
+    : immutable
+      ? 'immutableField'
+      : 'mutableField'
 
   return (
     <FocusWithin>
       {({focused, getRef}) => {
-        const secondaryPosition = focused || hasValue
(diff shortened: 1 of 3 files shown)


packages/tocco-ui/src/StatedValue/StatedValue.js
```diff
@@ -35,27 +35,33 @@ const StatedValue = props => {
...
             secondaryPosition={secondaryPosition}>
             <StyledStatedValueBox
-              immutable={immutable}
+              look={look}
               signal={signal}>
               {children}
               <StyledStatedValueLabel
-                htmlFor={immutable ? '' : id}
-                immutable={immutable}
+                htmlFor={look === 'mutableField' ? id : ''}
```
MethodenMann: {...look === 'mutableField' ? { htmlFor:  id}} : {}}  

Like this you can omit the property completely if not needed.

packages/tocco-ui/src/StatedValue/StatedValue.stories.js
```diff
@@ -59,8 +59,8 @@ class StatedValueWrapper extends React.Component {
           onChange={this.handleChange}
           style={{
             border: 0,
-            cursor: knobs.immutable ? 'not-allowed' : 'auto',
-            color: knobs.immutable ? '#909090' : '#000',
+            cursor: knobs.immutable && knobs.type === 'field' ? 'not-allowed' : 'auto',
```
MethodenMann: Chould you change the stories title so it wont have a separate chapter in storybook:
image

PR #699, opened 2019-05-09
feat(tocco-ui): show unit in DurationEdit if not disturbing

niklaushug:
- Label of StatedValue covers units of DurationEdit which looks ugly and is confusing.
- Show units only if focus is on DurationEdit or if hours or minutes are set.

### DIFF
--- a/packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
+++ b/packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
@@ -16,7 +16,8 @@ class DurationEdit extends React.Component {
     this.hoursShadow = React.createRef()
     this.minutesShadow = React.createRef()
     this.state = {
-      ...this.millisecondsToDuration(props.value)
+      ...this.millisecondsToDuration(props.value),
+      showUnits: props.value >= 0
     }
   }
 
@@ -95,6 +96,11 @@ class DurationEdit extends React.Component {
     this.props.onChange(calculateMilliseconds(hoursValue, minutesValue))
   }
 
+  handleOnBlur = () => this.setState({showUnits:
+    (this.state.hours.toString().length >= 1 || this.state.minutes.toString().length >= 1)})
+
+  handleOnFocus = () => this.setState({showUnits: true})
+
   preventNonNumeric = event => {
     if (!(event.charCode >= 48 && event.charCode <= 57)) {
       event.preventDefault()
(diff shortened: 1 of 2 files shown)


packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
```diff
@@ -16,7 +16,8 @@ class DurationEdit extends React.Component {
     this.hoursShadow = React.createRef()
     this.minutesShadow = React.createRef()
     this.state = {
-      ...this.millisecondsToDuration(props.value)
+      ...this.millisecondsToDuration(props.value),
+      focus: props.value >= 0
```
MethodenMann: focus is true when there is a value >= 0
That sounds confusing. Focus should be true when the field is focused. Maybe call the state `showUnits`?

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
```diff
@@ -95,6 +96,9 @@ class DurationEdit extends React.Component {
     this.props.onChange(calculateMilliseconds(hoursValue, minutesValue))
   }
 
+  handleOnBlur = () => this.setState({focus: (this.state.hours >= 1 || this.state.minutes >= 1)})
```
khanlifts: 0 0 
This display without labels can be confusing. Are you sure you want to hide the labels for this case?

MethodenMann: cyril is right. 0 is a valid value.

niklaushug: state.hours and state.minutes can be integer or string. Hence I prefer `this.setState({focus: (this.state.hours.length >= 1 || this.state.minutes.length >= 1)})` over `this.setState({focus: (parseInt(this.state.hours) >= 0 || parseInt(this.state.minutes) >= 0)})`.

packages/tocco-ui/src/EditableValue/typeEditors/DurationEdit.js
```diff
@@ -95,6 +96,9 @@ class DurationEdit extends React.Component {
     this.props.onChange(calculateMilliseconds(hoursValue, minutesValue))
   }
 
+  handleOnBlur = () => this.setState({showUnits: (this.state.hours.length >= 1 || this.state.minutes.length >= 1)})
```
khanlifts: When I execute the following steps only a `1` is shown:
1. Load the component.
2. Delete minutes. (Or hours respectively.)
3. Click outside.
I think there should be `1 hrs 0 min`.

PR #702, opened 2019-05-10
fix(scheduler): fix description HTML display

MethodenMann:
- An event description can contain HTML. Use FormattedValue to proper format
  otherwise HTML would be shown as string.

Changelog: Format HTML event descriptions


MethodenMann:
> Please use for value time as well a FormattedValue (type date, datetime or time).

That's not a typical value for a date time. It uses a a library to generate a readable value for a date range.

e.g.
`10. Mai, 11:20 - 11. Mai, 11:20 `
for a time over several days. Or
`10. Mai, 16:25 - 22:00`
for a range within the same day.

PR #705, opened 2019-05-13
fix(entity-browser): fix LoginApp showTitle to true/false

khanlifts:


packages/login/src/main.js
```diff
@@ -18,11 +18,13 @@ const EXTERNAL_EVENTS = [
...
 
 const initLoginApp = (id, input, events, publicPath, customTheme) => {
+  const showPasswordRequest = !!input.passwordRequest
+
   const actions = [
     passwordUpdate.setShowOldPasswordField(false),
     passwordUpdate.setForcedUpdate(true),
     passwordUpdate.setStandalone(false),
-    passwordRequest.setPasswordRequest(input.passwordRequest),
+    passwordRequest.setPasswordRequest(showPasswordRequest),
```
MethodenMann: i personally would write it directly 

`passwordRequest.setPasswordRequest(!!input.passwordRequest)`

PR #708, opened 2019-05-16
feat(tocco-test-util): establish themed component for testing

niklaushug:
- create theme object according ToccoTheme.defaultTheme
- use ThemeProvider of styled.components
- replace in all tests where a standardized theme can be used


packages/entity-list/src/components/SearchForm/SearchForm.spec.js
```diff
@@ -105,7 +68,7 @@ describe('entity-list', () => {
         }))
 
         const wrapper = intlEnzyme.mountWithIntl(
-          <ThemeProvider theme={theme}>
+          <testTheme.ToccoTheme>
```
MethodenMann: Rename it to something provider.

`TestThemeProvider` ?

PR #711, opened 2019-05-23
feat(tocco-client): add dotenv environment variables

khanlifts:
- add .env.example file
- add .env to gitignore
- install dotenv and dotenv-webpack
- configure config/_base to set secretGlobals


plop/templates/env/env.js
```diff
@@ -0,0 +1,2 @@
+CYPRESS_LOGIN={{CYPRESS_LOGIN}}
+SCHEDULER_LICENCE={{SCHEDULER_LICENCE}}
```
MethodenMann: Why dont you use this var in calendar app?

package.json
```diff
@@ -72,6 +72,8 @@
     "compression": "1.7.2",
     "copy-webpack-plugin": "4.5.2",
     "css-loader": "0.28.11",
+    "dotenv": "^8.0.0",
+    "dotenv-webpack": "^1.7.0",
```
MethodenMann: Remove the package "dotenv-webpack" when its not used. But why isnt it used?

plop/generators/env.js
```diff
@@ -0,0 +1,18 @@
+export default {
+  description: 'Create environment keys in a .env file',
```
MethodenMann: that ain't keys?

plop/generators/env.js
```diff
@@ -0,0 +1,18 @@
+export default {
+  description: 'Create environment variables in a .env file',
```
MethodenMann: I think you should mention that you talk about secret environment variables. There are other ways for adding ordinary environment variables. Or mention it in tocco-docs

config/_base.js
```diff
@@ -34,7 +38,8 @@ config.globals = {
   '__BACKEND_URL__': JSON.stringify(argv.backend) || JSON.stringify(process.env.BACKEND),
   '__PACKAGE__'   : argv.package,
   '__PACKAGE_NAME__'   : JSON.stringify(argv.package),
-  '__NO_MOCK__':  !!(process.env.BACKEND || argv.backend || argv.noMock)
+  '__NO_MOCK__':  !!(process.env.BACKEND || argv.backend || argv.noMock),
+  ...secretGlobals
```
MethodenMann: i think you can remove this code since you are using the webpack plugin

PR #716, opened 2019-05-28
chore: add cypress tests

MethodenMann:


cypress/plugins/index.js
```diff
@@ -0,0 +1,12 @@
+const renameKeys = o =>
+  Object.keys(o || {}).reduce(
+    (acc, key) =>
+      Object.assign({}, acc, {[key.replace(/^CYPRESS_/, '')]: o[key]})
```
khanlifts: Doesn't this work with the spread operator too?

cypress/plugins/index.js
```diff
@@ -0,0 +1,12 @@
+const renameKeys = o =>
```
khanlifts: ES5 or ES6 in cypress directory?

cypress/plugins/index.js
MethodenMann: Nope. not supported in this context

cypress/plugins/index.js
MethodenMann: ES6

PR #726, opened 2019-06-06
chore(tocco-ui): update flatpickr library

khanlifts:


packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -72,17 +73,19 @@ class DateAbstract extends React.Component {
     return null
   }
 
-  componentWillReceiveProps(props) {
-    this.handleButtonVisibility()
-    const locale = this.getLocale(props.intl.locale)
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    if (!_isEqual(prevProps.value, this.props.value)) {
+      this.handleButtonVisibility()
```
MethodenMann: does this have to be so complicated? Storing this simple value in store.

`
   <StyledDateAbstractControl
          data-clear	    
          hideButton={ !this.hasValue() && !this.props.readOnly}	          
        >	       
`

won't work?

packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -72,17 +73,19 @@ class DateAbstract extends React.Component {
...
+      const locale = this.getLocale(this.props.intl.locale)
```
MethodenMann: so if `this.props.intl.locale ` changes this will not update since you wrapped everything in a "only when value changed"-if?

packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -72,24 +72,23 @@ class DateAbstract extends React.Component {
...
-      this.Flatpickr.localize(locale)
-      this.flatpickr.set('locale', locale)
+      if (this.Flatpickr && this.flatpickr) {
+        this.Flatpickr.localize(locale)
+        this.flatpickr.set('locale', locale)
 
-      this.flatpickr.set('altFormat', props.options.flatpickrOptions.altFormat)
-      this.flatpickr.setDate(props.value, false)
-      this.flatpickr.redraw()
+        this.flatpickr.set('altFormat', this.props.options.flatpickrOptions.altFormat)
```
MethodenMann: so if `this.props.options.flatpickrOptions.altFormat` changes this will not update since you wrapped everything in a "only when value or local changed"-if?

...why do you event have this "if"? is the performance not good without?

packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -135,7 +130,7 @@ class DateAbstract extends React.Component {
         />
         <StyledDateAbstractControl
           data-clear
-          hideButton={this.state.hideButton || this.props.immutable}
+          hideButton={!this.hasValue() && !this.props.immutable}
```
MethodenMann: shouldn't this be an OR (||)?

packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
khanlifts: Yes it won't redraw correctly without the condition.

packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -72,24 +71,18 @@ class DateAbstract extends React.Component {
...
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    const locale = this.getLocale(this.props.intl.locale)
 
     if (this.Flatpickr && this.flatpickr) {
       this.Flatpickr.localize(locale)
       this.flatpickr.set('locale', locale)
 
-      this.flatpickr.set('altFormat', props.options.flatpickrOptions.altFormat)
-      this.flatpickr.setDate(props.value, false)
+      this.flatpickr.set('altFormat', this.props.options.flatpickrOptions.altFormat)
```
MethodenMann: you removed functionality. Open the story and try editing the Value via knobs. The value will not change...

packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -72,24 +72,27 @@ class DateAbstract extends React.Component {
...
-    this.handleButtonVisibility()
-    const locale = this.getLocale(props.intl.locale)
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    const valueDiff = !_isEqual(prevProps.value, this.props.value)
+    const localeDiff = !_isEqual(prevProps.intl.locale, this.props.intl.locale)
+    const altFormatDiff = !_isEqual(
+      prevProps.options.flatpickrOptions.altFormat,
+      this.props.options.flatpickrOptions.altFormat
+    )
+    if (valueDiff || altFormatDiff || localeDiff) {
```
MethodenMann: i would split it up

`
   if (this.Flatpickr && this.flatpickr) {
 if ( !_isEqual(prevProps.value, this.props.value)) {
 this.flatpickr.setDate(this.props.value, false)
}
...
}
`

PR #729, opened 2019-06-12
Pr/preparation to reduce tocco theme

niklaushug:


packages/tocco-ui/src/SearchBox/SearchBox.js
```diff
@@ -42,24 +44,28 @@ class SearchBox extends React.Component {
...
-            onChange={this.onChange}
-          />
-          <StyledInputGroupBtn>
-            <Button
-              icon="search"
-              iconPosition="sole"
-              type="submit"
+      <StyledSearchBoxForm onSubmit={this.handleSubmit}>
+        <StatedValue
+          hasValue={!!this.state.inputValue && !!this.state.inputValue.length > 0}
```
MethodenMann: ` !!this.state.inputValue.length > 0 ` is not needed? Or what values do you handle with that

PR #730, opened 2019-06-12
refactor(tocco-ui): replace FormField with StatedValue

niklaushug:
- remove all className attributes from login
- extend StatedValue to omit signal auto detection


packages/login/src/components/LoginForm/LoginForm.js
```diff
@@ -31,44 +35,51 @@ export class LoginForm extends Component {
...
         <form onSubmit={this.handleSubmit.bind(this)}>
-          <div className="form-group dense">
-            <div className="input-group">
-              <span className="input-group-addon"><Icon icon="user"/></span>
-              <input
-                type="text"
-                className="form-control"
+
+          <StatedValue
+            hasValue={!!this.props.username && this.props.username.length > 0}
```
MethodenMann: same question here as in other PR (why length check)

packages/login/src/components/LoginForm/LoginForm.js
```diff
@@ -31,44 +35,51 @@ export class LoginForm extends Component {
...
-            <div className="input-group">
-              <span className="input-group-addon"><Icon icon="unlock"/></span>
-              <input
-                type="password"
-                className="form-control"
+            </StyledLoginFormInputWrapper>
+          </StatedValue>
+
+          <StatedValue
+            hasValue={!!this.props.password && this.props.password.length > 0}
```
MethodenMann: same here

packages/login/src/components/PasswordRequest/PasswordRequest.js
```diff
@@ -32,25 +36,29 @@ export class PasswordRequest extends React.Component {
 
   render() {
     return (
-      <StyledPasswordRequest className="login-form">
+      <StyledLoginFormWrapper>
```
MethodenMann: Why append "Wrapper" in this name? Usually you just call is Styled{Component}?

packages/login/src/components/passwordUpdate/PasswordUpdateDialog/PasswordInput.js
```diff
@@ -1,30 +1,36 @@
...
-        value={props.value}
-        onChange={e => props.onChange(e.target.value)}
-        readOnly={props.readOnly === true}
-        disabled={props.readOnly === true}
-        onKeyDown={props.onKeyDown}
-        autoFocus={props.autoFocus}
-      />
-    </div>
+    <StatedValue
+      hasValue={!!props.value && props.value.length > 0}
```
MethodenMann: Same question here

packages/login/src/components/LoginForm/LoginForm.js
niklaushug: Thanks for question it. Since "double bang operator" is used the length check is obsolete.

packages/login/src/components/PasswordRequest/PasswordRequest.js
niklaushug: The naming convention Styled{Component} does only work if each component has one styled component associated. Purpose of <StyledLoginFormWrapper> is to stretch and center <StyledLoginForm>. Therefore it is a wrapper. In means of clarity, versatility and reusability it makes sense to split styling into several chunks. May it make sense to use more generic names and to place the commonly used styled components alltogether and only in StyledLogin.js? Currently following files contain styled components.
./LoginForm/StyledLoginForm.js
./StyledLoginForm.js
./TwoStepLoginForm/StyledTwoStepLoginForm.js
./passwordUpdate/PasswordUpdateDialog/StyledPasswordInput.js
./passwordUpdate/PasswordUpdateDialog/StyledPasswordUpdateDialog.js

packages/login/src/components/LoginForm/LoginForm.js
niklaushug: see 3a5f4e4a59fc0c8a82c4475e57874c0b1bfc85c9

packages/login/src/components/LoginForm/LoginForm.js
```diff
@@ -31,44 +35,51 @@ export class LoginForm extends Component {
...
-              <span className="input-group-addon"><Icon icon="user"/></span>
-              <input
-                type="text"
-                className="form-control"
+
+          <StatedValue
+            hasValue={!!this.props.username}
+            id="login-username"
+            label={this.msg('client.login.form.userPlaceholder')}
+            mandatory
```
MethodenMann: mandatory is superfluous in Login context if you ask me. I would remove it

packages/tocco-ui/src/index.js
```diff
@@ -1,7 +1,11 @@
 export {default as Button, StyledButton} from './Button'
 export {default as ButtonGroup, StyledButtonGroup} from './ButtonGroup'
 export {default as ButtonLink, StyledButtonLink} from './ButtonLink'
-export {default as EditableValue} from './EditableValue'
+export {
+  default as EditableValue,
+  StyledEditableWrapperCss,
+  StyledInputCss
+} from './EditableValue'
```
MethodenMann: Tocco Login styling still works with this changes?

packages/login/src/components/PasswordRequest/PasswordRequest.js
MethodenMann: The sole purpose of `StyledLoginFormWrapper` is to center (and stretch)? IMO it would be a lot clearer if this component is called something with center/stretch.

packages/tocco-ui/src/index.js
niklaushug: yes

packages/login/src/components/PasswordRequest/PasswordRequest.js
niklaushug: Stretch is only one of several purposes. A concise name is not available. I established a generic yet a homogen naming for fields.

packages/login/src/components/passwordUpdate/PasswordUpdateDialog/PasswordInput.js
```diff
@@ -1,30 +1,36 @@
...
+    <StatedValue
+      hasValue={!!props.value}
+      id={`${props.name}-input`}
+      immutable={props.readOnly === true}
+      label={props.label}
+      signal={getSignal}
+    >
+      <StyledPasswordInputWrapper immutable={props.readOnly === true}>
+        <StyledPasswordInput
+          id={`${props.name}-input`}
```
rzueger: id attributes used this way are prone to lead to conflicts and, thus, to problems.

can we, somehow, get rid of them?

- are they really necessary? isn't it possible to change the markup in StatedValue so that the input component is nested in the label component?
- if they are necessary, could we choose a slightly different approach where StatedValue generates a unique id which it passes down to the input component?

niklaushug: I totally agree. From React perspective there is no need for an id attribute. Probably a jQuery plugin or Tocco Framework depend on id's. Before investing it I'd like to ask @MethodenMann if he remember why id's were implemented. There are some other places in tocco-client where id's are used.

PR #733, opened 2019-06-14
Pr/async cypress tests

khanlifts:


cypress/integration/entitybrowser_spec.js
```diff
@@ -11,62 +11,80 @@ describe('Entity Browser', function() {
...
   describe('SearchForm', () => {
     it('should display extended search form', function() {
-      cy.get('[data-cy=extend-search-button]').children()
-        .click()
-      cy.get('[data-cy=cypress-form-field]')
-        .should('have.length.above', 1)
+      cy.get('[data-cy=extend-search-button]').children().first()
+        .click({force: true})
+      cy.get('[data-cy=cypress-form-field]').then(el => {
+        cy.wrap(el).should('have.length.above', 10)
```
MethodenMann: should this test not verify that before the click there were less than 10 fields

cypress/integration/entitybrowser_spec.js
```diff
@@ -11,62 +11,80 @@ describe('Entity Browser', function() {
...
       cy.get('[data-cy=cypress-form-field]').children().children()
         .type('Test Person')
-      cy.get('[data-cy=search-form_reset-button]').next().children().first()
-        .click()
-      cy.get('[data-list-cell]').should('have.length.above', 1)
+      cy.get('[data-cy=search-form_reset-button]')
+        .click({force: true})
+      cy.get('[data-cy=list-cell]').then(el => {
+        const listCells = el.children()
+        cy.wrap(listCells).should('have.length.above', 10)
```
MethodenMann: does this really verify what you want to test?
Shouldn't you check if "Test PersoN" isnt in the input field after clicking result rather than check if there are 10 record?

cypress/integration/entitybrowser_spec.js
```diff
@@ -11,62 +11,80 @@ describe('Entity Browser', function() {
...
     })
 
     it('should change page back to list view without warning', function() {
-      cy.get('[data-cy=entity-detail_back-button]').children().first().children().first()
-        .click()
-      cy.get('[data-list-cell]').should('have.length.above', 1)
+      cy.get('[data-cy=entity-detail_back-button]').children().first().then(el => {
+        cy.wrap(el).click()
+      })
+      cy.get('[data-cy=list-cell]').then(el => {
```
MethodenMann: would it be better to make a selector on the list view rather than check if after the click on "BACK" you see some records?

cypress/integration/entitybrowser_spec.js
```diff
@@ -11,62 +11,80 @@ describe('Entity Browser', function() {
 
   describe('ListView', () => {
     it('should load and display essential parts of the list view', function() {
+      cy.contains('Daten werden geladen...', {timeout: 15000})
```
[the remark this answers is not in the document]
khanlifts: The failing test indicated so.

cypress/integration/entitybrowser_spec.js
khanlifts: I added this.

cypress/integration/entitybrowser_spec.js
```diff
@@ -11,62 +11,78 @@ describe('Entity Browser', function() {
...
+        cy.wrap(el).should('have.length.above', 10)
+      })
     })
   })
 
   describe('DetailView', () => {
     beforeEach(() => {
-      cy.get('[data-list-cell]').first()
-        .click()
+      cy.get('[data-cy=list-cell]').then(el => {
```
rzueger: isn't it possible to select the first element directly and click it? (`cy.get('[data-cy=list-cell]:first').click()` or similar)

cypress/integration/entitybrowser_spec.js
```diff
@@ -11,62 +11,78 @@ describe('Entity Browser', function() {
...
+      })
+      cy.contains('Aktionen auf Alle')
     })
 
     it('should display warning on changing page back to list view on edited form', function() {
       cy.get('input#input-detailForm-callname')
         .type('{selectall}{del}')
-      cy.get('[data-cy=entity-detail_back-button]').children().first().children().first()
-        .click()
+      cy.get('[data-cy=entity-detail_back-button]').children().first().then(el => {
```
rzueger: see comment on line 27

cypress/integration/entitybrowser_spec.js
```diff
@@ -11,62 +11,78 @@ describe('Entity Browser', function() {
...
     })
   })
 
   describe('SearchForm', () => {
     it('should display extended search form', function() {
-      cy.get('[data-cy=extend-search-button]').children()
-        .click()
-      cy.get('[data-cy=cypress-form-field]')
-        .should('have.length.above', 1)
+      cy.get('[data-cy=cypress-form-field]').then(el => {
```
rzueger: why not `cy.get('[data-cy=cypress-form-field]').should('have.length', 1)`?

cypress/integration/login_spec.js
```diff
@@ -0,0 +1,77 @@
...
+  })
+
+  describe('Login', () => {
+    it('should load and display login view', function() {
+      cy.contains('Dieser Bereich ist privat.', {timeout: 15000})
+      cy.get('button').should('have.length', 2)
+    })
+
+    it('should not be possible to click login button', function() {
+      cy.get('[data-cy=login-form_user-input]').then(el => {
```
rzueger: Why do you use `then()` everywhere? You should be able to append `...type(...)` and all the other commands directly after `get()`?

cypress/integration/entitybrowser_spec.js
khanlifts: Yes. But then it won't wait in an async way.

cypress/integration/login_spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+    it('should request password from request page', function() {
+      cy.contains('Passwort vergessen?')
+      cy.get('[data-cy=login-form_request-button]')
+        .click()
+      cy.get('[data-cy=password-request_input]')
+        .type('{selectall}{del}test@testmail.com')
+      cy.get('[data-cy=password-request_submit-button]')
+        .click()
+      cy.contains('Dieser Bereich ist privat.')
```
MethodenMann: why check for "Dieser Bereich ist privat" ...has nothing to do with password request?

cypress/integration/login_spec.js
```diff
@@ -0,0 +1,64 @@
...
+    })
+
+    it('should login', function() {
+      cy.get('[data-cy=login-form_user-input]')
+        .type('{selectall}{del}cypress_test')
+      cy.get('[data-cy=login-form_password-input]')
+        .type('{selectall}{del}Test_pw1')
+      cy.get('[data-cy=login-form_login-button]')
+        .click()
+        .children().first().should('be', 'svg', {timeout: 5000})
```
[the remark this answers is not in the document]
khanlifts: We already spoke about this. You have to do this because I couldn't do it.

cypress/integration/entitybrowser_spec.js
```diff
@@ -5,28 +5,49 @@ describe('Entity Browser', function() {
...
+          .should('be.empty')
+      })
     })
   })
 
   describe('DetailView', () => {
     beforeEach(() => {
-      cy.get('[data-list-cell]').first()
-        .click()
+      cy.visit(`${helpers.getStoryUrl(['Apps', 'Entity Browser'], 'Entity Browser')}detail/13477`)
```
MethodenMann: 13477? Very random.
Good: Create a new user before every test run and test with this one. This will make sure this record really exists.
ok for the moment: Create a record by hand and make clear everyone sees that this is a record for testing and should not be deleted.

cypress/integration/entitybrowser_spec.js
```diff
@@ -5,28 +5,49 @@ describe('Entity Browser', function() {
...
     })
 
     it('should display detail view', function() {
-      cy.get('[data-cy=cypress-form-field]')
-        .should('have.length.above', 1)
+      cy.get('[data-cy=cypress-form-field]').should('have.length.above', 1)
```
MethodenMann: (this will also be true if you are on the list view)

MethodenMann: please rename to data=cy=form-field to be consistent

cypress/integration/entitybrowser_spec.js
```diff
@@ -5,28 +5,49 @@ describe('Entity Browser', function() {
...
+    })
+
+    describe('SearchForm', () => {
+      it('should display extended search form', function() {
+        cy.get('[data-cy=form-field]')
+          .should('have.length', 1)
+        cy.get('[data-cy=extend-search-button]')
+          .click()
+        cy.get('[data-cy=form-field]')
+          .should('have.length.above', 10)
```
rzueger: I would choose a slightly different approach when it comes to the selection of assertion criteria that make sense.

For instance, `10` is a magic arbitrary number. We actually don't care if there are more than 10 fields or if there are only 9. The only thing we really care about is that there is more than 1 field. So, it would be perfectly fine to simply assert that there is more than 1. This would also make the test more stable and leave less room for speculation (someone who stumbles over this test for some reason cannot know if that number `10` actually is important or not).

cypress/integration/login_spec.js
```diff
@@ -0,0 +1,59 @@
...
+      cy.contains('Passwort wurde per E-Mail versendet')
+    })
+
+    it('should change to login page on abort of password request ', function() {
+      cy.get('[data-cy=login-form_request-button]')
+        .click()
+      cy.contains('Passwort vergessen?')
+      cy.get('[data-cy=password-request_abort-button]')
+        .click()
+      cy.contains('Dieser Bereich ist privat.')
```
rzueger: Isn't there a more reliable assertion? Can't you, for example, expect the username and password field to be visible? (selected by `data-cy` attribute)

cypress/integration/login_spec.js
```diff
@@ -0,0 +1,60 @@
...
+    })
+
+    it('should change to login page on abort of password request ', function() {
+      cy.get('[data-cy=login-form_request-button]')
+        .click()
+      cy.contains('Passwort vergessen?')
+      cy.get('[data-cy=password-request_abort-button]')
+        .click()
+      cy.get('[data-cy=login-form_user-input]')
+      cy.get('[data-cy=login-form_password-input]')
```
rzueger: maybe also add `should('be.visible')` (I reckon Cypress is able to select elements even if they're hidden)

PR #742, opened 2019-06-26
Pr/replace old api

khanlifts:


packages/tocco-util/src/route/asyncRoute.js
```diff
@@ -3,31 +3,21 @@ import React from 'react'
...
     static Component = null
-    mounted = false
 
     state = {
       Component: AsyncComponent.Component
     }
 
-    componentWillMount() {
+    componentDidMount() {
+      this.mounted = true
```
rzueger: what's that `mounted` variable still needed for?

khanlifts: rebase mistake. It isn't.

khanlifts: If the component was already mounted and I change the page back from another page it will wait to call setState until the component is mounted.

packages/entity-list/src/main.js
```diff
@@ -105,12 +106,14 @@ class EntityListApp extends React.Component {
...
-      this.app.store.dispatch(action)
-    })
+  static getDerivedStateFromProps = props => {
+    const changedProps = _pickBy(props, (value, key) => !_isEqual(value, props[key]))
+    if (!_isEmpty(changedProps)) {
+      getDispatchActions(changedProps).forEach(action => {
+        this.app.store.dispatch(action)
+      })
+    }
+    return null
```
rzueger: this method actually never returns a result other than `null`. by intention?

I'm not familiar with using `getDerivedStateFromProps` but, as I understand it, this is not the use case this lifecycle method is meant for.

documentation (https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops):
"If you need to perform a side effect (for example, data fetching or an animation) in response to a change in props, use componentDidUpdate lifecycle instead."

Why did you choose `getDerivedStateFromProps` over `componentDidUpdate` in this case?

khanlifts: I chose it because it gets called before render like `componentWillReceiveProps` to keep the timing of updates as close to the original method as possible. 
Return of at least some state change (in this case `null`) is required.

packages/tocco-ui/src/Upload/Upload.js
```diff
@@ -28,8 +28,8 @@ export class Upload extends React.Component {
     this.props.onUpload(file)
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (!_isEqual(nextProps.value, this.props.value)) {
+  componentDidUpdate(prevProps, prevState, snapshot) {
```
MethodenMann: unused param snapshot

packages/tocco-ui/src/LoadMask/LoadMask.js
```diff
@@ -45,9 +44,10 @@ class LoadMask extends React.Component {
     this.mounted = false
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentDidUpdate(prevProps, prevState, snapshot) {
```
MethodenMann: unused param snapshot

packages/tocco-ui/src/LoadMask/LoadMask.js
```diff
@@ -45,9 +44,10 @@ class LoadMask extends React.Component {
...
     if (!this.state.initialized) {
-      if (nextProps.required && this.requiredLoaded(nextProps.required)) {
+      const test = this.requiredLoaded(this.props.required)
```
MethodenMann: test?

packages/simple-form/src/components/Form/Form.js
```diff
@@ -8,7 +8,7 @@ import {intlShape} from 'react-intl'
 const REDUX_FORM_NAME = 'simpleForm'
 
 class Form extends React.Component {
-  componentWillMount() {
+  componentDidMount() {
```
MethodenMann: maybe the constructor would be a better fit?

packages/scheduler/src/components/FullCalendar/FullCalendar.js
```diff
@@ -114,16 +114,19 @@ class FullCalendar extends React.Component {
     this.forceUpdate()
   }
 
-  componentWillReceiveProps(newProps) {
-    this.calendarElement.fullCalendar('option', this.getFullCalendarOptions(newProps))
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    if (!_isEqual(prevProps, this.props)) {
```
MethodenMann: does it really need a full compare of props with prevProps? getFullCalendarOptions does not use the whole prop object. Maybe you could write a more specific if

packages/scheduler/src/components/FullCalendar/FullCalendar.js
```diff
@@ -114,16 +114,19 @@ class FullCalendar extends React.Component {
...
+        this.calendarElement.fullCalendar('removeEventSources')
+        this.calendarElement.fullCalendar('addEventSource', this.props.events)
+      }
 
-    if (!_isEqual(newProps.resources, this.props.resources)) {
-      this.calendarElement.fullCalendar('refetchResources')
+      if (!_isEqual(prevProps.resources, this.props.resources)) {
+        this.calendarElement.fullCalendar('refetchResources')
+      }
+      this.forceUpdate()
```
MethodenMann: forceUpdate 🤨 why?

packages/login/src/components/passwordUpdate/PasswordUpdateDialog/PasswordUpdateDialog.js
```diff
@@ -9,7 +9,7 @@ import ValidationRules from '../ValidationRules'
 import FailureMessage from '../FailureMessage'
 
 class PasswordUpdateDialog extends Component {
-  componentWillMount() {
+  componentDidMount() {
```
MethodenMann: Maybe constructor is a better fit? (generel comment for all simular changes)

packages/entity-list/src/main.js
```diff
@@ -105,12 +105,13 @@ class EntityListApp extends React.Component {
...
   }
 
-  componentWillReceiveProps = nextProps => {
-    const changedProps = _pickBy(nextProps, (value, key) => !_isEqual(value, this.props[key]))
-
-    getDispatchActions(changedProps).forEach(action => {
-      this.app.store.dispatch(action)
-    })
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    if (!_isEqual(prevProps, this.props)) {
```
MethodenMann: `  if (!_isEqual(prevProps, this.props)) `
Needed? On the next line you allready check for changes and please keep the object-dif helper method as shown to you in mind

packages/scheduler/src/components/FullCalendar/FullCalendar.js
khanlifts: The component has to update one more time after the new props got set. This is the simpler way to setState, which would have included setting calendarElement in all lines to the state.

PR #747, opened 2019-07-08
Pr/remove lazy component

khanlifts:


packages/tocco-ui/src/EditableValue/typeEditors/HtmlEdit.js
```diff
@@ -42,15 +41,17 @@ class HtmlEdit extends React.Component {
...
-        <this.lazyQuill
-          name={this.props.name}
-          onChange={this.handleChange}
-          id={this.props.id}
-          theme="snow"
-          value={this.props.value}
-          readOnly={this.props.immutable}
-          modules={{toolbar: !this.props.immutable}}
-        />
+        <Suspense fallback={<i/>}>
```
rzueger: are there already other usages of `Suspense`? what fallback is used there?

khanlifts: This is the first usage of `Suspense`.
I used `<i/>` as fallback because we already used it for Icon.js. See Line 46 (red) in Icon.js below.

PR #748, opened 2019-07-09
Pr/refactor tocco theme

niklaushug:


packages/entity-list/src/components/PaginationPanel/PaginationPanel.js
```diff
@@ -0,0 +1,65 @@
...
+  const {
+    changePage,
+    components,
+    totalPages,
+    currPage
+  } = props
+
+  return (
+    <StyledPaginationPanel>
+      {components.totalText}
```
rzueger: `components` is not declared as a required prop and thus accessing `totalText`could lead to a null pointer exception.

the same applies to the other props.

packages/tocco-ui/src/ButtonGroup/ButtonGroup.js
```diff
@@ -12,7 +12,8 @@ const ButtonGroup = props => {
     <StyledButtonGroup melt={props.melt}>
       { // eslint-disable-next-line
         React.Children.map(props.children, child => {
-          return React.cloneElement(child, {look: props.look, buttonGroupInk: props.ink, buttonGroupMelt: props.melt})
+          return child !== null
```
rzueger: I personally wouldn't use type safe null checks in cases like this. The code becomes more stable if you remove `!== null`.

PR #749, opened 2019-07-11
feat(tocco-ui): establish PanelGroup

niklaushug:
- group panels to ensure that only one is open at the same time
- define a panel which is initially opened

Refs: WEB-120

### DIFF
--- a/packages/tocco-ui/src/Panel/Panel.js
+++ b/packages/tocco-ui/src/Panel/Panel.js
@@ -14,9 +14,18 @@ class Panel extends React.PureComponent {
 
   toggleOpenState = () => {
     if (this.props.isToggleable) {
+      if (typeof this.props.toggleOpenStateAsGroup === 'function') {
+        this.props.toggleOpenStateAsGroup(this.state.isOpen ? -1 : this.props.panelId)
+      }
       this.setState(prevState => ({
         isOpen: !prevState.isOpen
       }))
+    }
+  }
+
+  componentDidUpdate() {
+    if (typeof this.props.toggleOpenStateAsGroup === 'function') {
+      this.setState({isOpen: this.props.isOpen})
     }
   }
 
@@ -65,7 +74,15 @@ Panel.propTypes = {
   /**
    * Boolean to control if body can be opened or closed. Default value is 'true'.
    */
-  isToggleable: PropTypes.bool
+  isToggleable: PropTypes.bool,
+  /**
+   * Function gets passed from PanelGroup to ensure that only one Panel is open at one time.
+   */
+  toggleOpenStateAsGroup: PropTypes.func,
+  /**
+   * Id is passed from PanelGroup to identify each Panel in a group.
+   */
+  panelId: PropTypes.number
 }
 
 export default Panel
--- a/packages/tocco-ui/src/Panel/Panel.js
+++ b/packages/tocco-ui/src/Panel/Panel.js
@@ -14,8 +14,8 @@ class Panel extends React.PureComponent {
 
   toggleOpenState = () => {
     if (this.props.isToggleable) {
-      if (typeof this.props.toggleOpenStateAsGroup === 'function') {
-        this.props.toggleOpenStateAsGroup(this.state.isOpen ? -1 : this.props.panelId)
+      if (typeof this.props.onToggle === 'function') {
+        this.props.onToggle(!this.state.isOpen)
       }
       this.setState(prevState => ({
         isOpen: !prevState.isOpen
@@ -24,7 +24,7 @@ class Panel extends React.PureComponent {
   }
 
   componentDidUpdate() {
-    if (typeof this.props.toggleOpenStateAsGroup === 'function') {
+    if (typeof this.props.onToggle === 'function') {
       this.setState({isOpen: this.props.isOpen})
     }
   }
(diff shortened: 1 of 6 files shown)


packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -65,7 +74,15 @@ Panel.propTypes = {
   /**
    * Boolean to control if body can be opened or closed. Default value is 'true'.
    */
-  isToggleable: PropTypes.bool
+  isToggleable: PropTypes.bool,
+  /**
+   * Function gets passed from PanelGroup to ensure that only one Panel is open at one time.
+   */
+  toggleOpenStateAsGroup: PropTypes.func,
```
rzueger: the Panel itself actually shouldn't have to care about a surrounding group. Therefore, I'd simply rename the handler to `onToggle` and pass `true` to the event handler, if the panel is open or `false`, if closed.

`panelId` is not needed either.

Also see other comment in PanelGroup.js.

packages/tocco-ui/src/Panel/PanelGroup.js
```diff
@@ -0,0 +1,40 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import StyledPanelGroup from './StyledPanelGroup'
+
+class PanelGroup extends React.Component {
+  state = {
+    openPanelId: this.props.openPanelId >= 0 ? this.props.openPanelId : -1
```
rzueger: I'd personally prefer `undefined` or `null` over `-1`. IMO `-1` is more of a workaround for non nullable variables or types.

packages/tocco-ui/src/Panel/PanelGroup.js
```diff
@@ -0,0 +1,40 @@
...
+
+  render() {
+    return (
+      <StyledPanelGroup>
+        {
+          React.Children.map(this.props.children, (child, i) =>
+            React.cloneElement(child, {
+              isOpen: this.state.openPanelId === i,
+              panelId: i,
+              toggleOpenStateAsGroup: this.toggleOpenStateAsGroup
```
rzueger: Actually, the child doesn't have to receive a `panelId`, you can simply use the index without passing it down to the child.

what about doing it like this:

```
handleToggle = (index, open) => {
  this.setState({
    openPanelIndex: open === true ? index : undefined
  })
}

render () {
  return (
    <StyledPanelGroup>
      {
        React.Children.map(this.props.children, (child, i) =>
          React.cloneElement(child, {
            isOpen: this.state.openPanelIndex === i,
            onToggle: this.handleToggle.bind(this, i)
          })
        )
     }
    </StyledPanelGroup>
  )
}

File: packages/tocco-ui/src/Panel/Panel.js [hunk 4]
```diff
@@ -65,7 +74,11 @@ Panel.propTypes = {
   /**
    * Boolean to control if body can be opened or closed. Default value is 'true'.
    */
-  isToggleable: PropTypes.bool
+  isToggleable: PropTypes.bool,
+  /**
+   * Function gets passed from PanelGroup to ensure that only one Panel is open at one time.
```
rzueger: documentation: The Panel doesn't care if "only one Panel is open at one time". onToggle is simply a function that is called whenever the Panel is closed or opened.

packages/tocco-ui/src/Panel/PanelGroup.js
```diff
@@ -0,0 +1,41 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import StyledPanelGroup from './StyledPanelGroup'
+
+class PanelGroup extends React.Component {
+  state = {
+    openPanelId: this.props.openPanelId
```
rzueger: I'd rename `openPanelId` to `openPanelIndex` (both the state property and the prop; makes it easier to understand quickly what it represents -> `...Index` indicates that it's an integer starting at zero whereas `...Id` could be anything)

packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -14,12 +14,21 @@ class Panel extends React.PureComponent {
...
       this.setState(prevState => ({
         isOpen: !prevState.isOpen
       }))
     }
   }
 
+  componentDidUpdate() {
+    if (typeof this.props.onToggle === 'function') {
+      this.setState({isOpen: this.props.isOpen})
+    }
```
rzueger: - the local state should be updated with the `isOpen` prop, no matter if there's an `onToggle` callback or not
- I'm not sure if `componentDidUpdate` is the right lifecycle method to use. Reading the documentation, I think `getDerivedStateFromProps` could be the better choice (https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops).
- If using `componentDidUpdate` is the way to go, I think `setState` should only be called if the prop is different from the current state (for performance reasons)

niklaushug: getDerivedStateFromProps seems to be the better choice because getDerivedStateFromProps has no sideeffects (see http://projects.wojtekmaj.pl/react-lifecycle-methods-diagram/) and is executed before render. Hence Panel is rendered one time less. Also the constructor is not needed because getDerivedStateFromProps is executed while mounting and updating.

rzueger: OK, looks better now.

One more thing: IMO, a function that is supposed to return something (whether an actual value, `null` or `undefined`), should return a return value explicitly in all cases. Here, if the condition on line 23 doesn't evaluate to true, `undefined` is returned implicitly (nobody knows whether that's intended or you just forgot the return value). So, I suggest to add `return null` on line 26 so that it's clear immediately.

packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -65,7 +70,11 @@ Panel.propTypes = {
   /**
    * Boolean to control if body can be opened or closed. Default value is 'true'.
    */
-  isToggleable: PropTypes.bool
+  isToggleable: PropTypes.bool,
+  /**
+   * Function is triggered by componentDidUpdate or by toggleOpenState.
```
rzueger: I still wouldn't understand what this sentence means :) No one who uses this component cares how `componentDidUpdate` is implemented, and what is toggleOpenState? Why not simply write "Function that is called when Panel is opened or closed"?

PR #753, opened 2019-07-16
Pr/replace panel in resource scheduler

niklaushug:
getDerivedStateFromProps is only a good choice if Panel is wrapped in PanelGroup. If Panel has not such a wrapper getDerivedStateFromProps makes it impossible to overwrite state.isOpen dynamically from inside the component. Hence open state is not toggleable. Note that isOpen is used as initial value and to control panels by PanelGroup.


packages/resource-scheduler/src/components/SearchPanel/SearchPanel.js
```diff
@@ -1,80 +1,63 @@
...
-  }
-
   handleSelect = name => selection => {
     this.props.updateRequestedCalendars(name, selection)
   }
 
-  getSearchLists = calendarTypes => {
-    const accordionHeader = (label, name, color) =>
-      <div onClick={() => { this.handleAccordionSelect(name) }}>
+  accordionHeader = (label, color) =>
```
MethodenMann: make AccordionHeader a stateless react component rather than a function that returns jsx

packages/resource-scheduler/src/components/SearchPanel/SearchPanel.js
```diff
@@ -1,80 +1,63 @@
...
           style={{...(color ? {color} : {})}}
           position={design.position.PREPEND}
-        />
-        <Typography.H5>{label}</Typography.H5>
-        <Icon
-          icon={`chevron-${this.state.activeKey === name ? 'up' : 'down'}`}
-          style={{float: 'right'}}
-        />
-      </div>
+        />  {label}
```
MethodenMann: maybe label on a separate line?

packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -9,7 +9,7 @@ import StyledPanel from './StyledPanel'
  */
 class Panel extends React.PureComponent {
   state = {
-    isOpen: undefined
+    isOpen: this.props.isOpen
```
MethodenMann: it is an antipattern to copy props to the state.

packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -23,8 +23,10 @@ class Panel extends React.PureComponent {
     }
   }
 
-  static getDerivedStateFromProps(props, state) {
-    return (props.isOpen !== state.isOpen) ? {isOpen: props.isOpen} : null
+  componentDidUpdate() {
+    if (typeof this.props.onToggle === 'function') {
```
MethodenMann: are you sure about this if statement? so after any update you check whether onToogle is a function and if so you change the state (and trigger another update?)
if you call setState in componentDidUpdate you should compare the relevant props otherwise risking an endless loop.

please read https://reactjs.org/docs/react-component.html#componentdidupdate

packages/tocco-ui/src/Panel/Panel.js
```diff
@@ -74,7 +76,7 @@ Panel.propTypes = {
    */
   isToggleable: PropTypes.bool,
   /**
-   * Function that is called when Panel is opened or closed.
+   * Function is triggered by componentDidUpdate or by toggleOpenState.
```
MethodenMann: internal information like componentDidUpdate is not relevant for a documentation.

packages/resource-scheduler/src/components/SearchPanel/SearchPanel.js
niklaushug: A separate line was my first intention but I want preserve the spaces between icon and text.

PR #757, opened 2019-07-18
feat(entity-detail): add defaultValue App param

MethodenMann:
- this for example allows to pre set a relation if the create view is
  opened from a relation list.


packages/entity-detail/src/modules/entityDetail/sagas.js
```diff
@@ -76,16 +76,17 @@ export function* loadEntityModel(entityName) {
 }
 
 export function* loadDetailView() {
-  const {entityName, formName, mode} = yield select(entityDetailSelector)
+  const {entityName, formName, mode, defaultValues} = yield select(entityDetailSelector)
```
khanlifts: We could name these `storeDefaultValues` and `definitionDefaultValues` to avoid numbers in naming?

packages/entity-detail/README.md
```diff
@@ -13,6 +13,8 @@ React-registry name: `entity-detail`
 | `formName`             |           | Detail-form that should be loaded
 | `mode`                 |           | Determines if the detail is in `create` or `update` mode
 | `showSubGridsCreateButton`|        | (Temporary) Flag to show/hide a create button on any subgrid
+| `defaultValues`        |           | Object with fieldName as Key and value as value. e.g. {lastname: 'Simpson'}
+
```
khanlifts: **k**ey

packages/entity-detail/src/modules/entityDetail/sagas.js
```diff
@@ -76,16 +76,17 @@ export function* loadEntityModel(entityName) {
...
 
   yield call(loadEntityModel, entityName)
   const formDefinition = yield call(loadDetailFormDefinition, formName)
 
   if (mode === modes.CREATE) {
     yield put(actions.setEntity({paths: {}, model: entityName}))
     const fieldDefinitions = yield call(form.getFieldDefinitions, formDefinition)
-    const defaultValues = yield call(form.getDefaultValues, fieldDefinitions)
-    yield put(formActions.initialize(FORM_ID, defaultValues))
+    const fromDefaultValues = yield call(form.getDefaultValues, fieldDefinitions)
```
khanlifts: Why not `formDefaultValues`? I thought _from_ is bad practice.

PR #758, opened 2019-07-19
feat(entity-detail): add reduceInitialValue method

khanlifts:
- reduceInitialValue makes sure there are only the relevant
  attributes in formValues so redux-form's dirty selector can compare
  actual changes

Refs: TOCDEV-468

### DIFF
--- a/packages/app-extensions/src/form/reduxForm.js
+++ b/packages/app-extensions/src/form/reduxForm.js
@@ -1,5 +1,7 @@
 import _forOwn from 'lodash/forOwn'
 import _isEqual from 'lodash/isEqual'
+import _pick from 'lodash/pick'
+import _isEmpty from 'lodash/isEmpty'
 
 import {generalErrorField, entityValidatorErrorsField, relatedEntityErrorsField} from './formErrors'
 
@@ -62,19 +64,27 @@ export const transformFieldNameBack = fieldName => (fieldName.replace(/--/g, '.'
 
 export const entityToFormValues = entity => {
   if (!entity || !entity.paths) return {}
-  const result = {}
-  const paths = entity.paths
+
+  const result = {
+    [versionField]: entity.version
+  }
 
   Object.keys(entity.paths).forEach(key => {
-    const field = paths[key]
-    if (field.type === 'entity' || field.type === 'entity-list' || field.type === 'display-expression') {
-      result[transformFieldName(key)] = field.value
+    const field = entity.paths[key]
+    const transformedFieldName = transformFieldName(key)
+    const typeValueTransformer = typeValueTransformers[field.type]
+
+    const value = field.value
+
+    if (!value || _isEmpty(value)) {
+      result[transformedFieldName] = null
     } else {
-      result[transformFieldName(key)] = field.value ? field.value.value : null
+      result[transformedFieldName] = typeValueTransformer
+        ? typeValueTransformer(value)
+        : value.value || null
     }
   })
 
-  result[versionField] = entity.version
   return result
 }
 
@@ -88,4 +98,11 @@ export const getDirtyFields = (initialValues, values, isCreate) => {
   })
 
   return dirtyFields
+}
+
+const relevantRelationAttributes = ['key', 'display']
+const typeValueTransformers = {
+  'display-expression': value => value,
+  'entity': value => _pick(value, relevantRelationAttributes),
+  'entity-list': value => value.map(childValue => _pick(childValue, relevantRelationAttributes))
 }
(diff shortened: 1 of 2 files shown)


packages/entity-detail/src/util/detailView/reduceInitialValues.spec.js
```diff
@@ -0,0 +1,52 @@
...
+      key: '3',
+      model: 'Event_connection_type',
+      multiRelations: [],
+      paths: null,
+      singleRelations: [],
+      version: 1
+    }
+  ]
+}
+const result = {
```
MethodenMann: more like `expectedResult`

packages/entity-detail/src/modules/entityDetail/sagas.js
```diff
@@ -188,7 +189,9 @@ export function* loadData() {
   const {entityName, entityId, formName, formDefinition} = yield select(entityDetailSelector)
   const entity = yield call(loadEntity, entityName, entityId, formDefinition, formName)
   const formValues = yield call(form.entityToFormValues, entity)
-  yield put(formActions.initialize(FORM_ID, formValues, {keepDirty: true}))
+  const model = yield call(fetchModel, entityName)
```
MethodenMann: wouldn't `entityToFormValues` be a good place enhance rather than a new separate function?

packages/entity-detail/src/util/detailView/reduceInitialValues.spec.js
```diff
@@ -0,0 +1,85 @@
+import reduceInitialValues, {getRelevantAttributes} from './reduceInitialValues'
+
+const formValues = {
```
MethodenMann: a good tests should also contain data that is not be affected by your changes. e.g. a string field

packages/entity-detail/src/util/detailView/reduceInitialValues.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+  return relevantAttributes[modelType] ? relevantAttributes[modelType] : null
+}
+
+export default (initialValues, model) => {
+  const values = {...initialValues}
+
+  for (const key in values) {
+    const fieldModel = model[key]
+    if (values[key] && fieldModel && getRelevantAttributes(fieldModel.type)) {
```
MethodenMann: i would assign `values[key]` to a var. makes to code more readable

packages/entity-detail/src/util/detailView/reduceInitialValues.js
```diff
@@ -0,0 +1,30 @@
+import _pick from 'lodash/pick'
+import _isObject from 'lodash/isObject'
+
+export const getRelevantAttributes = modelType => {
```
MethodenMann: that does not really make sense since almost no other type is a object.
i thought more about something like this:

```
const typeTransformers = {
retlation: (value, model)=> {
 //do stuff
return transormedValue
},
string: (value, model) => {
return value + "!!" //example that appends !! to all strings
}

}


//and then:

typeTransformers[fieldModel.type](value, fieldModel)
```

khanlifts: Could you clarify your comment? `modelType` is always a string in my solution, not an object.

packages/entity-detail/src/modules/entityDetail/sagas.js
khanlifts: Functions should have rather one than multiple purposes. So no.

packages/app-extensions/src/form/reduxForm.js
```diff
@@ -62,19 +63,22 @@ export const transformFieldNameBack = fieldName => (fieldName.replace(/--/g, '.'
...
 
   Object.keys(entity.paths).forEach(key => {
-    const field = paths[key]
-    if (field.type === 'entity' || field.type === 'entity-list' || field.type === 'display-expression') {
-      result[transformFieldName(key)] = field.value
-    } else {
-      result[transformFieldName(key)] = field.value ? field.value.value : null
-    }
+    const field = entity.paths[key]
+    const type = field.type
```
rzueger: (there's only one usage of the `type` variable and it's only barely shorter than `field.type` -> I personally would inline the variable on line 75)

packages/app-extensions/src/form/reduxForm.js
```diff
@@ -62,19 +63,22 @@ export const transformFieldNameBack = fieldName => (fieldName.replace(/--/g, '.'
...
+    const transformedFieldName = transformFieldName(key)
+    const typeValueTransformer = typeValueTransformers[type]
+
+    result[transformedFieldName] = typeValueTransformer
+      ? typeValueTransformer(field.value)
+      : (field.value.value || null)
```
rzueger: there was a null check for `field.value` before (`field.value ? field.value.value : null`). not necessary?

PR #767, opened 2019-08-09
feat(sso-login): add automatic initialisation of login windows

aschi:
add possibility to pass an optional "autoLogin" parameter. if this parameter
is set, the login process is automatically started as soon as the providers
are loaded.


MethodenMann:
readme documentation of new input parameter is missing

PR #786, opened 2019-09-19
Pr/performac

MethodenMann:


packages/tocco-util/src/intl/intl.js
```diff
@@ -1,20 +1,24 @@
...
+import cache from '../cache'
 
-  return loadLocale.then(locale => setLocale(store, modules, locale))
+export const initIntl = async(store, modules, forcedLocale) => {
+  let locale = forcedLocale
+  if (!locale) {
+    locale = await getUserLocale()
+  }
+  await setLocale(store, modules, locale)
+  return true
```
rzueger: what's that hardcoded return value for?

packages/tocco-util/src/intl/intl.js
```diff
@@ -24,16 +28,57 @@ const fetchOptions = {
...
+  cache.add('user', 'locale', locale)
+  return locale
+}
+
+export const loadTextResources = async(locale, modules) => {
+  let result = {}
+  const notLoadedModules = []
+
+  modules.forEach(module => {
+    const cachedModuleResource = cache.get('textResource', module)
```
rzueger: what happens when the user changes his locale? the cache contains only text resources for locale, right? is the cache invalidated somewhere when the locale changes?

MethodenMann: not yet. Could be done in the future. For now the user has do close the tab and reopen the page.

PR #791, opened 2019-09-30
Pr/cypress2

MethodenMann:


cypress/integration/entitybrowser_spec.js
```diff
@@ -1,59 +1,88 @@
 import helpers from '../helpers/helpers'
 
+const getPkFromUrl = url => url.substring(url.lastIndexOf('/') + 1)
+
+const createUser = (firstname, lastname) => {
+  return cy.request({
+    url: `${Cypress.env('BACKEND_URL')}/nice2/rest/entities/User`,
+    method: 'POST',
+    json: true,
+    body: {model: 'User', key: null, paths: {firstname, lastname, relGender: {display: 'Weiblich', key: '2'}}}
```
rzueger: what's the `display` property for?

cypress/integration/entitybrowser_spec.js
```diff
@@ -1,59 +1,88 @@
...
   before(() => {
     cy.login()
   })
 
-  describe('ListView', () => {
-    beforeEach(() => {
-      cy.visit(`${helpers.getStoryUrl(['Apps', 'Entity Browser'], 'Entity Browser')}`)
-    })
+  describe('Create', () => {
+    it.only('should navigate to create view wit NEW button', function() {
```
rzueger: `function() {...` -> why not arrow function?

cypress/integration/entitybrowser_spec.js
rzueger: hardcoded key `2` could lead to problems someday

cypress/integration/entitybrowser_spec.js
```diff
@@ -62,17 +91,98 @@ describe('Entity Browser', function() {
...
       cy.get('[data-cy=entity-detail_back-button]').click()
       cy.contains('ungespeicherte Änderungen')
     })
+
+    it('should allow update of various fields', function() {
+      cy.get('#input-detailForm-firstname').type('!')
+      cy.get('#input-detailForm-lastname').type('!')
+      cy.get('#input-detailForm-email').type('test@tocco.ch')
+      cy.get('#input-detailForm-callname').type('cyp')
+      cy.get('#input-detailForm-relGender').type('{downarrow}').get('#react-select-3-option-1').click()
```
rzueger: `#react-select-3-option-1` looks like it could lead to problems someday - I guess we don't control the id of the option?

cypress/integration/entitybrowser_spec.js
MethodenMann: cypress does not fully support arrow functions in combination with this.* for aliases. 
https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#Aliases

But we dont need that anyways --> i change all  to arrow syntax

cypress/integration/entitybrowser_spec.js
MethodenMann: you are right but for the moment i dont have a better solution

cypress/integration/entitybrowser_spec.js
MethodenMann: Do you think the pk of a gender can change some days? I think anything else than hardcode would be an overkill.

PR #793, opened 2019-09-30
Pr/fix debounce

MethodenMann:


packages/tocco-util/src/react/Debouncer.spec.js
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable react/prop-types */
+import React from 'react'
+import {mount} from 'enzyme'
+
+import Debouncer from './Debouncer'
+
+describe('tocco-util', () => {
+  describe('hooks', () => {
```
rzueger: reaname 'hooks' to 'react'?

PR #796, opened 2019-10-03
Pr/fixing styled notifier

yilkur:


packages/app-extensions/src/notifier/components/StyledNotifier.js
```diff
@@ -177,12 +176,13 @@ const StyledNotifier = styled.div`
 
       .rrt-holder {
         opacity: .9;
+        
+        svg {
```
MethodenMann: With this change you will assume, that the Icon Component contains an svg element. This does not have to be the case. You have to set the style attribute on the icon component :  
`<Icon icon="square" style={{fonzSize: '5rem}}/>`

PR #799, opened 2019-10-07
Pr/list api

MethodenMann:


packages/app-extensions/src/rest/helpers.js
```diff
@@ -102,14 +102,26 @@ export function* fetchEntities(
...
 
-  const response = yield call(requestSaga, `forms/${formName}`)
+  const options = {
+    ...(allowNotFound && {acceptedStatusCodes: [404]})
+  }
+
+  const response = yield call(requestSaga, `forms/${formName}`, options)
+
+  if (allowNotFound && response.status === 404) {
+    cache.add('form', formName, null)
```
rzueger: what's the point of adding it to the cache? as far as I can see, there is no immediate return if `null` is obtained from the cache on line 110. Or should the condition on line be `if (cachedForm || allowNotFound)` to prevent an additional request on line 118 if the form has already been fetched as wasn't found?

PR #811, opened 2019-10-14
Pr/date picker

MethodenMann:


packages/scheduler/src/components/NavigationFullCalendar/NavigationFullCalendar.js
```diff
@@ -37,42 +48,32 @@ const NavigationFullCalendar = props => {
...
-            icon={isLoading ? '' : 'sync'}
-            look="raised"
-            onClick={() => { if (!isLoading) { refresh() } }}
-            pending={isLoading}
-            title={msg('client.scheduler.reload', intl)}
-          />
-        </Menu.Item>
-        <Menu.Item>
-          <Typography.H3>{title}</Typography.H3>
+          <DatePicker
```
rzueger: entire clickable area should have pointer cursor (both the date and arrow down)

packages/tocco-ui/src/DatePicker/useDatePickr.js
```diff
@@ -0,0 +1,61 @@
+import {useEffect, useState} from 'react'
+import '!style-loader!css-loader!flatpickr/dist/themes/light.css'
+import flatpickr from 'flatpickr'
```
rzueger: apparently, there's a plugin for a month picker called `monthSelectPlugin` which could be useful for the monthly mode (https://github.com/flatpickr/flatpickr/issues/269)

but I reckon you talked about this and decided that it wouldn't be worth the effort to implement a different picker for the monthly view.

MethodenMann: Haven't seen this plugin. Thanks. I talked to juan and we both think (at the moment) it's not worth the effort.

PR #814, opened 2019-10-17
Pr/admin resource scheduler

MethodenMann:


packages/admin/src/components/Navigation/Navigation.js
```diff
@@ -33,6 +33,18 @@ EntityExplorerMenuEntry.propTypes = {
...
 
+const ActionMenuEntry = ({onClick, item}) => {
+  return <StyledMenuLink onClick={onClick} to={`/a/${item.name}`}>{item.label}</StyledMenuLink>
+}
+
+ActionMenuEntry.propTypes = {
+  item: PropTypes.shape({
+    name: PropTypes.string,
+    label: PropTypes.string
+  }),
```
rzueger: render function looks like `item`, `name` and `label` are required

PR #825, opened 2019-11-12
refactor(scheduler): ui adjustments

MethodenMann:
- add custom views with custom date formatters
- make font size smaller
- fix resizing by adding plugin
- add min and max time (6-23:00)


packages/scheduler/src/components/FullCalendar/FullCalendar.js
```diff
@@ -146,13 +147,16 @@ const FullCalendar = ({
     })
   }
 
+  const getFormattedTime = date =>
```
rzueger: some documentation about expected input and output would be useful.
maybe introduce more global util, if it could be used somewhere else as well?

rzueger:
Commit message:

"Changelog: Make scheduler smaller (Date formatting, time slot, stying)"
-> "... styling"?

PR #828, opened 2019-11-19
feat(admin): Get list view store from cache when user returns to a list

rzueger:
- For instance, when user selects a User from the list and then
  loads the list of related Email_archive entities and then returns
  to the list of Users, the User list should be rendered just like
  he left it (same selection, same table page etc.)
- Whenever the user steps down in hierarchy (to deeper levels),
  we don't want to fetch any stores from the storage. That's why all
  stores of the deeper levels are removed in `ListView`.
- As we don't want to fetch the store for the main list from the
  storage when a user navigates to the list from the main menu,
  we clear the store storage when a menu item is clicked (in `Admin`
  component)
- The `keepStore` property of the entity-list app is not supported
  anymore. The usage of of the entity-list in the entity-browser
  app has been refactored for the props (`store` and `onStoreCreate`).
  -> Still to be refactored: SubGrid, RelationsView

Refs: TOCDEV-1078


packages/admin/src/routes/entities/components/ListView/ListView.js
```diff
@@ -47,15 +47,23 @@ const ListView = ({match, history, currentViewInfo, emitAction}) => {
       onNavigateToCreate={handleNavigateToCreate}
       searchFormPosition="left"
       searchFormType="admin"
+      store={persistedViewInfo.store}
```
rzueger: `persistedViewInfo` never null here?

MethodenMann: no. never.

`export const viewInfoSelector = (state, location) =>
  _get(state, ['viewPersistor', 'persistedViews', location, 'info'], {})`

PR #830, opened 2019-11-21
Pr/rest helpers

MethodenMann:


packages/tocco-util/src/mockData/entities.js
```diff
@@ -210,6 +216,19 @@ const extractBodyParams = body => () => {
   return {limit, offset, sort, search, where}
 }
 
+const createEntitiesDisplayResponse = (entityStore, timeout) =>
+  (url, opts) =>
+    sleep(timeout).then(() => ({
+      data: JSON.parse(opts.body).map(val => {
+        return {
```
rzueger: (return not necessary)

packages/tocco-util/src/mockData/entities.js
```diff
@@ -210,6 +216,19 @@ const extractBodyParams = body => () => {
...
+          model: val.model,
+          values: val.keys.map(key => {
+            return {key, display: `${val.model} (${key})`}
```
rzueger: (return not necessary)

PR #831, opened 2019-11-21
Pr/login screen

yilkur:


packages/login/src/components/StyledLoginForm.js
```diff
@@ -26,15 +28,36 @@ export const StyledLoginFormInputWrapper = styled.div`
...
+  &&&& {
+    margin-right: 0;
+  }
+`
+
+export const StyledPasswordButton = styled(Button)`
+  &&& {
+    margin-right: 0;
+    background-color: transparent;
+    color: ${theme.color('signal.danger.text')};
```
MethodenMann: i would use the primary color instead of danger that suggest something negative

packages/login/src/components/StyledLoginForm.js
```diff
@@ -26,15 +28,36 @@ export const StyledLoginFormInputWrapper = styled.div`
   }
 `
 
-export const StyledLoginFormWrapper = styled.div`
+export const StyledLoginButton = styled(Button)`
+  &&&& {
```
MethodenMann: The loading icon should be scaled as well 
image

MethodenMann: &&&& seems a bit odd to me. are you sure there is nothing a bit more elegant to solve this :)

packages/login/src/components/StyledLoginForm.js
```diff
@@ -1,10 +1,12 @@
 import styled from 'styled-components'
 import {
   StyledInputCss,
-  StyledEditableWrapperCss
+  StyledEditableWrapperCss,
+  Button
 } from 'tocco-ui'
 import {StyledButton} from 'tocco-ui/src/Button'
 import {StyledButtonGroup} from 'tocco-ui/src/ButtonGroup'
+import {theme} from 'tocco-ui/src/utilStyles'
```
MethodenMann: wrong import. should be:
`import {theme} from 'tocco-ui`

PR #832, opened 2019-11-22
refactor(entity-list): replace parent search field

MethodenMann:
- if a search form has a parent condition and a search-field with the same
  path, this field must not be a full-text search field. And it should be
  read only. This change makes sure such fields are replaced with a readonly
  single-select field.

Refs: TOCDEV-1084
Changelog: Fix parent Search field bug

### DIFF
--- a/packages/entity-list/src/util/api/forms.js
+++ b/packages/entity-list/src/util/api/forms.js
@@ -60,3 +60,29 @@ const searchChildren = node => {
     }), {})
   }
 }
+
+/**
+ * If the search form contains a field pointing to the parent entity this must not be a full-text
+ * search field and should be readonly. This function will search for the correct field and replace its
+ * type and change the readonly attribute.
+ */
+export const changeParentFieldType = (formElement, parentPath) => {
+  if (formElement.componentType === 'field-set') {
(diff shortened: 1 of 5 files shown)


packages/entity-list/src/util/api/forms.js
```diff
@@ -60,3 +60,29 @@ const searchChildren = node => {
...
+          ? {...child, dataType: 'single-select-box'}
+          : child)
+    }
+  }
+  if (formElement.children) {
+    return {
+      ...formElement,
+      children: formElement.children.map(child => changeParentFieldType(child, parentPath))
+    }
+  }
```
rzueger: what to return if `formElement.children` is not defined? IMO, there should be an explicit `return` statement for this case as well (don't return `undefined` implicitly).

packages/entity-list/src/util/api/forms.spec.js
```diff
@@ -258,6 +258,15 @@ describe('entity-list', () => {
             expect(result).to.eql(expectedResult)
           })
         })
+
+        describe('changeParentFieldType', () => {
+          test('should ', () => {
```
rzueger: should what? :)

IMO, `changeParentFieldType` deserves more than one test case

PR #837, opened 2019-12-04
refactor(admin): modify/extend styled components

yilkur:
Refs: TOCDEV-1166
Changelog: style top bar and menu


packages/admin/src/components/Admin/StyledComponents.js
```diff
@@ -1,39 +1,35 @@
...
 
-export const StyledTitle = styled(StyledH1)`
- && {
-   color: white;
-   padding-left: 8px;
-   padding-top: 5px;
-   font-size: 2.5rem;
- }
-`
+import toccoLogo from '../Login/img/tocco-circle.svg'
```
MethodenMann: tocco-circle.svg is a shared image. i dont think it should be located in the Login folder.

packages/admin/src/components/Navigation/StyledComponents.js
```diff
@@ -1,53 +1,52 @@
...
   overflow: auto;
+  padding-left: 2rem;
+  
+  div > div:last-child {
+    padding-bottom: 2rem;
+  }
+
   ::-webkit-scrollbar {
-      width: 5px;
+      width: 10px;
```
MethodenMann: i think we should move a scrollbar css into tocco-ui since we want to reuse the same scrollbar in various views.

packages/tocco-theme/src/ToccoTheme/defaultTheme.js
```diff
@@ -3,6 +3,9 @@ export default {
     paper: '#fff',
     primary: '#9E2124',
     text: '#4f5e60',
+    background: '#EDF1F5',
```
MethodenMann: i would be **very** restrictive when it comes to extending the theme.
Every var should serve a sole purpose. Maybe it's best to start with a documentation where you write for what each var is used for?
A theme should allow to style the application for every customer in their corporate colors.

Concern one: What is the difference between paper and background? What is used when is not clear to me.

Concern two: As said above a  theme should allow to set styling aspects of the application without assuming any color palette. So its would be possible to set primaryBlue to Pink? If there is no way around these two new vars i would at least give them a more generic name (and document well for what they are used for)

packages/admin/src/components/Navigation/StyledComponents.js
```diff
@@ -1,53 +1,52 @@
...
   
   &:hover {
-      color: #555;
+      color: ${theme.color('secondaryBlue')};
     }
   
   &:focus {
-     text-decoration: underline;
-      outline: none;
+     outline: none;
```
MethodenMann: i think text-decoration or something similar is still needed. It is possible to select a link with tab. But witout a distinction in styling there is no way to tell what link has the focus.

packages/tocco-ui/src/SearchBox/SearchBox.js
```diff
@@ -2,9 +2,8 @@ import PropTypes from 'prop-types'
 import React from 'react'
```
MethodenMann: SearchBox styling will also affect the searchbox used in Detail > Relation Preview. But there we have a litte imperfection:
image

PR #841, opened 2019-12-09
Pr/breadcrumbs

yilkur:


packages/admin/src/routes/entities/components/Breadcrumbs/Breadcrumbs.js
```diff
@@ -24,9 +16,9 @@ const Breadcrumbs = ({breadcrumbsInfo}) => {
       breadcrumbsInfo
         .map((b, idx) => {
           return <Typography.Span key={idx}>
-            <StyledLink to={`/e/${b.path}`}>
+            <StyledBreadcrumbsLink neutral to={`/e/${b.path}`}>
```
MethodenMann: what is this "neutral" for?

packages/admin/src/routes/entities/components/Breadcrumbs/StyledBreadcrumbs.js
```diff
@@ -0,0 +1,37 @@
...
+  width: 100%;
+  padding: 1.3rem 1.7rem;
+  margin-bottom: 6px;
+  
+  span:nth-child(even) {
+    margin-left: 1.5rem;
+    margin-right: 1.5rem;
+  }
+  
+  span:last-child a {
```
MethodenMann: alternatively you could have set a property in the loop (Breadcrumbs.js).

` <StyledBreadcrumbsLink active={idx === breadcrumbsInfo.length -1} .../>`

PR #846, opened 2019-12-20
Pr/button style

yilkur:


packages/tocco-ui/src/StatedValue/StyledStatedValue.js
```diff
@@ -97,6 +97,15 @@ const StyledStatedValueBox = styled.div`
     > ${StyledHtmlFormatter} {
       margin-bottom: 0;
     }
+    
+    button {
```
MethodenMann: i dont think we should overwrite styles like this. 

imo we either create a own component for the Input fields or we change the standard Button look 'Ball' so it would fit the purpose.

PR #857, opened 2020-01-14
Pr/filter styles

yilkur:


packages/entity-list/src/components/SearchFilterList/SearchFilterList.js
```diff
@@ -1,33 +1,71 @@
...
-import {Typography, Button} from 'tocco-ui'
+import {Typography, Button, theme, StyledSpan} from 'tocco-ui'
 import styled from 'styled-components'
 
+const SearchFilterWrapper = styled.div`
+  margin-top: .4rem;
+  margin-bottom: .4rem;
+`
+
+const StyledSpanSearchFilter = styled(StyledSpan)`
```
MethodenMann: Is it not possible to make a Typography.Span selector in StyledSearchFilterButton and set display and width there?
(is the Span needed anyway?)

packages/tocco-theme/src/ToccoTheme/defaultTheme.js
```diff
@@ -1,7 +1,7 @@
 export default {
   colors: {
     paper: '#fff',
-    primary: '#9E2124',
+    primary: '#B22A31',
```
MethodenMann: is this the official tocco red? Was it wrong the whole time? :)

packages/entity-list/src/components/SearchFilterList/SearchFilterList.js
```diff
@@ -43,14 +81,9 @@ const AdminSearchForm = ({searchFilters, setSearchFilterActive, executeSearch})
     return <div style={{paddingLeft: '8px'}}><Typography.I>No Searchfilters available</Typography.I></div>
   }
 
-  return <div>
+  return <SearchFilterWrapper>
```
MethodenMann: Is a wrapper for SearchFilterList so maybe this name is not 100% accurate.

PR #858, opened 2020-01-15
Pr/fix split panel

yilkur:


packages/tocco-theme/src/ToccoTheme/defaultTheme.js
```diff
@@ -5,6 +5,7 @@ export default {
     secondary: '#11587f',
     secondaryLight: '#9BC0D5',
     text: '#4f5e60',
+    border: '#CCCCCC',
```
MethodenMann: Would it be sufficient to calculate a variation to the background color with polished like lighten?
The theme ist getting more complex this could lead to a problem.

PR #860, opened 2020-01-16
refactor(admin): change welcome screen

yilkur:
Refs: TOCDEV-1347
Changelog: change welcome screen


packages/admin/src/routes/dashboard/components/Dashboard/Dashboard.js
```diff
@@ -1,23 +1,55 @@
...
-  justify-content: center;
+  height: 75%;
+  justify-content: flex-end;
+  align-items: flex-end;
+  margin-right: -180px;
   
-  div { 
-      padding:  15px;
+  && {
+    * {
```
MethodenMann: Are you using the asterisk selector to select the tile and later to select the version text? Isn't there a more transparent way to to that?

PR #862, opened 2020-01-21
Pr/list

yilkur:


packages/resource-scheduler/src/components/ResourceScheduler/ResourceScheduler.js
```diff
@@ -22,7 +34,7 @@ class ResourceScheduler extends React.Component {
     const props = this.props
     return (
       <React.Fragment>
```
MethodenMann: In this case you can remove React.Fragment. Fragment is only a wrapper that does not create a dom element. See react docs for more information

PR #876, opened 2020-02-14
refactor(admin): make login responsive

yilkur:
Refs: TOCDEV-1467
Changelog: make login responsive


server/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
   <title>Tocco</title>
   <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" name="viewport"/>
```
MethodenMann: This will only affect the local dev server. If important on not already done, this has to be done on nice side as well.

packages/admin/src/assets/tocco_white.svg
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="utf-8"?>
```
MethodenMann: Ist a new file necessary?  Usually the color of a svg can be changed and of course the orientation.

PR #877, opened 2020-02-14
refactor(entity-list): crop search filter text accordingly

yilkur:
Refs: TOCDEV-1559
Changelog: crop search filter text with ellipsis

### DIFF
--- a/packages/entity-list/src/components/SearchFilterList/SearchFilterList.js
+++ b/packages/entity-list/src/components/SearchFilterList/SearchFilterList.js
@@ -10,8 +10,7 @@ const SearchFilterListWrapper = styled.div`
 `
 
 const StyledSpanSearchFilter = styled(StyledSpan)`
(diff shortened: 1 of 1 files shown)


packages/entity-list/src/components/SearchFilterList/SearchFilterList.js
```diff
@@ -57,15 +56,14 @@ export const StyledSearchFilterButton = styled.div`
...
-      <StyledButton
-        onClick={e => {
-          setActive(false)
-          e.stopPropagation()
-        }}
-        icon="plus"
-        dense />
-    </StyledSpanSearchFilter>
+    <StyledSpanSearchFilter>{label}</StyledSpanSearchFilter>
+    <StyledButton
```
MethodenMann: When text is cropped, the full text should be visible on mouseover > set title on button

PR #881, opened 2020-02-17
chore: pipe some requests through to Nice2 instance in dev server

rzueger:
If a /nice2/* request reaches the dev server because it wasn't mocked,
it's probably better to let the Nice2 instance (which is probably
running on http://localhost:8080) answer the request, instead of
returning the index.html.

### DIFF
--- a/server/main.js
+++ b/server/main.js
@@ -4,6 +4,7 @@ const opn = require('opn')
 const express = require('express')
 const webpack = require('webpack')
 const compress = require('compression')
+const request = require('request')
 
 const webpackConfig = require('../build/webpack.config').default
 const config = require('../config').default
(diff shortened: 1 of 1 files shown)


server/main.js
```diff
@@ -42,6 +43,15 @@ if (config.env === 'development') {
   app.use(express.static(publicPath))
   app.use('/static', express.static('server/static'))
 
+  // Most probably the following requests should be answered by the Nice2 instance
+  // -> pipe them through
+  app.use(['/nice2/*', '/js/*', '/img/*'], function(req, res, next) {
+    const newUrl = 'http://localhost:8080' + req.originalUrl
```
MethodenMann: Maybe use __BACKEND_URL__ instead of hardcoded 8080. Then it should work with master backend as well

PR #885, opened 2020-02-21
refactor(app-extensions): load legacy action sources async when needed

rzueger:
Now the sources are loaded when the legacy action handler is invoked
for the first time. As each file depends on the prior file we have
to load them sequentially.

### DIFF
--- a/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
+++ b/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
@@ -1,4 +1,51 @@
+import {call} from 'redux-saga/effects'
+
+export const loadScript = src => new Promise((resolve, reject) => {
+  const s = document.createElement('script')
+  s.src = src
+  s.onload = resolve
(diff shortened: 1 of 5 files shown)


packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -1,30 +1,74 @@
...
+  {src: '/css/themes/blue-medium.css', handler: loadCss},
+  {src: '/css/nice2-admin.css', handler: loadCss}
+]
 
-  action.getSelectionNumber = function() {
-    return selection.ids.length
-  }
-  action.getSelection = function() {
-    return {
+const initLegacyActionsEnv = () => window.legacyActionsEnvInitialized === true
```
MethodenMann: Have you noticed that this is a generator? Usually there is no need for Promises and suchlike. And since neither loadScript nor loadCss is async i dont get the hassle. Or am i missing something?

PR #894, opened 2020-02-26
feat(app-extension): add handler for legacy actions (ExtJs)

rzueger:
- The first time a legacy action handler is executed, the required
  ExtJS and legacy-client sources are loaded
- There's are still some things to do. For instance, there are several
  types of record selections that must be supported eventually (e.g.
  actual selection of some particular records by key or a set of entities
  defined by query conditions only). At the moment, only actual selection
  by entity key is implemented.

Refs: TOCDEV-1507

### DIFF
--- a/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
+++ b/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
@@ -1,0 +1,73 @@
+import {call} from 'redux-saga/effects'
+
+export const loadScript = src => new Promise((resolve, reject) => {
+  const s = document.createElement('script')
+  s.src = src
+  s.onload = resolve
+  s.onerror = reject
+  document.head.appendChild(s)
+})
+
+export const loadCss = src => new Promise((resolve, reject) => {
+  const head = document.getElementsByTagName('head')[0]
+  const link = document.createElement('link')
+  link.rel = 'stylesheet'
+  link.type = 'text/css'
+  link.href = src
+  link.media = 'all'
+  link.onload = resolve
+  link.onerror = reject
+  head.appendChild(link)
+})
+
+export function* loadSequentially(sources) {
+  for (const source of sources) {
+    yield call(source.handler, source.src)
+  }
+}
+
+export const sources = [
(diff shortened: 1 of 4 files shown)


packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -0,0 +1,77 @@
...
+  link.type = 'text/css'
+  link.href = src
+  link.media = 'all'
+  link.onload = resolve
+  link.onerror = reject
+  head.appendChild(link)
+})
+
+export function* loadSequentially(sources) {
+  for (const source of sources) {
```
MethodenMann: same result but i personally prefer:

`yield all(sources.map(source => call(source.handler, source.src)))`

packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const ctx = new window.nice2.modules.entityexplorer.EntityExplorerActionContext(entityExplorer, null)
+
+  const situation = {
+    entityName: selection.entityName
+  }
+
+  const action = window.NetuiActionRegistry.newActionFromDefinition(actionDefinition, situation, ctx)
+
+  action.getSelectionNumber = function() {
```
MethodenMann: `action.getSelectionNumber = () => selection.ids.length`
works?

packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -0,0 +1,77 @@
...
+    return selection.ids.length
+  }
+  action.getSelection = function() {
```
MethodenMann:
```
action.getSelection =() => (
   {
      entityName: selection.entityName,
      selectedEntities: selection.ids
    }
)
```

packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
rzueger: not possible to use `all` here as it yields the `call` effects **in parallel**. we need to load the sources **sequentially**.

PR #897, opened 2020-03-04
feat(admin): add arrow key navigation support to Navigation

MethodenMann:
Refs: TOCDEV-1281
Changelog: Arrow navigation


packages/admin/src/components/Navigation/Navigation.js
```diff
@@ -96,7 +97,22 @@ const Navigation = ({modulesMenuTree, settingsMenuTree, menuOpen, onClick, activ
 
   const msg = id => intl.formatMessage({id})
 
-  return <StyledNav>
+  const handleCursorNavigation = key => {
+    const focusableElements = navigationEl.current.querySelectorAll('a, input')
+    const currentIndex = [...focusableElements].findIndex(el => document.activeElement.isEqualNode(el))
```
rzueger: what if `focusableElements` is empty or if the currently active element is not found in the `focusableElements` array for whatever reason? simply let it fail on line 105?

PR #898, opened 2020-03-05
Pr/sso login button fix

yilkur:


packages/sso-login/src/components/ProviderButton/StyledProviderButton.js
```diff
@@ -25,5 +27,8 @@ export default styled(StyledButton)`
       padding: .9rem 0;
     }
 
-   margin-right: 0;
+    &:hover * {
+      color: ${theme.color('paper')};
```
MethodenMann: an sso-button can have a dedicated font color and background color:
```
    bg: props.primaryColor,
      fg: props.secondaryColor || `${_get(props.theme, 'colors.paper')}, ${_get(props.theme, 'colors.text')}`

```

Wouldn't this change be unfavorable if the background color is a very light color?

PR #900, opened 2020-03-06
feat(input-edit): create InputEditTable

PadPalon:
feat(input-edit): create InputEditTable

- setup package input-edit
- add type mapping to formattedValueFactory


packages/input-edit/src/components/inputedittable/InputEditTable.js
```diff
@@ -0,0 +1,168 @@
...
+  const dataCells = dataFormCells.map(column =>
+    <DataCell key={keyBuilder(index, column)} index={index} nodes={nodes} column={column}/>
+  )
+  const editCells = inputEditForm.map(column =>
+    <InputCell key={keyBuilder(index, column)} index={index} nodes={nodes} column={column} updateValue={updateValue}/>
+  )
+  return [...dataCells, ...editCells]
+}
+
+const DataCell = ({index, nodes, column}) => {
```
MethodenMann: could be refactored to direct return (const var is not really necessary)

packages/input-edit/src/components/inputedittable/InputEditTable.js
```diff
@@ -0,0 +1,168 @@
...
+  const field = column.children[0]
+  return <td>
+    {formField.formattedValueFactory(field.dataType)(
+      field, undefined, 'Input_edit_data_list', nodes[field.path], {}, {}, {}, `${index}:${column.id}`
+    )}
+  </td>
+}
+
+const InputCell = ({index, nodes, column, updateValue}) => {
+  const width = column.dataType === 'single-select'
```
MethodenMann: :laughing:

MethodenMann: please check this with yilmaz. As a temporary workaround until the layout is a bit more mature we can leave it :)

packages/input-edit/src/modules/inputEditTable/actions.js
```diff
@@ -0,0 +1,59 @@
...
+
+export const setDataForm = form => ({
+  type: SET_DATA_FORM,
+  payload: form
+})
+
+export const updateValue = (inputDataKey, node, value) => ({
+  type: UPDATE_VALUE,
+  payload: {
+    inputDataKey: inputDataKey,
```
MethodenMann: shorthand can be used 

```
payload: {
    inputDataKey,
    node,
....
```

packages/input-edit/src/modules/inputEditTable/reducer.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const setValue = (state, action) => {
+  const {inputDataKey, node, value} = action.payload
+  const indexToUpdate = Object.keys(state.data).find(index => state.data[index].pk === inputDataKey)
+  return {
+    ...state,
+    data: {
+      ...state.data,
+      [indexToUpdate]: {
```
MethodenMann: wouldnt it be more practical to use the pk as index?

packages/input-edit/src/modules/inputEditTable/sagas.js
```diff
@@ -0,0 +1,66 @@
...
+export function* sortData({payload}) {
+  yield call(loadData, payload.sorting)
+}
+
+export function* loadData(action, sorting) {
+  const {inputEntityKey} = yield select(inputSelector)
+  const data = yield call(rest.requestSaga, `inputEdit/${inputEntityKey}/data/search`, {
+    method: 'POST',
+    body: {
+      sorting: sorting || (yield select(inputEditSelector)).sorting
```
MethodenMann: i think its more efficient and cleaner to just read the sorting on line 34

packages/app-extensions/src/formField/formattedValueFactory.js
```diff
@@ -17,7 +17,19 @@ const getOptions = (type, modelField, formData) => {
   return {}
 }
 
+const typeMapping = {
```
MethodenMann: maybe we can find a  better place for this mapping. i will look into it and give you a more detailed feedback

packages/input-edit/src/components/inputedittable/InputEditTable.js
```diff
@@ -0,0 +1,168 @@
...
+import {EditableValue} from 'tocco-ui'
+import {consoleLogger} from 'tocco-util'
+import {formField} from 'tocco-app-extensions'
+
+import {StyledCell, StyledHeader, StyledTable} from './StyledComponents'
+
+const InputEditTable = ({data, inputDataForm, inputEditForm, updateValue, sorting, setSorting}) => {
+  const dataFormCells = getDataFormCells(inputDataForm)
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
PadPalon: I didn't manage to quite adapt your solution to fit my table, but it inspired me to adjust mine to not be globally registered and be independent of the size of the table.

packages/input-edit/src/components/inputedittable/InputEditTable.js
PadPalon: Will do.
Sorry, I meant to warn you about this "special" piece of code :P As mentioned today, a surprising amount of people say this is the way to do it, but I'm sure there is a better way.

packages/input-edit/src/modules/inputEditTable/reducer.js
PadPalon: Good point. I used to do it like that, but then changed it around when implementing the arrow key navigation, because it made that way easier. But in the end I changed another thing that does not need these indexes anymore. So now this can be switched back as well =)

packages/input-edit/src/components/InputEditTable/InputEditTable.js
```diff
@@ -0,0 +1,158 @@
...
+    const idToFind = `${currentIndex + indexDiff}:${node}`
+    const elementToFocus = document.getElementById(idToFind)
+    if (elementToFocus) {
+      elementToFocus.focus()
+      elementToFocus.select()
+    }
+  }
+}
+
+const getDirection = ({field, direction}, newSortingField) => {
```
MethodenMann:
```
const getDirection = ({field, direction}, newSortingField) => {
  if (field === newSortingField) {
    return direction === 'asc' ? 'desc' : 'asc'
  } else {
    return 'asc'
  }
}
```

shorter version (imo easier to read)

packages/input-edit/src/modules/inputEditTable/reducer.js
```diff
@@ -0,0 +1,39 @@
...
+  inputEditForm: [],
+  inputDataForm: {},
+  sorting: {}
+}
+
+const setValue = (state, action) => {
+  const {inputDataKey, node, value} = action.payload
+  return {
+    ...state,
+    data: state.data.reduce((val, acc) => {
```
MethodenMann: ok sorry my fault, the first parameter is the accumulator (acc) and the second is the currentValue (value). you sould probably rename them. sorry :)

PR #905, opened 2020-03-12
refactor(app-extensions): improve modal ux

yilkur:
Refs: TOCDEV-1335
Changelog: improve modal ux

### DIFF
--- a/packages/app-extensions/src/notifier/modules/modalComponents/StyledModalContent.js
+++ b/packages/app-extensions/src/notifier/modules/modalComponents/StyledModalContent.js
@@ -1,8 +1,14 @@
 import styled from 'styled-components'
 import {
   scale,
-  theme
+  theme,
+  StyledScrollbar
 } from 'tocco-ui'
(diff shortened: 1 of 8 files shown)


MethodenMann:
image

i think this need a small adjustment. Storybook > Admin > Person > Detail > Open Nationality Remote

MethodenMann:
image

Addresse > Veranstaltung Extranet

MethodenMann:
image

on small devices the padding-right is not correct?

packages/app-extensions/src/notifier/modules/modalComponents/StyledModalContent.js
```diff
@@ -48,6 +64,39 @@ const StyledModalContent = styled.div.attrs({
         outline: none;
       }
     }
+
+    ${StyledTitleWrapper}, ${StyledAdvancedSearchButtonWrapper} {
```
MethodenMann: works like a charm but i think in this case we should prefer class names / selector over styled-component. imo this would be more transparent and less coupled. Import gives it away ('../../../')

PR #906, opened 2020-03-12
feat(app-extension): support action properties in legacy actions

rzueger:
Refs: TOCDEV-1570

### DIFF
--- a/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
+++ b/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
@@ -1,4 +1,5 @@
-import {call} from 'redux-saga/effects'
+import _isPlainObject from 'lodash/isPlainObject'
+import {call, all} from 'redux-saga/effects'
 
 export const loadScript = src => new Promise((resolve, reject) => {
   const s = document.createElement('script')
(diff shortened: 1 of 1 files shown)


packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -48,6 +49,11 @@ export default function* (definition, selection, parent, params, config) {
 
   const actionDefinition = new window.nice2.netui.actions.model.ClientActionDefinition(definition.id, definition.id)
   actionDefinition.setEnabled(!definition.readOnly)
+  if (_isPlainObject(definition.properties)) {
+    Object.keys(definition.properties).forEach(propertyName => {
+      actionDefinition.setProperty(propertyName, definition.properties[propertyName])
```
MethodenMann: if you want to test your code later you probably have to rewrite it to `yield call(actionDefintion.setProperty, propertyName....`

packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -48,6 +49,11 @@ export default function* (definition, selection, parent, params, config) {
 
   const actionDefinition = new window.nice2.netui.actions.model.ClientActionDefinition(definition.id, definition.id)
   actionDefinition.setEnabled(!definition.readOnly)
+  if (_isPlainObject(definition.properties)) {
+    Object.keys(definition.properties).forEach(propertyName => {
```
MethodenMann: (purely a matter of taste) `Object.entries(definition.properties).forEach(([key, value]) => actionDefinition.setProperty(key, value)) ` would save you the bracket property notation.

PR #907, opened 2020-03-12
feat(tocco-ui): introduce animated burger icon

MethodenMann:
- use in Admin

Changelog: Animate burger icon

### DIFF
--- a/packages/admin/src/components/Admin/Admin.js
+++ b/packages/admin/src/components/Admin/Admin.js
@@ -2,8 +2,9 @@ import React, {useState, useEffect} from 'react'
 import PropTypes from 'prop-types'
 import {Router, Route, Redirect, Switch} from 'react-router-dom'
 import {createBrowserHistory} from 'history'
-import {Icon, LoadMask} from 'tocco-ui'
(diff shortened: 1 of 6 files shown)


packages/admin/src/components/Admin/Admin.js
```diff
@@ -68,7 +68,7 @@ const Admin = ({
             isOpen={menuOpen}
             onStateChange={isMenuOpen}
             customCrossIcon={false}
-            customBurgerIcon={menuOpen ? <Icon icon="times"/> : <Icon icon="bars"/>}
+            customBurgerIcon={<BurgerButton isOpen={menuOpen} size={20} color="white"/>}
```
yilkur: can't we grab the theme color paper here?

packages/admin/src/components/Admin/StyledComponents.js
```diff
@@ -31,7 +31,7 @@ export const burgerMenuStyles = {
     width: '24px',
     height: '24px',
     left: '13px',
-    top: '9px',
+    top: '13px',
     color: '#fff'
```
yilkur: maybe it would be also better to grab the theme color of paper here?

PR #909, opened 2020-03-16
Pr/background adaptations

yilkur:


MethodenMann:
isn't it possible to set the background in a shared root component?

> packages/app-extensions/src/appFactory/App.js

PR #911, opened 2020-03-18
Pr/range

MethodenMann:


MethodenMann:
> Looks good to me.
> 
> Two questions:
> 
> There exists still the other Date-range component. What are we going to do with that one?
> 
> If we're going to keep it too, would it make sense to use the same value object for both of them?
> 
> On-change event of the existing Date-Range component looks like this:
> 
> changed: (1) [Object]
> 
> * 0: Object
>   -- from: Tue Dec 03 2019 00:00:00 GMT-0500 (Ecuador Time)
>   -- to: Thu Dec 12 2019 00:00:00 GMT-0500 (Ecuador Time)
> 
> On-change event of the new range component looks like this:
> 
> changed: (1) [Object]
> 
> * 0: Object
>   -- from: "2020-01-06"
>   -- isRangeValue: true
>   -- to: "2020-12-31"
> 
> Would it make sense to validate that the `to` value cannot be before the `from` value and vice versa - or grey out the invalid range in the second picker, if the first value has already been picked?

1. i deleted the old date range component, no need for that anymore
2. good point. a added something for the date and datetime type

PR #912, opened 2020-03-18
Pr/shy

MethodenMann:


packages/tocco-ui/src/StatedValue/StatedValue.js
```diff
@@ -41,8 +42,10 @@ const StatedValue = props => {
     touched
   } = props
 
+  const adjustedLabel = js.adjustedHTMLString(label)
```
rzueger: why do you think we cannot interpret the labels als HTML? aren't they all coming from our language properties files?

MethodenMann: you are right. since this is a trusted source we should be allowed to that. i changed it

PR #924, opened 2020-04-01
New Table

MethodenMann:


packages/entity-list/src/components/Table/SortingState.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import {Icon} from 'tocco-ui'
+
+const SortingState = ({column, sorting}) => {
+  const isSortedBy = () => {
+    let result = null
+    sorting.forEach((s, idx) => {
```
rzueger, line 8: (sorting not marked as required)

packages/entity-list/src/components/Table/StaticCell.js
```diff
@@ -0,0 +1,37 @@
...
+
+import cellRenderer from '../../util/cellRenderer'
+
+const StaticCell = React.memo(props => {
+  const {column, rightAligned, entity, refresh, intl, parent} = props
+  return <td
+    {...rightAligned === true && {style: {textAlign: 'right'}}}
+    data-cy="list-cell"
+  >
+    {column.cellRenderer
```
rzueger, line 13: (column not marked as required)

packages/entity-list/src/components/Table/Table.js
```diff
@@ -1,141 +1,144 @@
...
+import useResize from './useResize'
+import ResizingController from './ResizingController'
+import SortingState from './SortingState'
+import columnEnhancers from './columnEnhancers'
+import StaticCell from './StaticCell'
+import {selectionStylePropType} from '../../util/selectionStyles'
+import StyledTable, {StyledTableWrapper, StretchingTableContainer, PaginationContainer} from './StyledTable'
 
-const RIGHT_ALIGNED_TYPES = ['moneyamount', 'counter', 'integer', 'long']
+const rightAlignedTypes = ['moneyamount', 'counter', 'integer', 'long']
```
rzueger, line 17: there's probably a bit more:

missing number base types:
- double
- decimal
- latitude
- longitude

missing virtual types:
- percent (decimal)
- sorting (integer)
- version (long)
- serial (long)

packages/entity-list/src/components/Table/Table.js
```diff
@@ -1,141 +1,144 @@
...
-          <span onClick={e => e.stopPropagation()}>
-            {props.linkFactory.detail(null, null, row.__key, <Icon icon="arrow-right"/>)}
-          </span>
-        )}
+  const ThContent = ({column}) =>
+    column.headerRender
+      ? column.headerRender({...props, singleSelectHandler, multiSelectHandler, isSelected})
+      : column.label
+
+  const InProgressRow = () => <tr><td className="fullRow progress">
```
rzueger, line 68: what are CSS class names needed for at all? why isn't everything styled purely with styled components without such selectors?

packages/entity-list/src/components/Table/Table.js
```diff
@@ -1,141 +1,144 @@
...
+    <LoadingSpinner size="20"/>
```
rzueger, line 69: (I know it's not a new component, but why is that size prop a string and not a number?)

packages/entity-list/src/components/Table/Table.js
```diff
@@ -1,141 +1,144 @@
...
+    </td></tr>
+
+  return <StyledTableWrapper>
+    <StretchingTableContainer>
+      <StyledTable ref={tableEl} columns={columns} resizingColumn={resizingColumn}>
+        <thead>
+          <tr>
+            {columns.map(column =>
+              <th
+                id={`header-cell-${column.id}`}
```
rzueger, line 89: what's the `id` prop required for? if only for cypress, i'd prefer `data-cy` attributes.

packages/entity-list/src/components/Table/Table.js
MethodenMann: as i said. styling is not finished yet. @yilkur will look into that

packages/entity-list/src/components/Table/Table.js
MethodenMann: id is required for resizing purposes. see:
https://github.com/tocco/tocco-client/blob/191c6ba38d987d4d92f47f6065edc29f4aa2db73/packages/entity-list/src/components/Table/useResize.js#L24

packages/entity-list/src/components/Table/Table.js
MethodenMann: because it is set as css attribute fontSize and therefore could have a unit. e.g. `2rem`

cypress/integration/entitybrowser_spec.js
```diff
@@ -135,12 +135,52 @@ describe('Entity Browser', () => {
 
     it('should load and display essential parts of the list view', () => {
       cy.contains('Daten werden geladen...', {timeout: LONG_TIMEOUT})
-      cy.get('[title="Vorname"]')
+      cy.get('#header-cell-firstname')
```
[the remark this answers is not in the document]
MethodenMann: because i have an ID that i use anyway. but you are right. its probably more consistent to only work with data-cy selectors in cypress. will change this

PR #925, opened 2020-04-01
Pr/detail deleted switch list

rzueger:


packages/entity-detail/src/modules/entityDetail/sagas.js
```diff
@@ -235,3 +236,16 @@ export function* actionInvoked(action) {
 export function* navigateToCreate({payload}) {
   yield put(externalEvents.fireExternalEvent('onNavigateToCreate', payload.relationName))
 }
+
+export const entityDeleted = (event, entityName, entityId) =>
+  !!event.payload.keys.find(key => key._entityName === entityName && key.getKey().asString() === entityId)
```
MethodenMann, line 241: Shouldn't we introduce an abstraction level for legacy remote events? 
` key.getKey().asString()` isn't that convenient . so is _entityName

```
payload: {
 entities: [{
   entity: 'User',
   key: '3'
}]
}
```

for example. i think at some point we want to dispatch remote events with the new api (sockets?) and there i think it would be nice if the payload matches the legacy events

packages/admin/src/routes/entities/components/EditView/EditView.js
```diff
@@ -28,6 +28,11 @@ const EditView = props => {
     }
   }
 
+  const handleEntityDeleted = () => {
```
MethodenMann, line 31: what will happen if the detail is in a dirty state?

rzueger: good point.

In this case, the same notification is displayed as when you're trying to navigate away manually ("... are you sure you want to leave ..?"). If you deny that question, you stay on the detail form of the deleted entity.

What would be the desired behavior in your opinion?

packages/entity-detail/src/modules/entityDetail/sagas.js
rzueger: Yes I thought about that too. But I'm not sure if it's worth the effort here because I don't really like how the legacy events are structured and fired. The three event types "created", "updated" and "deleted" are very different from each other in the legacy client.

The format you proposed would be useful and it works for the "deleted" event. However, we have a problem with the "created" and the "updated" event:

- "created": Here, only a list of the names of the models of the created entities is passed to the data registry (for instance, if we created 5 entities of type "User" and 2 of type "Principal", the data registry will only receive `['User', 'Principal']` instead of a list of all entity instances).
- "updated": Here, the event is fired for all updated entities one by one in the legacy client. So, the structure you proposed will always contain only one entity.

I suppose we won't use the "created" and "updated" events in the same way like the legacy client uses it. I suppose we will use the same structure for all 3 event types in the new web sockets solution for the new client -> the structure you proposed above (all entity instances in the payload array).

Because the current payload structure doesn't really match the desired structure, I also decided to call the legacy events "legacy-..." - if they matched the desired future payload structure, we wouldn't need to prefix them with "legacy".

---

What I'll do:

I'll change the payload of the "deleted" event to the structure you proposed above. So the payload of the "deleted" event will match the structure of the future event.

But unfortunately we can't really do that for the "created" and "updated" event - those we'll have to update resp. handle separately once our new web sockets solution is in place.

OK?

rzueger: Sorry, disregard. :)

I added new data registry methods for the new client which receives the entities in a consistent format for all three event types (https://git.tocco.ch/c/nice2/+/36375).

packages/admin/src/routes/entities/components/EditView/EditView.js
MethodenMann: The prompt makes no sense since the current entity is already deleted, right? Hence validation will immediately throw errors when the users stays on the page. But i think it can be tricky to ignore the prompt in this case. Or better said: not worth the effort for a legacy action?

PR #940, opened 2020-04-30
Pr/list actions

MethodenMann:


packages/admin/src/routes/entities/components/RelationsView/RelationsView.js
```diff
@@ -83,6 +90,11 @@ const RelationsView = ({
           <StyledLink to={match.url.replace(/(relations|detail)$/, selectedRelation.relationName)}>
             <Icon icon="arrow-right"/>
           </StyledLink>
+          <PreviewCreateWrapper>
+            <StyledLink to={match.url.replace(/(relations|detail)$/, selectedRelation.relationName) + '/create'}>
```
rzueger, line 94: (maybe define shape of `match` in proptypes)

packages/entity-list/src/components/ListView/ListView.js
```diff
@@ -46,15 +46,23 @@ class ListView extends React.Component {
...
-                {this.props.showSelectionController && <SelectionControllerContainer/>}
-                <ActionContainer
-                  key={`listAction-${child.id}`}
-                  definition={child}
-                  parent={props.parent}
-                  disabled={this.props.dataLoadingInProgress}
-                />
-              </ActionWrapper>
+              const content = [
+                ...this.props.showSelectionController
```
rzueger, line 50: there is `const props = this.props` defined on line 36. many usages over `this.props` in this method not necessary.

PR #942, opened 2020-05-01
feat(app-extensions): Add form scope to action situation

rzueger:
The scope is required at least for the LabelReportAction legacy action.

Refs: TOCDEV-1825

### DIFF
--- a/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
+++ b/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
@@ -110,6 +110,9 @@ export function* initLegacyActionsEnv() {
 
 export const listSelector = state => state.list
 
+export const entityListSelector = state => state.entityList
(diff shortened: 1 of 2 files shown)


packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -160,6 +163,31 @@ export function* getSelection(selection) {
   return legacySelection
 }
 
+export function* getScope() {
+  const list = yield select(entityListSelector)
+  const detail = yield select(entityDetailSelector)
+
+  if (list && detail) {
```
MethodenMann, line 170: could this ever be the case? :)

rzueger: I don't hope so :) but it's not impossible and I don't like debugging bugs that occur because of problems like that :)

PR #962, opened 2020-05-19
Pr/disable caching

MethodenMann:


packages/tocco-util/src/nice/utils.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable no-undef */
```
rzueger, line 1: `eslint-disable` could be removed, if call on line 6 was `return window.app.getRunEnv()`?

PR #977, opened 2020-06-11
refactor(app-extensions): reduce z-index of modal

yilkur:
Refs: TOCDEV-1725

### DIFF
--- a/packages/app-extensions/src/notifier/components/StyledNotifier.js
+++ b/packages/app-extensions/src/notifier/components/StyledNotifier.js
@@ -198,7 +198,7 @@ const StyledNotifier = styled.div`
       }
 
       .rrt-confirm-holder {
-        z-index: 100000000;  // reset: react-redux-toastr (index.scss) - Stacking: Toaster > Modal > Blocker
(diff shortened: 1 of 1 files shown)


packages/app-extensions/src/notifier/components/StyledNotifier.js
```diff
@@ -198,7 +198,7 @@ const StyledNotifier = styled.div`
       }
 
       .rrt-confirm-holder {
-        z-index: 100000000;  // reset: react-redux-toastr (index.scss) - Stacking: Toaster > Modal > Blocker
+        z-index: 99999;  // reset: react-redux-toastr (index.scss) - Stacking: Toaster > Modal > Blocker
```
MethodenMann, line 201: Do we have some kind of definition when to use what number or is 9999 more or less random?

yilkur: The number **9999** is lower than **99999** of .flatpickr-calendar.open class of the flatpickr. So that the dialog of the flatpickr will always be on top.

PR #981, opened 2020-06-18
refactor(app-extensions): resolve visual overlapping of components

yilkur:
Refs: TOCDEV-2034


packages/app-extensions/src/notifier/modules/modalComponents/ModalDisplay.js
```diff
@@ -5,7 +5,7 @@ import ModalContent from './ModalContent'
 
 const ModalDisplay = props => {
   return (
-    <React.Fragment>
+    <>
```
MethodenMann, line 8:
```
const ModalDisplay = props => 
   props.modals.map(....

```

should work

PR #982, opened 2020-06-19
Pr/two factor connector styling

yilkur:


packages/two-factor-connector/src/components/Result/Result.js
```diff
@@ -1,15 +1,21 @@
...
+  margin: ${scale.space(0)};
+`
+
+const StyledParagraph = styled(StyledP)`
+  text-align: center;
 `
 
 const Result = ({secret}) => {
-  return <React.Fragment>
+  return <>
```
MethodenMann, line 18: you can omit the return statement

PR #983, opened 2020-06-19
Pr/fullscreen action selection url

rzueger:


packages/admin/src/routes/entities/components/ActionView/ActionView.js
```diff
@@ -1,11 +1,18 @@
 import React, {useEffect} from 'react'
 import PropTypes from 'prop-types'
-import queryString from 'query-string'
+import {selection as selectionUtil} from 'tocco-util'
 import {intlShape} from 'react-intl'
 
 import Action from '../Action'
 import {currentViewPropType} from '../../utils/propTypes'
 
+const getSelection = location => {
```
MethodenMann, line 9:
```
const getSelection = location =>
location.state && location.state.selection
 ?  location.state.selection
 : selectionUtil.queryStringToSelection(location.search)
```

i prefer ternary. but It just comes down to a matter of taste

packages/admin/src/routes/entities/components/EditView/EditView.js
```diff
@@ -40,8 +41,10 @@ const EditView = props => {
 
   const handleNavigateToAction = ({definition, selection}) => {
     const entityBaseUrl = goBack(props.match.url)
+    const searchString = selectionUtil.selectionToQueryString(selection)
```
MethodenMann, line 44: if you would name the var "search" you could use the property shorthand on line 47

packages/admin/src/routes/entities/components/ListView/ListView.js
```diff
@@ -30,8 +31,10 @@ const ListView = ({match, history, currentViewInfo, emitAction, persistViewInfo,
 
   const handleNavigateToAction = ({definition, selection}) => {
     const entityBaseUrl = goBack(match.url)
+    const searchString = selectionUtil.selectionToQueryString(selection)
```
MethodenMann, line 34: if you would name the var "search" you could use the property shorthand on line 37

PR #988, opened 2020-06-26
feat(entity-list): add new input property `tql`

rzueger:
The given TQL condition acts like a constriction that is applied
additionally to every query.

Refs: TOCDEV-2075


packages/admin/src/routes/entities/components/Action/actions/ShowOutputJobsAction.js
```diff
@@ -0,0 +1,29 @@
...
+
+  useEffect(() => {
+    // If we redirect to the Output_job list right away, the UI gets stuck between
+    // two views because of some race conditions. This `setTimeout` makes sure all
+    // running sagas have completed when we redirect to avoid messing up the state.
+    setTimeout(() => setDoRedirect(true), 0)
+  }, [])
+
+  return doRedirect
+    ? <Redirect to={{
```
MethodenMann, line 16: i would add the "push" prop

PR #990, opened 2020-07-01
refactor(app-extensions): add principals helper and mock data

MethodenMann:


packages/app-extensions/src/rest/helpers.spec.js
```diff
@@ -486,6 +486,27 @@ describe('app-extensions', () => {
             expect(result).to.eql(expectedResult)
           })
         })
+
+        describe('fetchPrincipal', () => {
+          test('should ', () => {
```
PadPalon, line 491: beschreibung anpassen

PR #995, opened 2020-07-06
Pr/delete preparations

MethodenMann:


packages/tocco-util/src/selection/selection.js
```diff
@@ -7,3 +11,31 @@ export const queryStringToSelection = queryString => {
...
+ * @param {function} fetchEntities - rest helper function
+ * @returns {object} Object containing entityName and an array of keys (limited to 100000)
+ */
+export function* getEntities(selection, fetchEntities) {
+  const {entityName} = selection
+  if (selection.type === selectionTypes.QUERY) {
+    const entities = yield call(fetchEntities, selection.entityName, {...selection.query, limit})
+
+    if (entities.length > limit) {
+      yield call(consoleLogger.logError, 'Selected records exceed limit of 100000')
```
rzueger, line 28: (could use the `limit` variable in the error message to reduce the risk of becoming outdated)

PR #999, opened 2020-07-07
feat(delete): add delete action

MethodenMann:
Refs: TOCDEV-1881
Changelog: Delete Action

### DIFF
--- a/packages/delete/src/components/Dialog/index.js
+++ b/packages/delete/src/components/Dialog/index.js
@@ -1,0 +1,3 @@
+import Dialog from './DialogContainer'
+
+export default Dialog
--- a/packages/delete/src/components/Dialog/index.js
+++ b/packages/delete/src/components/Dialog/index.js
@@ -1,3 +1,1 @@
-import Dialog from './DialogContainer'
-
-export default Dialog
+export {default} from './DialogContainer'
--- a/packages/delete/src/components/InfoPart/InfoPart.js
+++ b/packages/delete/src/components/InfoPart/InfoPart.js
@@ -3,27 +3,27 @@ import {Typography, Link} from 'tocco-ui'
 import PropTypes from 'prop-types'
 
 import LinkPopOver from './LinkPopOver'
-import {relatedPropTypes} from '../../utils/deleteRequestParser'
+import {relatedPropType} from '../../utils/deleteRequestParser'
 
-const InfoPart = ({entityName, entityModel, primaryPks, relatedEntities, maxCountLink}) => {
-  const primaryTql = 'IN(pk,' + primaryPks.join(',') + ')'
+const InfoPart = ({entityName, entityLabel, keys, relatedEntities, maxCountLink}) => {
+  const primaryTql = 'IN(pk,' + keys.join(',') + ')'
   return <Typography.Span>
     <Typography.B>
-      {entityName} (<Link href={`/e/${entityModel}/list?tql=${primaryTql}`} rel="noopener noreferrer" target="_blank">
-        {primaryPks.length}
+      {entityLabel} (<Link href={`/e/${entityName}/list?tql=${primaryTql}`} rel="noopener noreferrer" target="_blank">
+        {keys.length}
       </Link>)
     </Typography.B>
     {
       Object.keys(relatedEntities).length > 0
       && <>
         <span> / </span>
-        {Object.keys(relatedEntities).map(entityModel => {
-          const relatedEntity = relatedEntities[entityModel]
-          const tql = 'IN(pk,' + relatedEntity.pks.slice(0, maxCountLink).join(',') + ')'
-          const linkText = [...relatedEntity.pks, ...relatedEntity.pksOtherBu].length
+        {Object.keys(relatedEntities).map(entityName => {
+          const relatedEntity = relatedEntities[entityName]
+          const tql = 'IN(pk,' + relatedEntity.keys.slice(0, maxCountLink).join(',') + ')'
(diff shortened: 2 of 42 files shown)


packages/delete/src/components/InfoPart/InfoPart.js
```diff
@@ -0,0 +1,52 @@
...
+      </>
+    }
+
+  </Typography.Span>
+}
+
+InfoPart.propTypes = {
+  entityName: PropTypes.string,
+  entityModel: PropTypes.string,
+  primaryPks: PropTypes.arrayOf(PropTypes.string),
```
rzueger, line 47: `primaryPks` seems to be required

packages/delete/src/components/InfoPart/InfoPart.spec.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import {mount} from 'enzyme'
+import {Link} from 'tocco-ui'
+
+import InfoPart from './InfoPart'
+
+describe('delete', () => {
+  describe('components', () => {
+    describe('Dialog', () => {
```
rzueger, line 9: 'InfoPart'

packages/delete/src/main.js
```diff
@@ -0,0 +1,76 @@
...
+  const events = EXTERNAL_EVENTS.reduce((acc, event) => ({
+    ...acc,
+    ...(props[event] ? {[event]: props[event]} : {})
+  }), {})
+
+  return initApp('delete', props, events).component
+}
+
+DeleteApp.propTypes = {
+  selection: selection.propType
```
rzueger, line 73: I assume this should be required

packages/delete/src/utils/deleteRequestParser.js
```diff
@@ -0,0 +1,77 @@
+import _get from 'lodash/get'
+import PropTypes from 'prop-types'
+
+const deleteStatus = {
+  DELETABLE: 'DELETABLE',
+  DENIED_BY_CASCADE: 'DENIED_BY_CASCADE',
+  NO_DELETE_PERMISSION: 'NO_DELETE_PERMISSION'
+}
+
+const transformRelatedEntities = (relatedEntities, entityToDelete, deletable, currentBuId) => ({
```
rzueger, line 10: 🙈  is this tested well enough? :)

packages/delete/src/components/InfoPart/InfoPart.js
```diff
@@ -0,0 +1,52 @@
...
+        }).reduce((prev, curr) => [prev, ', ', curr])}
+      </>
+    }
+
+  </Typography.Span>
+}
+
+InfoPart.propTypes = {
+  entityName: PropTypes.string,
+  entityModel: PropTypes.string,
```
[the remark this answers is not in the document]
PadPalon: From the usage I assumed entityName was the localized name of the entityModel (It's displayed in the bold tag on line 12)

packages/tocco-util/src/intl/intl.js
```diff
@@ -76,11 +76,11 @@ export const loadTextResources = async(locale, modules) => {
 
     notLoadedModules.forEach(module => {
       const filtered = Object.keys(resources)
-        .filter(key => key.includes(module))
-        .reduce((obj, key) => {
-          obj[key] = resources[key]
-          return obj
-        }, {})
+        .filter(key => !!key.match(new RegExp('^client\\.' + module + '.*', 'g')))
```
PadPalon, line 79: !! = true

packages/delete/src/components/InfoPart/InfoPart.js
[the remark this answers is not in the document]
MethodenMann: yes i know. it's a bit of a mess
I think from now on forward we should use
entityName / name for technical name e.g. User
entityLabel / label for the localized name e.g. Person
entityModel / model should be reserved for the model object (that contains the name and so on)
And entityKey / key for primary keys and avoid "id" or "pk" as substitute

entityId ist an object containing entityName and entityKey

what do you think? I could write it down in read the docs. but i think this will trigger some major changes. e.g. the selection right now looks like this:

```
{
  type: 'ID',
  ids: [....]
}

```
but should be renamed to 

```
{
 type: 'KEY',
 keys: []
}
```

packages/delete/src/utils/deleteRequestParser.js
MethodenMann: i think so. I started writing many tests but ended up copy alot of test data and do duplicate stuff. So i decided to pack all special cases in this one test and compare the whole output.

packages/delete/src/components/Dialog/index.js
```diff
@@ -0,0 +1,3 @@
+import Dialog from './DialogContainer'
+
+export default Dialog
```
rzueger, line 3: or `export {default} from './DialogContainer'`

packages/delete/src/modules/delete/sagas.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+const deleteDialogResource = 'client/delete/dialog'
+
+export function* loadDialogInfo() {
+  const body = yield call(getDeleteBodyFromSelection)
+
+  const response = yield call(rest.requestSaga, deleteDialogResource, {method: 'POST', body})
+  const principal = yield call(rest.fetchPrincipal)
```
rzueger, line 27: could `requestSaga` and `fetchPrincipal` also be executed in parallel? seems like they don't depend on each other.

PR #1001, opened 2020-07-08
integrate input-edit into client

PadPalon:
- clean up of several parts of input-edit


packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -202,7 +202,7 @@ export default function* (definition, selection, parent, params, config) {
 
   const situation = {
     entityName: selection.entityName,
-    scope: yield call(getScope)
+    scope: definition.scope ? definition.scope : yield call(getScope)
```
MethodenMann, line 205: `scope: definition.scope || yield call(getScope)`

packages/app-extensions/src/actions/modules/sagas.js
```diff
@@ -19,12 +19,14 @@ export function* invokeAction(config, {payload}) {
     const actionHandler = actionHandlers[definition.actionType]
     const response = yield call(actionHandler, definition, selection, parent, params, config)
 
-    if (response.remoteEvents) {
-      yield all(response.remoteEvents.map(remoteEvent => put(remoteEvents.remoteEvent(remoteEvent))))
-    }
+    if (response) {
```
MethodenMann, line 22: i did something simular with the now merged delete action. you can apply this change.

packages/entity-browser/package.json
```diff
@@ -14,6 +14,7 @@
     "tocco-test-util": ">0.0.0",
     "tocco-theme": ">0.0.0",
     "tocco-ui": "0.1.22",
-    "tocco-util": "0.1.19"
+    "tocco-util": "0.1.19",
```
MethodenMann, line 17: i think we can introduce a "greater than" (>0.1.0). This way always the locale version is used

packages/input-edit/src/components/InputEdit/InputEdit.js
```diff
@@ -1,39 +1,88 @@
...
+    minSize={300}
+    resizerStyle={resizerStyle}
+    split="vertical">
+    <StyledSplitPanelWrapper key={'sidebar'}>
+      <StyledSplitPane defaultSize={300}
+        minSize={300}
+        resizerStyle={resizerStyle}
+        split="horizontal">
+        <StyledSplitPanelWrapper key={'search'}>
+          <Typography.H2>{intl.formatMessage({id: 'client.input-edit.sidebar.search'})}</Typography.H2>
```
MethodenMann, line 56: in this case you can use FormattedMessage component since you use localized text in jsx directly.
https://formatjs.io/docs/react-intl/components

packages/input-edit/src/components/InputEdit/InputEdit.js
```diff
@@ -1,39 +1,88 @@
...
-      <Panel.Footer>
-        <InputEditPagination/>
-      </Panel.Footer>
-    </Panel.Wrapper>
-  </FlexRow>
+  const actionDefinitions = useMemo(() => inputDataForm.children ? inputDataForm.children
+    .find(child => child.id === 'main-action-bar')
+    .children
+    .filter(child => child.componentType === 'action')
+    .map(child => {
```
MethodenMann, line 39: .map(child => ({...child, scope: 'detail'}))

packages/input-edit/src/components/InputEdit/InputEdit.js
```diff
@@ -1,39 +1,88 @@
...
+          <Typography.H2>{intl.formatMessage({id: 'client.input-edit.sidebar.search'})}</Typography.H2>
+          <InputEditSearch/>
+        </StyledSplitPanelWrapper>
+        <StyledSplitPanelWrapper key={'information'}>
+          <Typography.H2>{intl.formatMessage({id: 'client.input-edit.sidebar.information'})}</Typography.H2>
+          <InputEditInformation/>
+        </StyledSplitPanelWrapper>
+      </StyledSplitPane>
+    </StyledSplitPanelWrapper>
+    <StyledSplitPanelWrapper key={'table'}>
```
MethodenMann, line 65: key="table"

packages/input-edit/src/components/InputEdit/InputEdit.js
```diff
@@ -1,39 +1,88 @@
...
+      {actionDefinitions.map(definition =>
+        <actions.Action key={definition.id}
```
MethodenMann, line 67: new line after actions.Action

packages/input-edit/src/components/InputEditTable/InputEditTable.js
```diff
@@ -40,17 +40,58 @@ const InputEditTable = ({data, inputDataForm, inputEditForm, updateValue, sortin
 }
 
 const arrowKeyHandler = event => {
-  if (document.activeElement.tagName === 'INPUT' && (event.key === 'ArrowUp' || event.key === 'ArrowDown')) {
+  if (document.activeElement.tagName === 'INPUT'
+    && (
+      event.key === 'ArrowUp'
```
MethodenMann, line 45: could be rewritten to:

```
 ['ArrowDown', 'ArrowLeft', 'ArrowRight', 'ArrowUp'].includes(event.key)
```

packages/input-edit/src/main.js
```diff
@@ -10,10 +10,15 @@ import InputEdit from './components/InputEdit/InputEditContainer'
...
-const initApp = (id, input, events, publicPath) => {
+const initApp = (id, input, events = {}, publicPath) => {
   const content = <InputEdit/>
 
   const store = appFactory.createStore(reducers, sagas, input, packageName)
+  actionEmitter.addToStore(store)
+  actions.addToStore(store)
+  notifier.addToStore(store, false)
+  errorLogging.addToStore(store, false)
+  externalEvents.addToStore(store, events)
```
MethodenMann, line 21: externalEvents is not used yet? I think there should be more code that it works properly.
see:
https://github.com/tocco/tocco-client/blob/master/packages/entity-detail/src/main.js#L107

packages/input-edit/src/components/InputEditTable/InputEditTable.js
```diff
@@ -104,15 +146,37 @@ const DataCell = ({nodes, column}) => {
...
+    }
+  } else {
+    return current
+  }
+}
+
+const InputCell = ({id, nodes, column, updateValue}) => {
   const width = column.dataType === 'single-select'
     ? Math.max(...(column.options.options.map(option => option.display.length))) * 8 + 60
     : undefined
```
MethodenMann, line 174: `undefined` is usually an indicator that nothing was set. if you want it to be nothing intentionally i would use "null"

packages/input-edit/src/components/InputEditTable/InputEditTable.js
```diff
@@ -104,15 +146,37 @@ const DataCell = ({nodes, column}) => {
   </td>
 }
 
-const InputCell = ({index, nodes, column, updateValue}) => {
+const loadValue = (nodes, column) => {
+  const pathElements = column.children[0].path.split('.')
+  return resolvePath(nodes, pathElements)
+}
+
+const resolvePath = (current, pathElements) => {
```
MethodenMann, line 154: i would move this method in some kind of util file and write unit a unit tests. that will also help to kind of document what is happening here

packages/input-edit/src/main.js
PadPalon: When initially developing I got an error from the entity-browser when I didn't include this line and then opened my action. I couldn't reproduce it now though, so I'm just removing it again.

PR #1018, opened 2020-07-20
Pr/dnd table

MethodenMann:


packages/app-extensions/src/rest/helpers/preferences.js
```diff
@@ -0,0 +1,31 @@
...
+  yield call(requestSaga, `client/preferences/${path}`, {method: 'DELETE'})
+}
+
+/**
+ * To add or overwrite preferences.
+ *
+ * @param preferences {object} Key/Value pairs.
+ */
+export function* savePreferences(preferences) {
+  yield call(requestSaga, 'client/preferences', {method: 'PUT', body: {values: preferences}})
```
rzueger, line 30: this will update preferences only partially?

e.g. if there are the following preferences already saved:
- foo: "bar"
- dings: 5

and you send with your request:
- dings: 6
- newpref: "hello"

will the saved preferences after your request be {foo: "bar", dings: 6, newpref: "hello"} or will they only be {dings: 6, newpref: "hello"} (only the data you sent)?

if it's the first, the HTTP method PATCH should be used instead of PUT. if it's the latter, PUT is correct.

packages/entity-list/src/components/Table/Table.js
```diff
@@ -200,7 +221,9 @@ Table.propTypes = {
     reverseRelationName: PropTypes.string
   }),
   showLink: PropTypes.bool,
-  linkFactory: PropTypes.objectOf(PropTypes.func)
+  linkFactory: PropTypes.objectOf(PropTypes.func),
+  changePosition: PropTypes.func.isRequired,
+  positions: PropTypes.objectOf(PropTypes.number)
```
rzueger, line 226: `positions` actually required? (i guess there will be no errors if it's undefined, but would it make sense?)

packages/entity-list/src/components/Table/useDnD.js
```diff
@@ -0,0 +1,46 @@
+import {useState} from 'react'
+
+const initialState = {
+  currentlyDragging: null,
+
```
rzueger, line 5: (unnecessary blank line)

packages/entity-list/src/components/Table/useDnD.js
```diff
@@ -0,0 +1,46 @@
...
+  currentlyDragOver: null
+}
```
rzueger, line 7: (maybe the blank line could be inserted here:))

packages/entity-list/src/modules/preferences/reducer.spec.js
```diff
@@ -0,0 +1,34 @@
+import reducer from './'
+import * as actions from './actions'
+
+const EXPECTED_INITIAL_STATE = {
+  positions: null
+}
+
+describe('entity-list', () => {
+  describe('modules', () => {
+    describe('selectable', () => {
```
rzueger, line 10: `'preferences'`

packages/entity-list/src/modules/preferences/sagas.spec.js
```diff
@@ -0,0 +1,56 @@
...
+import {expectSaga} from 'redux-saga-test-plan'
+import * as matchers from 'redux-saga-test-plan/matchers'
+
+import {entityListSelector} from '../list/sagas'
+import * as actions from './actions'
+import rootSaga, * as sagas from './sagas'
+
+describe('entity-list', () => {
+  describe('modules', () => {
+    describe('selection', () => {
```
rzueger, line 12: `'preferences'`

packages/entity-list/src/util/preferences.spec.js
```diff
@@ -0,0 +1,66 @@
...
+            lastname: 2,
+            email: 3
+          }
+
+          expect(changePosition(positions, 'lastname', 'relGender')).to.eql(expectedResult)
+        })
+      })
+
+      describe('getPositionsPreferencesToSave', () => {
+        test('should ', () => {
```
rzueger, line 46: should what? 😄

packages/entity-list/src/components/Table/Table.js
MethodenMann: there will be an error in the useDnD hook.

packages/entity-list/src/components/Table/useDnD.js
MethodenMann: i moved the blank line :)

packages/entity-list/src/util/preferences.spec.js
MethodenMann: every time 😄

packages/app-extensions/src/rest/helpers/preferences.js
MethodenMann: ok we will change this to PATCH

packages/entity-list/src/components/Table/Table.js
rzueger: in which case? if `positions` is undefined? then I suppose we should append the `isRequired`?

PR #1019, opened 2020-07-20
refactor(entity-list): style table with props rather than id selectors

MethodenMann:


packages/entity-list/src/components/Table/Table.js
```diff
@@ -107,15 +107,18 @@ const Table = props => {
 
   return <StyledTableWrapper>
     <StretchingTableContainer>
-      <StyledTable ref={tableEl} columns={columns} resizingColumn={resizingColumn}>
+      <StyledTable ref={tableEl} columns={columns} >
```
yilkur, line 110: unnecessary space after {columns}

PR #1026, opened 2020-07-28
handle sorting preferences

PadPalon:


packages/entity-list/src/components/Table/NavigationCellHeader.js
```diff
@@ -0,0 +1,22 @@
...
+import {BallMenu, MenuItem} from 'tocco-ui'
+import {connect} from 'react-redux'
+import PropTypes from 'prop-types'
+import {FormattedMessage} from 'react-intl'
+
+import {resetSorting} from '../../modules/preferences/actions'
+
+const NavigationCellHeader = props =>
+  <BallMenu buttonProps={{icon: 'ellipsis-v'}}>
+    <MenuItem onClick={() => props.resetSorting()}><FormattedMessage id="client.entity-list.sorting.reset"/></MenuItem>
```
MethodenMann, line 11: onClick={props.resetSorting}

packages/entity-list/src/modules/preferences/sagas.js
```diff
@@ -34,9 +38,38 @@ export function* changePosition({payload}) {
...
   const positionPreferences = yield call(util.getPositionsPreferencesToSave, formName, newPositions)
   yield call(rest.savePreferences, positionPreferences)
 }
+
+export function* saveSorting() {
+  const {sorting: sortings} = yield select(listSagas.listSelector)
+  if (sortings.length > 0) {
+    const listState = yield select(listSagas.entityListSelector)
+    const formName = `${listState.formName}_list`
+    yield all([
```
MethodenMann, line 53: would  `yield call(rest.deleteUserPreferences, `${formName}.sorting*`)` work instead of two requests?

packages/entity-list/src/util/preferences.js
```diff
@@ -39,3 +39,42 @@ export const getPositionsPreferencesToSave = (formName, positions) =>
...
+    .reduce((acc, [key, value]) => {
+      if (!acc[key.index]) {
+        acc[key.index] = {}
+      }
+      acc[key.index][key.field] = value
+      return acc
+    }, [])
+}
+
+const transformSortingKey = keyMatch => {
```
MethodenMann, line 57: return statement can be omitted

packages/entity-list/src/util/preferences.js
```diff
@@ -39,3 +39,42 @@ export const getPositionsPreferencesToSave = (formName, positions) =>
...
+  return {
+    field: transformSortingField(keyMatch[1]),
+    index: keyMatch[2] ? parseInt(keyMatch[2]) : 0
+  }
+}
+
+function transformSortingField(field) {
```
MethodenMann, line 64: Any reason why this is a named function?
Anyhow, this is shorter:

`const transformSortingField = field => field === 'Direction' ? 'order' :  field.toLowerCase()`

packages/entity-list/src/util/preferences.js
```diff
@@ -39,3 +39,42 @@ export const getPositionsPreferencesToSave = (formName, positions) =>
       ...acc,
       [`${formName}.${key}.position`]: positions[key].toString()
     }), {})
+
+export const getSorting = preferences => {
+  const keyRegex = /^.*\.sorting(Field|Direction)\.?([1-9]*)$/
+  return Object.entries(preferences)
+    .filter(([key]) => keyRegex.test(key))
+    .map(([key, value]) => [transformSortingKey(key.match(keyRegex)), value])
+    .reduce((acc, [key, value]) => {
```
MethodenMann, line 48:
```
.reduce((acc, [key, value]) => ({
...acc,
[key.index]: {
  ...acc[key.index],
  [key.field] : value
}
}), []
```

should also work

packages/entity-list/src/modules/preferences/sagas.js
PadPalon: I was afraid that a potential positioning of a field called "sorting_<something>" would also be deleted, and since this doesn't support full Regex, I can't specify that it should check if it is the end of the preference key.
Think this is relevant?

packages/entity-list/src/util/preferences.js
PadPalon: Because my IDE still isn't configured to not automatically generate these when I extract code to a method :P

PR #1044, opened 2020-08-06
adjust input-edit after testing

PadPalon:


packages/input-edit/src/components/InputEdit/InputEdit.js
```diff
@@ -39,50 +51,56 @@ const InputEdit = ({entityKey, initializeTable, initializeSearch, initializeInfo
...
 
-  return <StyledSplitPane
-    defaultSize={300}
-    minSize={250}
-    resizerStyle={resizerStyle}
-    split="vertical">
-    <StyledSplitPanelWrapper key={'sidebar'}>
-      <StyledSplitPane defaultSize={300}
-        minSize={300}
+  if (valid === false) {
```
MethodenMann, line 55: return null?

packages/input-edit/src/components/InputEditTable/InputEditTable.js
```diff
@@ -156,17 +156,25 @@ const InputCell = ({id, nodes, column, updateValue}) => {
...
-        onChange: changedValue => updateValue(nodes.pk.value, column.id, changedValue),
-        onFocus: ({target}) => target.select()
-      }}/>
+    {
+      column.readonly
+        ? <FormattedValue
+          id={id}
+          type={column.dataType}
+          value={data ? (Object.prototype.hasOwnProperty.call(data, 'value') ? data.value : data) : null}
+          options={column.options}/>
```
MethodenMann, line 165: /> on new line

packages/tocco-ui/src/EditableValue/typeEditors/NumberEdit.js
```diff
@@ -16,20 +16,28 @@ export const calculateMaxValue = (prePointDigits, postPointDigits, maxValue) =>
   return maxValue
 }
 
-export const isAllowedValue = (prePointDigits, postPointDigits, minValue, maxValue) => ({floatValue}) => {
-  const calculatedMaxValue = calculateMaxValue(prePointDigits, postPointDigits, maxValue)
+const getPreDecimalPositions = number => Math.ceil(Math.log10(number + 1))
```
MethodenMann, line 19: change seems to work. 
but now when the minimum is 10 and someone enters "1" nothing indicates that this value is not set and therefore not saved unless its a mandatory field.
Maybe if the input is shorter than the minimum value, the input should be cleared on focus change.

packages/tocco-ui/src/FormattedValue/typeFormatters/NumberFormatter.js
```diff
@@ -10,14 +10,15 @@ const NumberFormatter = props => {
...
-        minimumFractionDigits={2}
+        minimumFractionDigits={props.options ? props.options.postPointDigits || 2 : 2}
       />
     </Typography.Span>
   )
 }
 
 NumberFormatter.propTypes = {
   value: PropTypes.number,
+  options: PropTypes.object,
```
MethodenMann, line 21: maybe specify the object with propTypes.shape

packages/input-edit/src/components/InputEdit/InputEdit.js
```diff
@@ -39,50 +51,57 @@ const InputEdit = ({entityKey, initializeTable, initializeSearch, initializeInfo
...
-    </StyledSplitPanelWrapper>
-  </StyledSplitPane>
+      : null}
+  </LoadMask>
 }
 
 InputEdit.propTypes = {
-  entityKey: PropTypes.string.isRequired,
+  selection: PropTypes.object.isRequired,
+  validation: PropTypes.object.isRequired,
```
rzueger, line 96: Maybe define shape for `validation`? I personally would use `PropType.object` only if you don't know how the given object is structured and if you really don't care (e.g. because you only pass it down to the child components). In my experience any developer that doesn't know a certain component is able to understand and use it much quicker if the prop types (especially the more complex ones) are defined properly.

The `valid` property of this prop seems to be required (as far as i've seen the rendering of the component doesn't fail but there won't be rendered a whole lot if there is also no `message`).

PR #1056, opened 2020-08-19
handle empty and invalid values better in NumberEdit

PadPalon:


packages/tocco-ui/src/EditableValue/typeEditors/NumberEdit.js
```diff
@@ -19,6 +19,10 @@ export const calculateMaxValue = (prePointDigits, postPointDigits, maxValue) =>
 const getPreDecimalPositions = number => Math.ceil(Math.log10(number + 1))
 
 const checkValueRange = (minValue, maxValue, value) => {
+  if (!value) {
```
rzueger, line 22: if the `value` is `0`, you also return true, because `0` coerces to `false`. Is that really what you want?

IMO if the value is `0` and the min value is `10`, we should return `false`, for example. So I think the value should be checked for `null` or `undefined`.

The same applies to the lines 26 and 30 further below. We should be able to set `maxValue` and `minValue` to `0` to enforce or exclude negative numbers.

Maybe some tests for this function would be useful. :)

PR #1068, opened 2020-09-04
feat(entity-list): Support `clickable` flag in table form model

rzueger:
By default, a row is clickable. If the `clickable` flag is set on the
table in form model, the rows aren't clickable anymore. However, if
`selectable` is true they still can be selected using the checkbox.

Refs: TOCDEV-2364

### DIFF
--- a/packages/entity-list/src/components/Table/StyledTable.js
+++ b/packages/entity-list/src/components/Table/StyledTable.js
@@ -64,7 +64,7 @@ export const PaginationContainer = styled.div`
 
 const selectionStyles = css`
   display: contents;
-  cursor: pointer;
+  ${({clickable}) => clickable && 'cursor: pointer;'}
 
   &.selected {
     > ${StyledTableCell} {
--- a/packages/entity-list/src/components/Table/Table.js
+++ b/packages/entity-list/src/components/Table/Table.js
@@ -80,7 +80,7 @@ const Table = props => {
       singleSelectHandler(entity.__key, true, true)
     } else if (e.metaKey || e.ctrlKey) {
       singleSelectHandler(entity.__key)
-    } else {
+    } else if (props.clickable !== false) {
       props.onRowClick(entity.__key)
     }
   }
@@ -158,6 +158,7 @@ const Table = props => {
                     key={`list-row-${entity.__key}`}
                     className={`selectableRow ${isSelected(entity.__key) && 'selected'}`}
                     onClick={trOnClick(entity)}
+                    clickable={props.clickable}
                     data-cy="list-row"
                   >
                     {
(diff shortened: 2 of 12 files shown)


packages/entity-list/src/components/Table/StyledTable.js
```diff
@@ -64,7 +64,7 @@ export const PaginationContainer = styled.div`
 
 const selectionStyles = css`
   display: contents;
-  cursor: pointer;
+  ${props => props.clickable && 'cursor: pointer;'}
```
MethodenMann, line 67: imo better looking with object destructuring: 
` ${({clickable}) =>clickable && 'cursor: pointer;'}`

packages/entity-list/src/components/Table/Table.js
```diff
@@ -80,7 +80,7 @@ const Table = props => {
       singleSelectHandler(entity.__key, true, true)
     } else if (e.metaKey || e.ctrlKey) {
       singleSelectHandler(entity.__key)
-    } else {
+    } else if (props.clickable !== false) {
```
MethodenMann, line 83: you have to rebase. i added a check for "onRowClick defined" recently

MethodenMann (dismissed):

packages/entity-list/src/components/Table/StyledTable.js
rzueger: you're right dake 👍

packages/entity-list/src/components/Table/Table.js
rzueger: this PR is for 2.26 unfortunately. in the master branch, this file looks quite different. i will resolve the conflicts and reapply this clickable check when merging the 2.26 branch into the master branch.

MethodenMann:
Maybe rebase this branch with nice-releases/226. Travis fails because of a problem luca already solved.
"Error: Can't resolve 'tocco-input-edit/src/main' "

In the package.json of admin package there should be an asterisk for the input-edit" version to force a local lookup (an not npm registry)

rzueger:
I'm not sure if I understand what you mean :)

If i check out the clean nice-releases/226 branch on my local machine I get the same error when executing `lerna bootstrap --concurrency=1 --mutex network`. I can also set `"tocco-input-edit": "*"` in the package.json of the admin package, but that doesn't help.

What's the problem Luca already solved? Does `lerna bootstrap` work on a clean 226 branch on your machine?

MethodenMann:
> I'm not sure if I understand what you mean :)
> 
> If i check out the clean nice-releases/226 branch on my local machine I get the same error when executing `lerna bootstrap --concurrency=1 --mutex network`. I can also set `"tocco-input-edit": "*"` in the package.json of the admin package, but that doesn't help.
> 
> What's the problem Luca already solved? Does `lerna bootstrap` work on a clean 226 branch on your machine?

ok luca solved it for the entity-browser package. i pushed a commit to nice-releases/226 to fix the travis run. you can rebase now :) 
https://github.com/tocco/tocco-client/commit/e979125afd3e3444c6921bd4495914e745e8df2b

rzueger:
messed up rebase, replaced with pr/clickable2

PR #1073, opened 2020-09-07
refactor(tocco-ui): style draggable th and kabob

yilkur:
Refs: TOCDEV-2351
Changelog:  Style draggable table headers and kabob menu

### DIFF
--- a/packages/tocco-ui/src/Table/StyledTable.js
+++ b/packages/tocco-ui/src/Table/StyledTable.js
@@ -22,8 +22,16 @@ export const StyledTableCell = styled.td`
   display: flex;
 `
 
-export const StyledTableHeaderCell = styled.th`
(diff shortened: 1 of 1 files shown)


packages/tocco-ui/src/Table/StyledTable.js
```diff
@@ -53,6 +61,15 @@ export const StyledTableHeaderCell = styled.th`
     }
     `
   }
+  &#header-cell-navigation-column {
```
MethodenMann, line 64: The navigation cell is created and maintained in the entity-list app. Introducing a id selector in the generic tocco-ui table will compromise our goal so separate these app. I think we have to extend the column definition. Can you think of a generic attribute for this column to style accordingly? 
i must have missed this in the last review

MethodenMann (approved):

PR #1086, opened 2020-09-15
fix(admin): Take `baseRoute` into account for resource scheduler

rzueger:
For the deployment on master.tocco.ch, the admin input prop `baseRoute`
(which is used as `basename` for the router) is set to `/beta`.
By using `history#createHref` the URL is prefixed with the basename
which was missing before -> the entities that should be opened when
the user clicks on an event in the scheduler couldn't be displayed
on master.tocco.ch because the URL was wrong.

Refs: TOCDEV-2347

### DIFF
--- a/packages/admin/src/routes/entities/components/Action/actions/ResourceScheduler.js
+++ b/packages/admin/src/routes/entities/components/Action/actions/ResourceScheduler.js
@@ -5,11 +5,12 @@ import {selection as selectionPropType} from 'tocco-util'
 
 import {goBack} from '../../../../../utils/routing'
 
-const ResourceScheduler = ({match, selection, actionProperties}) => {
+const ResourceScheduler = ({match, selection, actionProperties, history}) => {
   const entityBaseUrl = goBack(match.url, 2)
   return <ResourceSchedulerApp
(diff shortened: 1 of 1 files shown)


MethodenMann (approved):

packages/admin/src/routes/entities/components/Action/actions/ResourceScheduler.js
```diff
@@ -5,11 +5,12 @@ import {selection as selectionPropType} from 'tocco-util'
 
 import {goBack} from '../../../../../utils/routing'
 
-const ResourceScheduler = ({match, selection, actionProperties}) => {
+const ResourceScheduler = ({match, selection, actionProperties, history}) => {
   const entityBaseUrl = goBack(match.url, 2)
   return <ResourceSchedulerApp
     onEventClick={({model, key}) => {
-      window.open(`${entityBaseUrl}/${model}/${key}`, '_blank')
+      const url = history.createHref({pathname: `${entityBaseUrl}/${model}/${key}`})
```
MethodenMann, line 12: didnt know about createHref ... good to know 👍

MethodenMann (approved):

PR #1088, opened 2020-09-16
Pr/column preferences

PadPalon:


PadPalon:
I've made a draft out of this, since I'll implement the reset button in here as well

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
```diff
@@ -0,0 +1,53 @@
+import PropTypes from 'prop-types'
```
MethodenMann, line 1: To one commit you added the changelog "Hide all columns that were set to hidden in the preferences" imo what counts is that you added the dialog. No need for to much detail when it comes to the changelog. (Just a detail you dont need to change that)

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
```diff
@@ -0,0 +1,53 @@
...
+import React, {useState} from 'react'
```
MethodenMann, line 2: Have you created a issue for yilmaz to style this? I would recommend not merging this change until styling is done.

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
```diff
@@ -0,0 +1,53 @@
...
+  && {
+    vertical-align: top;
+  }
+`
+
+const ColumnPicker = ({columns, onOk, intl}) => {
+  const [selectedColumns, setSelectedColumns] = useState({})
+  const [searchTerm, setSearchTerm] = useState(null)
+  const items = columns
+    .filter(column => searchTerm === null || column.label.match(new RegExp(`${searchTerm}`, 'i')))
```
MethodenMann, line 23: unused template string in RegExp?

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
```diff
@@ -0,0 +1,53 @@
...
+    .map(column => <Typography.Li key={column.id}>
+      <StyledCheckbox
+        type={'checkbox'}
+        id={column.id}
+        checked={Object.prototype.hasOwnProperty.call(selectedColumns, column.id)
+          ? selectedColumns[column.id] : !column.hidden}
+        onChange={value => setSelectedColumns({...selectedColumns, [column.id]: value.target.checked})}/>
+      <Typography.Label for={column.id}>{column.label}</Typography.Label>
+    </Typography.Li>)
+  return <React.Fragment>
```
MethodenMann, line 33: if you like you can shorten <React.Fragment> with return <> ... </>

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
```diff
@@ -0,0 +1,53 @@
...
+const StyledCheckbox = styled('input')`
+  && {
+    vertical-align: top;
+  }
+`
+
+const ColumnPicker = ({columns, onOk, intl}) => {
+  const [selectedColumns, setSelectedColumns] = useState({})
+  const [searchTerm, setSearchTerm] = useState(null)
+  const items = columns
```
MethodenMann, line 22: you could create the "useMemo" hook. Then items would not be calculated on every rerender.

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
MethodenMann: a think a linebreak before the return statement would it make better readable

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
```diff
@@ -0,0 +1,53 @@
...
+      <StyledCheckbox
+        type={'checkbox'}
+        id={column.id}
+        checked={Object.prototype.hasOwnProperty.call(selectedColumns, column.id)
+          ? selectedColumns[column.id] : !column.hidden}
+        onChange={value => setSelectedColumns({...selectedColumns, [column.id]: value.target.checked})}/>
+      <Typography.Label for={column.id}>{column.label}</Typography.Label>
+    </Typography.Li>)
+  return <React.Fragment>
+    <SearchBox placeholder={intl.formatMessage({id: 'client.entity-list.preferences.columns.search'})}
```
MethodenMann, line 34: if a component does no fit on a single line i would introcude a linnebreak for each property

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
```diff
@@ -0,0 +1,53 @@
...
+      onSearch={setSearchTerm}/>
+    <StyledUl>{items}</StyledUl>
+    <Button onClick={() => onOk(selectedColumns)}>
+      {intl.formatMessage({id: 'client.entity-list.preferences.columns.button'})}
```
MethodenMann, line 38: i would name it client.entity-list.preferences.columns.okButton to clear things up in the text file (sorry havent thought of that when reviewing the nice2 change)

packages/entity-list/src/components/ColumnPicker/ColumnPickerContainer.js
```diff
@@ -0,0 +1,15 @@
+import {hot} from 'react-hot-loader/root'
+import {connect} from 'react-redux'
+import {injectIntl} from 'react-intl'
+
+import ColumnPicker from './ColumnPicker'
+
+const mapActionCreators = {
```
MethodenMann, line 7: could be removed

packages/entity-list/src/components/ListView/ListView.js
```diff
@@ -118,7 +119,8 @@ ListView.propTypes = {
   sorting: PropTypes.arrayOf(PropTypes.shape({
     field: PropTypes.string,
     order: PropTypes.string
-  }))
+  })),
+  columnPreferences: PropTypes.objectOf(PropTypes.bool)
```
MethodenMann, line 123: i would either rename this prop/state to columnDisplayPreferences (or something like that) oder change the structure like this:
```
{
  "colnameX": {"display": true},
  "colnameY": {"display": true},
}

```
otherwise the name is a bit confusing for an object of booleans.

packages/entity-list/src/modules/preferences/reducer.js
```diff
@@ -5,20 +5,31 @@ import * as actions from './actions'
 const resetSorting = state => {
   return {
```
MethodenMann, line 6: return statement could be omitted

packages/entity-list/src/modules/preferences/reducer.js
```diff
@@ -5,20 +5,31 @@ import * as actions from './actions'
...
     ...state,
-    sorting: []
+    sorting: initialState.sorting
+  }
+}
+
+const resetColumns = state => {
```
MethodenMann, line 12: return statement could be omitted

packages/entity-list/src/modules/preferences/sagas.js
```diff
@@ -81,3 +87,53 @@ export function* resetPreferences() {
   yield call(listSagas.setSorting)
   yield call(listSagas.reloadData)
 }
+
+export function* displayColumnModal() {
+  const {formDefinition} = yield select(listSagas.listSelector)
+  const {columns: preferencesColumns} = yield select(preferencesSelector)
+  const formColumns = formDefinition.children
```
MethodenMann, line 94: could be moved to util file:
https://github.com/tocco/tocco-client/blob/master/packages/entity-list/src/util/api/forms.js
And use "getTable" from there

packages/entity-list/src/modules/preferences/sagas.js
```diff
@@ -81,3 +87,53 @@ export function* resetPreferences() {
...
+    true
+  ))
+
+  yield saveColumnPreferences(answerChannel, preferencesColumns, formDefinition)
+}
+
+function* saveColumnPreferences(answerChannel, preferencesColumns, formDefinition) {
+  const selectedColumns = yield take(answerChannel)
+  yield put(setColumns({...preferencesColumns, ...selectedColumns}))
+  if (Object.entries(selectedColumns).some(([, value]) => value)) {
```
MethodenMann, line 125: this will not work perfectly. if you open the dialog to just remove a column it will not refresh (because nothing in selected Columns has the value true")

packages/tocco-ui/src/Typography/Misc.js
```diff
@@ -74,6 +75,13 @@ const Kbd = props =>
     title={props.breakWords ? undefined : getTextOfChildren(props.children)}
   >{props.children}</StyledKbd>
 
+const Label = props =>
+  <StyledLabel
+    breakWords={props.breakWords}
+    title={props.breakWords ? undefined : getTextOfChildren(props.children)}
+    htmlFor={props.for}
+  >{props.children}</StyledLabel>
```
MethodenMann, line 83: new line for </StyledLabel>

packages/entity-list/src/components/ColumnPicker/ColumnPicker.js
PadPalon: https://toccoag.atlassian.net/browse/TOCDEV-2437
Created one

packages/tocco-ui/src/Typography/Misc.js
PadPalon: Either there's some part of this comment missing or I just don't understand =) I copied the layout from the other elements in this file, lemme know if there's something that should change

packages/entity-list/src/modules/preferences/sagas.js
PadPalon: Thing is, if we only remove columns we don't need to refresh the whole list. Since the list knows about the preferences it will remove the column and re-render as soon as they are updated, without refreshing. The only point where we need to refresh the whole list and reload data is when we add a column to the list, since the data to display is then not already loaded.

I guess "listActions.refresh()" is not optimally named for what it does, but I wouldn't want to rename it at this point.

Thoughts?

packages/tocco-ui/src/Typography/Misc.js
MethodenMann: sure something's missing here. anyhow i thought of rewriting it to this:

```
const Label = props =>
  <StyledLabel
    breakWords={props.breakWords}
    title={props.breakWords ? undefined : getTextOfChildren(props.children)}
    htmlFor={props.for}
  >
      {props.children}
  </StyledLabel>

```

but you have a point there with the consistency in the file. better leave it.

MethodenMann (approved):

PR #1089, opened 2020-09-16
Pr/recaptcha v2

MethodenMann:


packages/login/src/main.js
```diff
@@ -92,9 +92,9 @@ const initPasswordUpdateApp = (id, input, events, publicPath, customTheme) => {
       setupFetchMocks(packageName, fetchMock)
     }
 
-    const app = initLoginApp('id', require('./dev/login_input.json'))
+    // const app = initLoginApp('id', require('./dev/login_input.json'))
     // uncomment to develop passwordUpdate App
```
PadPalon, line 96: Did you exchange which lines are commented out on purpose or is it just a leftover from development?

MethodenMann: nope that was not on purpose :)

PadPalon (approved):

PR #1097, opened 2020-09-18
Pr/safari bugs

yilkur:


packages/admin/src/routes/entities/components/Breadcrumbs/StyledBreadcrumbs.js
```diff
@@ -15,45 +15,36 @@ export const StyledBreadcumbs = styled.div`
   }
 `
 
+export const StyledBreadcrumbsGroup = styled.div`
```
MethodenMann, line 18: imo changing from a prop to a "last" selector is a step backwards. Less transparent and more error-prone.

Why is this even necessary? But at least we then should remove the props "active" completely

MethodenMann (approved):

PR #1119, opened 2020-10-06
refactor(entity-list): style column picker

yilkur:
Refs: TOCDEV-2437
Changelog: Style column picker dialog inside modal


MethodenMann:
Something is a little bit off:
image
Entity: User 
UI: Language EN

MethodenMann:
And: Should the OK button be primary (red) oder "raised"? Is there a rule?

PR #1121, opened 2020-10-06
refactor(tocco-ui): refactor preview and upload

yilkur:
Refs: TOCDEV-2420
Changelog: Reset css overwrites of Upload.js and refactor Preview.js

### DIFF
--- a/packages/tocco-ui/src/Upload/Upload.js
+++ b/packages/tocco-ui/src/Upload/Upload.js
@@ -40,9 +40,8 @@ export class Upload extends React.Component {
     } else {
       if (this.state.isUploading) {
         return <UploadProgress file={this.state.previewFile} text={props.textResources.uploading} {...props}/>
-      } else {
-        return <UploadInput {...props} text={props.textResources.upload} onDrop={this.onDrop.bind(this)}/>
       }
+      return <UploadInput {...props} onDrop={this.onDrop.bind(this)}/>
     }
   }
 
@@ -77,7 +76,6 @@ Upload.propTypes = {
    * Allows to overwrite texts of the component
    */
   textResources: PropTypes.shape({
-    upload: PropTypes.string,
     uploading: PropTypes.string,
     delete: PropTypes.string,
     download: PropTypes.string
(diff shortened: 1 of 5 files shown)


packages/tocco-ui/src/Upload/UploadInput.js
```diff
@@ -30,7 +30,6 @@ const UploadInput = props => {
           position={design.position.PREPEND}
         />
       </IconWrapper>
-      {props.text || 'drag and drop or click'}
```
MethodenMann, line 33: Is it possible to set this text as "title" attribute? otherwise we should delete it completely wich means here:
- stories
- tests
- form field factory
- remove text resource from nice2

packages/tocco-ui/src/Upload/Upload.js
```diff
@@ -9,47 +9,43 @@ import UploadInput from './UploadInput'
...
-    super(props)
-    this.state = {isUploading: false}
-  }
+export const Upload = props => {
+  const {
+    value,
+    textResources,
+    onUpload
+  } = props
+  const [preview, setPreview] = useState({isUploading: false})
```
MethodenMann, line 18: same, introduce a "setState" for isUploading and previewFile

packages/tocco-ui/src/Upload/Upload.js
```diff
@@ -9,47 +9,43 @@ import UploadInput from './UploadInput'
...
-  getContent(props) {
-    if (props.value && props.value.binaryLink) {
-      return <View {...props} deleteTitle={props.textResources.delete} downloadTitle={props.textResources.download} />
-    } else {
-      if (this.state.isUploading) {
-        return <UploadProgress file={this.state.previewFile} text={props.textResources.uploading} {...props}/>
-      } else {
-        return <UploadInput {...props} text={props.textResources.upload} onDrop={this.onDrop.bind(this)}/>
-      }
+  const getContent = props => {
```
MethodenMann, line 39: imo it doesn't make sense to pass props here. props is in the closure of getContent? Especially makes no sense since you change props.value to value. So you are not using the parameter anyways.

MethodenMann (approved):

PR #1122, opened 2020-10-06
Pr/hide readonly

MethodenMann:


packages/app-extensions/src/field/mappings.js
```diff
@@ -89,15 +88,19 @@ const readOnlyMapping = {
...
   'string': formattedValueFactory('string'),
   'text': formattedValueFactory('text'),
   'text-area': formattedValueFactory('text'),
+  'time': formattedValueFactory('time'),
+  'updatets': formattedValueFactory('datetime'),
   'upload': formattedValueFactory('document'),
   'url': formattedValueFactory('url'),
-  'uuid': formattedValueFactory('string')
+  'uuid': formattedValueFactory('string'),
+  'version': formattedValueFactory('string')
```
PadPalon, line 103: Did you add these changes on purpose? They're fine, just wondering

PadPalon (approved):

packages/app-extensions/src/field/mappings.js
MethodenMann: not really related to this change but yes, on purpose. i did some testing with the User_Test form and noticed that these types are not yet mapped.

PR #1123, opened 2020-10-06
refactor(app-extensions): use new displays endpoint everywhere

PadPalon:
Refs: TOCDEV-1567
Changelog: Remove all usages of old display endpoints and use 'entities/2.0/displays'

### DIFF
--- a/packages/app-extensions/src/rest/helpers.js
+++ b/packages/app-extensions/src/rest/helpers.js
@@ -2,6 +2,7 @@ import _reduce from 'lodash/reduce'
 import _isObject from 'lodash/isObject'
 import {call} from 'redux-saga/effects'
 import {cache} from 'tocco-util'
+import _get from 'lodash/get'
 
 import {requestSaga} from './rest'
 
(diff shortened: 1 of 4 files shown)


packages/app-extensions/src/rest/helpers.js
```diff
@@ -33,21 +34,10 @@ export function* fetchEntity(entityName, key, query, transformer = json => json)
...
-    method: 'GET'
-  }
-
-  const resp = yield call(requestSaga, `entities/2.0/${entityName}/${key}/display`, options)
-  const display = resp.body.display
-  cache.add('display', `${entityName}.${key}`, display)
-
-  return display
+export function* fetchDisplay(entityName, key, type) {
+  return _get(yield fetchDisplays({[entityName]: [key]}, type), `${entityName}.${key}`, null)
```
MethodenMann, line 40: unnecessary return

packages/app-extensions/src/rest/helpers.spec.js
```diff
@@ -159,7 +159,19 @@ describe('app-extensions', () => {
 
       describe('fetchDisplay', () => {
         test('should call fetch', () => {
-          const response = {display: 'Test'}
+          const response = {
```
MethodenMann, line 162: it would be nice to have a test that verifies that only non-cached displays are loaded and that verifies, that cached and loaded displays are combined in the answer

packages/app-extensions/src/rest/helpers.js
PadPalon: I don't think it is, because it's not an arrow function. At least it doesn't work when I remove it.

MethodenMann (approved):

PR #1126, opened 2020-10-08
refactor(tocco-ui): fix stated value error display

yilkur:
Refs: TOCDEV-2413
Changelog: 
- Remove stated value error animation, as it caused a wobble effect
- Add focused prop to styled stated value error to prevent evaluation on every key stroke

### DIFF
--- a/packages/tocco-ui/src/StatedValue/StyledStatedValue.js
+++ b/packages/tocco-ui/src/StatedValue/StyledStatedValue.js
@@ -1,4 +1,4 @@
-import styled, {css} from 'styled-components'
+import styled, {css, keyframes} from 'styled-components'
 
 import {
   colorizeBorder,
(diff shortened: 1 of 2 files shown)


packages/tocco-ui/src/StatedValue/StyledStatedValue.js
```diff
@@ -106,22 +106,18 @@ const StyledStatedValueDescription = styled.p`
...
-    }
   }
+
+  to {
+    opacity: 1;
+  }
+`
+
+const StyledStatedValueError = styled.div`
+  animation: ${fadeIn} 500ms ease-in-out both;
```
MethodenMann, line 120: will-change recommended?

MethodenMann (approved):

PR #1155, opened 2020-11-05
refactor(entity-detail): improve form error notifications

MethodenMann:
- Move error-box to save buttomn
- show validation errors in toaster

Refs: TOCDEV-2581, TOCDEV-2582
Changelog: Improve form error notification

### DIFF
--- a/packages/entity-detail/src/components/DetailForm/SaveButton.js
+++ b/packages/entity-detail/src/components/DetailForm/SaveButton.js
@@ -1,28 +1,61 @@
-import {Button} from 'tocco-ui'
+import {Button, Popover} from 'tocco-ui'
 import React from 'react'
 import {intlShape} from 'react-intl'
 import PropTypes from 'prop-types'
+import styled from 'styled-components'
 
 import modes from '../../util/modes'
+import ErrorBox from '../ErrorBox'
 
-const SaveButton = ({submitting, mode, intl}) => {
+const StyledButton = styled(Button)`
+  ${({hasErrors, theme}) => hasErrors && `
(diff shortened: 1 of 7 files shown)


packages/entity-detail/src/components/DetailForm/SaveButton.js
```diff
@@ -1,28 +1,61 @@
...
 import PropTypes from 'prop-types'
+import styled from 'styled-components'
 
 import modes from '../../util/modes'
+import ErrorBox from '../ErrorBox'
 
-const SaveButton = ({submitting, mode, intl}) => {
+const StyledButton = styled(Button)`
+  ${({hasErrors, theme}) => hasErrors && `
+   background-color: ${theme.colors.signal.warning.text};
```
rzueger, line 12: `text` color as background color? won't this mess things up when someone changes that color in the theme? I guess I wouldn't expect it to be used as background color somewhere.

packages/entity-detail/src/modules/entityDetail/sagas.spec.js
```diff
@@ -65,64 +67,76 @@ describe('entity-detail', () => {
...
+              .call(sagas.handleSubmitError, error)
+              .run()
           })
+
+          test('should call handleInvalidForm on invalid form', () =>
+            expectSaga(sagas.submitForm)
+              .provide([
+                [select(isValidSelector(FORM_ID)), false]
+              ])
+              .not.call(sagas.handleInvalidForm)
```
rzueger, line 113: I thought it **should** call `handleInvalidForm`?

rzueger (approved):

PR #1157, opened 2020-11-09
refactor(tocco-ui): prevent touch on disabled inputs

yilkur:
Refs: TOCDEV-2724
Changelog: Disable touch on mobile and adjust margin

### DIFF
--- a/packages/tocco-ui/src/EditableValue/StyledEditableValue.js
+++ b/packages/tocco-ui/src/EditableValue/StyledEditableValue.js
@@ -23,6 +23,7 @@ const StyledInputCss = css`
   background-color: transparent;
   border: 0;
   cursor: ${({immutable}) => immutable ? 'not-allowed' : 'default'};
+  pointer-events: ${({immutable}) => immutable ? 'none' : 'auto'};
   flex-grow: 1;
   min-height: 2.6rem;
   min-width: 0;
@@ -64,6 +65,11 @@ const StyledInputCss = css`
   &[type='week'] {
     appearance: none;
     display: flex;
+
+    /* reset margin on smaller screens */
+    ::-webkit-date-and-time-value {
+      ${({immutable}) => immutable && 'margin: unset;'}
+    }
   }
 `
 


packages/tocco-ui/src/EditableValue/StyledEditableValue.js
```diff
@@ -23,14 +23,15 @@ const StyledInputCss = css`
   background-color: transparent;
   border: 0;
   cursor: ${({immutable}) => immutable ? 'not-allowed' : 'default'};
+  pointer-events: ${({immutable}) => immutable ? 'none' : 'auto'};
```
MethodenMann, line 26: We still have at least one case in which the editable value should remain clickable: the value of a remotefield to open the record in a new tab.
But i tested it and the link still works. Why though?

packages/tocco-ui/src/EditableValue/StyledEditableValue.js
```diff
@@ -23,14 +23,15 @@ const StyledInputCss = css`
...
   flex-grow: 1;
   min-height: 2.6rem;
   min-width: 0;
   outline: 0;
   padding: 0;
   ${() => declareFont({
-    color: theme.color('text')
-  })}
+  color: theme.color('text')
+})}
```
MethodenMann, line 34: intentionally? looks a bit off now.

packages/tocco-ui/src/EditableValue/StyledEditableValue.js
yilkur: I thought the same thing. Theoretically the input field shouldn't be clickable since `pointer-events: none` will make the element completely unresponsive to a hover or click. I think it's because there is always another parent element wrapping around this css. I tested it on mobile and on disabled fields the touch is properly disabled.

packages/tocco-ui/src/EditableValue/StyledEditableValue.js
yilkur: Must have been the auto format feature...looks weird to the human eye now indeed.

PR #1167, opened 2020-11-17
feat(tocco-util): Allow caching of loaded async route component

rzueger:
If caching of a route should be enabled, a key for the cache can be
passed to `loadRoute` as the 4th argument.

In this case the already loaded component will be returned which might
eliminate unnecessary rerenders of the same route.

Refs: TOCDEV-2560


MethodenMann (approved):
Maye add a test?

PR #1169, opened 2020-11-18
refactor(resource-scheduler): refactor resource scheduler

yilkur:
Refs: TOCDEV-1467

### DIFF
--- a/packages/resource-scheduler/src/components/ResourceScheduler/ResourceScheduler.js
+++ b/packages/resource-scheduler/src/components/ResourceScheduler/ResourceScheduler.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, {useEffect} from 'react'
 import SplitPane from 'react-split-pane'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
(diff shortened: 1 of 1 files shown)


packages/resource-scheduler/src/components/ResourceScheduler/ResourceScheduler.js
```diff
@@ -21,37 +21,36 @@ const StyledSplitPane = styled(SplitPane)`
...
-
-  paneWidth = 325
+const ResourceScheduler = ({
+  initialize,
+  locale,
+  calendarTypes,
+  updateRequestedCalendars,
+  requestedCalendars
+}) => {
+  initialize()
```
MethodenMann, line 31: wrap in useEffect? otherwise it will be called on every rerender

packages/resource-scheduler/src/components/ResourceScheduler/ResourceScheduler.js
```diff
@@ -21,37 +21,36 @@ const StyledSplitPane = styled(SplitPane)`
...
+  const paneWidth = 325
```
MethodenMann, line 32: is this var only used once? Maybe just set it on line 36

MethodenMann (approved):

PR #1179, opened 2020-11-23
refactor: split caching into short and long term one

stewue:


packages/tocco-util/src/intl/intl.js
```diff
@@ -42,19 +42,33 @@ const fetchOptions = {
...
   if (cachedUserLocale) {
     return cachedUserLocale
   }
 
+  const locale = await loadUserLocale()
+  cache.addLongTerm('user', 'locale', locale)
+  return locale
+}
+
+export const checkUserLocale = async() => {
```
MethodenMann, line 56: same as above

packages/login/src/modules/sagas.spec.js
```diff
@@ -94,21 +94,40 @@ describe('login', () => {
...
-          expect(gen.next().value).to.deep.equal(call(cache.clear))
-          expect(gen.next().value).to.deep.equal(put(setPassword('')))
-          expect(gen.next().done).to.deep.equal(true)
+          expectSaga(sagas.handleSuccessfulLogin, {})
+            .provide([
+              [matchers.call.fn(intl.hasUserLocaleChanged), false],
+              [matchers.call.fn(rest.hasRevisionIdChanged), false]
+            ])
+            .put(externalEvents.fireExternalEvent('loginSuccess', {timeout: sagas.DEFAULT_TIMEOUT}))
+            .call(cache.clearShortTerm)
```
MethodenMann, line 116: you could also check this

`.not.call(call(cache.clearAll))`

MethodenMann (approved):

PR #1192, opened 2020-11-27
Pr/table button

yilkur:


packages/tocco-ui/src/Table/StyledTable.js
```diff
@@ -21,6 +22,15 @@ export const StyledTableCell = styled.td`
   align-content: center;
   max-height: 23px;
   box-sizing: content-box;
+
+  ${StyledButton} {
```
MethodenMann, line 26: I dont like this kind if styling too much.

Wouldn't it be better to style the usage of the button in the table?

Like this we introduce a dependency from the table to the button. I think the table should not care about its content when it comes to specific styling.

MethodenMann (approved):

PR #1195, opened 2020-11-30
fix(scheduler): improve querySelector with boundary

MethodenMann:
- this makes sure, if multiple calendars are in the dom (multiple nice tabs opened), the right one is selected

Refs: TOCDEV-2552
Changelog: Fix deselect all bug


PadPalon:
Spelling mistake in commit message (imprive -> improve) but otherwise good

PadPalon (approved):

PR #1201, opened 2020-12-03
Pr/gitlab ci

MethodenMann:


.gitlab-ci.yml
```diff
@@ -1,23 +1,24 @@
...
 stages:
+  - setup
   - quality
   - build
 
+setup:
+  stage: setup
+  script:
+    - yarn install --frozen-lockfile
+    - yarn lernabootstrap:ci
```
pgerber, line 21: I'm not sure the setup stage does what you want. Before you had that in a before_script section. The way it is setup now, it'll run `yarn install --frozen-lockfile` and `yarn lernabootstrap:ci` in a separate docker container rather then before everything else. Is this intended?

deployment/deploy_storybook_gitlab.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e # Exit with nonzero exit code if anything fails
+PAGES_BRANCH="gh-pages"
+
+echo "open ssl connection"
```
pgerber, line 5: Should probably be something like "setup ssh". No openssl used here and no connection opened.

deployment/deploy_storybook_gitlab.sh
```diff
@@ -0,0 +1,32 @@
...
+eval `ssh-agent -s`
+ssh-add <(echo "$GITHUB_DEPLOY_KEY" | base64 -d)
+mkdir -p ~/.ssh
+ssh-keyscan -t rsa github.com >> ~/.ssh/known_hosts
```
pgerber, line 9: I'd check in the key retrieved via ssh-keyscan.

deployment/deploy_storybook_gitlab.sh
```diff
@@ -0,0 +1,32 @@
...
+echo "Copy storybook to folder in repo"
+BRANCH_FOLDER_NAME="${CI_COMMIT_BRANCH//\//_}"
+echo "folder name: ${BRANCH_FOLDER_NAME}"
+rm -rf $PAGES_BRANCH/$BRANCH_FOLDER_NAME
+mkdir $PAGES_BRANCH/$BRANCH_FOLDER_NAME
+mv storybook/* $PAGES_BRANCH/$BRANCH_FOLDER_NAME
+cd $PAGES_BRANCH
+
+echo "Commit and push"
+git config user.name "Travis CI"
```
pgerber, line 25: "Travis CI"?

deployment/deploy_storybook_gitlab.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e # Exit with nonzero exit code if anything fails
```
pgerber, line 2: `set -eu` would also catch unset variables.

pgerber (changes requested):

deployment/deploy_storybook_gitlab.sh
MethodenMann: could you elaborate a little more please.

PR #1208, opened 2020-12-08
refactor(merge): merge v2 selection

stewue:
- add logic for radio button and checkbox
- add set target entity functionality
- call merge action endpoint

Refs: TOCDEV-2743

### DIFF
--- a/packages/merge/src/components/MergeTable/MergeTable.js
+++ b/packages/merge/src/components/MergeTable/MergeTable.js
@@ -1,88 +1,13 @@
 import React, {useMemo} from 'react'
+import {connect} from 'react-redux'
 import PropTypes from 'prop-types'
-import {Table, FormattedValue, Typography, Button} from 'tocco-ui'
-import {FormattedMessage} from 'react-intl'
+import {Table, Typography, Button} from 'tocco-ui'
+import {FormattedMessage, injectIntl} from 'react-intl'
 
 import {getColumnDefinition, getDataRows} from '../../util/sourceData'
 import sourceDataPropType from '../../util/sourceDataPropType'
-
-const ManyRelationsCheckBox = ({entityData}) => <>
-  <input type="checkbox"/>
-  <Typography.Span><FormattedMessage id="client.merge.allRelations"/> ({entityData.value.length})</Typography.Span>
-</>
-
-ManyRelationsCheckBox.propTypes = {
-  entityData: PropTypes.shape({
-    value: PropTypes.arrayOf(PropTypes.string)
-  })
-}
-
-const RelationsCheckBoxes = ({entityData}) => {
-  return entityData.value.map(value =>
-    <div key={`relation-checkbox-${value.key}`}>
-      <input type="checkbox"/>
-      <FormattedValue type="single-select" value={value}/>
-    </div>
-  )
-}
-
-RelationsCheckBoxes.propTypes = {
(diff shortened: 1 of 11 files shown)


packages/merge/src/components/MergeTable/MergeTable.js
```diff
@@ -95,14 +20,41 @@ LabelCellRenderer.propTypes = {
   allPaths: PropTypes.arrayOf(PropTypes.string)
 }
 
+const ColumnHeaderRenderer = ({label, entityKey, setTargetEntity, targetEntity}) => {
+  const isChecked = targetEntity === entityKey
+  const onChange = () => setTargetEntity(entityKey)
+
+  return <><input type="radio" name="column" onChange={onChange} checked={isChecked}/> {label}</>
```
MethodenMann, line 27: same here with clickable label

packages/merge/src/modules/merge/reducer.js
```diff
@@ -1,17 +1,122 @@
...
+      multiple[key] = {}
+      obj.value.map(e => e.key).forEach(relatedEntityKey => {
+        multiple[key][relatedEntityKey] = entityKey
+      })
+    } else {
+      single[key] = entityKey
+    }
+  }
+
+  state.sourceData.relations.filter(e => e.entityKey === entityKey && e.keys.length > 0).map(e => e.relationName)
```
MethodenMann, line 81: i would prefer direct assigns over this variable / loop approach:

```
  const multipleAll = state.sourceData.relations
    .filter(e => e.entityKey === entityKey && e.keys.length > 0)
    .reduce((acc, val) => ({
      ...acc,
      [val.relationName]: [entityKey]
    }), {})

```

packages/merge/src/modules/merge/reducer.js
```diff
@@ -1,17 +1,122 @@
...
+    .forEach(relationName => {
+      multipleAll[relationName] = [entityKey]
+    })
+
+  return (
```
MethodenMann, line 86: omit round bracket?

packages/merge/src/modules/merge/reducer.js
```diff
@@ -1,17 +1,122 @@
...
+  [actions.SET_SELECTED_MULTIPLE_ALL]: setSelectedMultipleAll,
+  [actions.SET_TARGET_ENTITY]: setTargetEntity
 }
 
 const initialState = {
   selection: null,
   sourceData: null,
-  mergeResponse: null
+  mergeResponse: null,
+  selected: {
```
MethodenMann, line 114: reducers would have been a bit simpler if you moved "selected" into a own module. but you can leave it like that.

packages/merge/src/modules/merge/sagas.js
```diff
@@ -28,14 +28,54 @@ export function* loadSourceData(selection) {
...
   yield put(actions.setSourceData(sourceData))
+  yield put(actions.setTargetEntity(sourceData.entities[0].key))
 }
 
 export function* executeMerge() {
-  const body = {} // TODO: Implement
+  const {selection, selected, sourceData} = yield select(mergeSelector)
+
+  const paths = {}
+  for (const [name, entityKey] of Object.entries(selected.single)) {
```
MethodenMann, line 38: i would move this into a separate function. The code would be better readable and better testable.

MethodenMann (approved):

PR #1214, opened 2020-12-11
refactor(tocco-ui): prevent click on disabled

yilkur:
Refs: TOCDEV-2895
Changelog: Prevent click on disabled date input field


packages/tocco-ui/src/EditableValue/typeEditors/DateAbstract.js
```diff
@@ -153,6 +155,7 @@ class DateAbstract extends React.Component {
 
         >
           <StyledDateAbstractInput
+            disabled={this.props.immutable && 'disabled'}
```
MethodenMann, line 158: `{...this.props.immutable ? {disabled: 'disabled} : {}}`

i think like this would be a little bit better. Since its more clear imo and omits the disabled attribute completely if not immutable.

MethodenMann (approved):

PR #1229, opened 2020-12-17
refactor(app-extensions): fix modal on mobile

yilkur:
Refs: TOCDEV-2628
Changelog: Prevent mobile keyboard from pushing the modal up

### DIFF
--- a/packages/app-extensions/src/notifier/modules/modalComponents/ModalContent.js
+++ b/packages/app-extensions/src/notifier/modules/modalComponents/ModalContent.js
@@ -1,4 +1,4 @@
-import React, {useState} from 'react'
+import React, {useState, useRef, useLayoutEffect} from 'react'
 import PropTypes from 'prop-types'
 
 import TitleMessage from '../../components/TitleMessage'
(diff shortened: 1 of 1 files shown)


packages/app-extensions/src/notifier/modules/modalComponents/ModalContent.js
```diff
@@ -18,15 +18,23 @@ const ModalContent = ({
...
   const [isClosing, setIsClosing] = useState(false)
+  const ref = useRef(false)
 
   const handleCloseClick = () => {
     setIsClosing(true)
     close(id)
   }
 
+  useLayoutEffect(() => {
+    if (ref) {
```
MethodenMann, line 29: if (ref.current) would make sure, that the element is initialized.

packages/app-extensions/src/notifier/modules/modalComponents/ModalContent.js
```diff
@@ -18,15 +18,23 @@ const ModalContent = ({
   component: Component
 }) => {
   const [isClosing, setIsClosing] = useState(false)
+  const ref = useRef(false)
```
MethodenMann, line 21: refs are usually initialized with null

MethodenMann (approved):

PR #1236, opened 2020-12-29
refactor(merge): merge v2 selection & summary

stewue:
- Add merge summary component
- Refactor merge relations and selection data structure for rest request
- Check if field is writable
- Fix issue with not existing display of entity

### DIFF
--- a/packages/merge/src/components/MergeSummary/MergeSummary.js
+++ b/packages/merge/src/components/MergeSummary/MergeSummary.js
@@ -38,10 +38,17 @@ const MergeSummary = ({mergeResponse}) => {
   </>
 }
 
+const entityWarningPropType = PropTypes.shape({
+  pk: PropTypes.string.isRequired,
+  entity: PropTypes.string.isRequired,
+  entityLabel: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired
+})
+
 MergeSummary.propTypes = {
   mergeResponse: PropTypes.shape({
-    notCopiedRelations: PropTypes.arrayOf(PropTypes.object),
-    notDeletedEntities: PropTypes.arrayOf(PropTypes.object),
+    notCopiedRelations: PropTypes.arrayOf(entityWarningPropType),
+    notDeletedEntities: PropTypes.arrayOf(entityWarningPropType),
     showPermissionMessage: PropTypes.bool.isRequired
   })
 }
--- a/packages/merge/src/components/MergeSummary/MergeSummary.js
+++ b/packages/merge/src/components/MergeSummary/MergeSummary.js
@@ -1,0 +1,49 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import {Typography} from 'tocco-ui'
+import {FormattedMessage} from 'react-intl'
+import _groupBy from 'lodash/groupBy'
+
+const MergeSummary = ({mergeResponse}) => {
+  const notCopiedRelations = Object.entries(_groupBy(mergeResponse.notCopiedRelations, e => e.entityLabel))
+    .map(([entityLabel, list]) => `${entityLabel} (${list.length})`).join(', ')
+
+  const notDeletedEntities = Object.entries(_groupBy(mergeResponse.notDeletedEntities, e => e.entityLabel))
+    .map(([entityLabel, list]) => `${entityLabel} (${list.length})`).join(', ')
+
+  return <>
+    <Typography.P>
+      <FormattedMessage id="client.merge.summary.success"/>
+      {
+        (mergeResponse.notCopiedRelations.length > 0 || mergeResponse.notDeletedEntities.length > 0)
+        && <FormattedMessage id="client.merge.summary.successHint"/>
+      }
+    </Typography.P>
+    {
+      mergeResponse.notCopiedRelations.length > 0
+      && <Typography.P>
+        <FormattedMessage id="client.merge.summary.notCopiedRelations"/>:<br/>{notCopiedRelations}
+      </Typography.P>
+    }
+    {
+      mergeResponse.notDeletedEntities.length > 0
+      && <Typography.P>
+        <FormattedMessage id="client.merge.summary.notDeletedEntities"/>:<br/>{notDeletedEntities}
+      </Typography.P>
+    }
+    {
+      mergeResponse.showPermissionMessage
+      && <Typography.P><FormattedMessage id="client.merge.summary.permission"/></Typography.P>
+    }
+  </>
+}
+
+MergeSummary.propTypes = {
+  mergeResponse: PropTypes.shape({
+    notCopiedRelations: PropTypes.arrayOf(PropTypes.object),
+    notDeletedEntities: PropTypes.arrayOf(PropTypes.object),
+    showPermissionMessage: PropTypes.bool.isRequired
+  })
+}
+
+export default MergeSummary
--- a/packages/merge/src/components/MergeSummary/MergeSummaryContainer.js
+++ b/packages/merge/src/components/MergeSummary/MergeSummaryContainer.js
@@ -1,0 +1,12 @@
+import {connect} from 'react-redux'
+import {injectIntl} from 'react-intl'
+
+import MergeSummary from './MergeSummary'
+
+const mapActionCreators = {}
+
+const mapStateToProps = (state, props) => ({
+  mergeResponse: state.merge.mergeResponse
+})
+
+export default connect(mapStateToProps, mapActionCreators)(injectIntl(MergeSummary))
--- a/packages/merge/src/components/MergeTable/MergeTable.js
+++ b/packages/merge/src/components/MergeTable/MergeTable.js
@@ -20,13 +20,14 @@ LabelCellRenderer.propTypes = {
   allPaths: PropTypes.arrayOf(PropTypes.string)
 }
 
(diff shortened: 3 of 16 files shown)


packages/merge/src/components/MergeSummary/MergeSummary.js
```diff
@@ -0,0 +1,46 @@
...
+import PropTypes from 'prop-types'
+import {Typography} from 'tocco-ui'
+import {intlShape} from 'react-intl'
+import _groupBy from 'lodash/groupBy'
+
+const MergeSummary = ({mergeResponse, intl}) => {
+  const msg = id => intl.formatMessage({id})
+
+  const notCopiedRelations = Object.entries(_groupBy(mergeResponse.body.notCopiedRelations, e => {
+    return e.entityLabel
```
MethodenMann, line 11: "return" statement could be ommited

packages/merge/src/components/MergeSummary/MergeSummary.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import {Typography} from 'tocco-ui'
+import {intlShape} from 'react-intl'
+import _groupBy from 'lodash/groupBy'
+
+const MergeSummary = ({mergeResponse, intl}) => {
+  const msg = id => intl.formatMessage({id})
+
+  const notCopiedRelations = Object.entries(_groupBy(mergeResponse.body.notCopiedRelations, e => {
```
MethodenMann, line 10: i wouldnt add "body" to the store but the object containing "notCopiedRelations" directly. Body is a http request thing and a component should know nothing about it

packages/merge/src/components/MergeSummary/MergeSummary.js
```diff
@@ -0,0 +1,46 @@
...
+    return e.entityLabel
+  })).map(([entityLabel, list]) => `${entityLabel} (${list.length})`).join(', ')
+
+  const notDeletedEntities = Object.entries(_groupBy(mergeResponse.body.notDeletedEntities, e => {
+    return e.entityLabel
+  })).map(([entityLabel, list]) => `${entityLabel} (${list.length})`).join(', ')
+
+  return <>
+    <Typography.P>
+      {msg('client.merge.summary.success')}
```
MethodenMann, line 20: If possible (and here it is), use the react componenten rather than the function call.

<FormattedMessage id="client.merge.summary.success"/>

packages/merge/src/components/MergeSummary/MergeSummary.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+    {
+      mergeResponse.body.showPermissionMessage
+      && <Typography.P>{msg('client.merge.summary.permission')}</Typography.P>
+    }
+  </>
+}
+
+MergeSummary.propTypes = {
+  mergeResponse: PropTypes.object,
```
MethodenMann, line 42: you can specify this object a little more:

```
mergeResponse: PropTypes.shape({
   notDeletedEntities: PropTypes.......
})
```

packages/merge/src/components/MergeTable/MergeTable.js
```diff
@@ -20,28 +20,33 @@ LabelCellRenderer.propTypes = {
   allPaths: PropTypes.arrayOf(PropTypes.string)
 }
 
-const ColumnHeaderRenderer = ({label, entityKey, setTargetEntity, targetEntity}) => {
+const ColumnHeaderRenderer = ({label, entityKey, setTargetEntity, targetEntity, mergeStrategyDisplay, intl}) => {
+  const msg = id => intl.formatMessage({id})
   const isChecked = targetEntity === entityKey
+  const status = isChecked ? msg('client.merge.strategy.keep') : mergeStrategyDisplay
```
MethodenMann, line 26: Here you can use` <FormattedMessage>` as well

packages/merge/src/components/MergeSummary/MergeSummaryContainer.js
```diff
@@ -0,0 +1,13 @@
...
+
+import MergeSummary from './MergeSummary'
+
+const mapActionCreators = {}
+
+const mapStateToProps = (state, props) => ({
+  mergeResponse: state.merge.mergeResponse
+})
+
+// eslint-disable-next-line no-undef
```
MethodenMann, line 12: eslint  ignore can be removed?

packages/merge/src/modules/merge/sagas.js
```diff
@@ -53,20 +66,13 @@ export function* getMergeBody() {
...
-    const isSelected = selected.multipleAll[relation.relationName].includes(relation.entityKey)
-    if (isSelected || relation.entityKey === selected.targetEntity) {
-      mergeRelations.push({
-        relationName: relation.relationName,
-        sourceKey: relation.entityKey,
-        selected: isSelected
-      })
-    }
-  })
+  const mergeRelations = _uniq(sourceData.relations.map(r => r.relationName)).map(relationName => ({
```
MethodenMann, line 69: 2 maps necessary? Couldn't you just remove the first map and use "r.relationName" in the second one?

packages/merge/src/modules/merge/sagas.spec.js
```diff
@@ -55,23 +68,36 @@ describe('merge', () => {
...
+              selection: expectedSelection,
               targetEntity: {
                 key: '1',
                 paths: {}
               },
               mergeRelations: []
             }
             const response = {
-              success: true
+              notCopiedRelations: [
```
MethodenMann, line 79: is there a test that this line 
https://github.com/tocco/tocco-client/pull/1236/files#diff-bedba0a7c6df678291a8c34ff4d281b65eb1442e06adc222aedfd9d001fd6d00R47
with false?

packages/merge/src/modules/merge/sagas.js
stewue: the first map is inside the _uniq method to remove duplicates. merging into one map is not possible

packages/merge/src/modules/merge/sagas.spec.js
stewue: in the test getMergeBody the callname has writable=false, so this line is already tested

packages/merge/src/components/MergeSummary/MergeSummary.js
```diff
@@ -0,0 +1,50 @@
...
+    {
+      mergeResponse.showPermissionMessage
+      && <Typography.P>{intl.formatMessage({id: 'client.merge.summary.permission'})}</Typography.P>
+    }
+  </>
+}
+
+MergeSummary.propTypes = {
+  mergeResponse: PropTypes.shape({
+    notCopiedRelations: PropTypes.arrayOf(PropTypes.object),
```
MethodenMann, line 43: Since we do not have type safety, it's important to have the prop types as specific possible. I would use ProTypes.shape and describe the attributes even further

packages/merge/src/components/MergeSummary/MergeSummary.js
```diff
@@ -0,0 +1,50 @@
...
+const MergeSummary = ({mergeResponse, intl}) => {
+  const notCopiedRelations = Object.entries(_groupBy(mergeResponse.notCopiedRelations, e => e.entityLabel))
+    .map(([entityLabel, list]) => `${entityLabel} (${list.length})`).join(', ')
+
+  const notDeletedEntities = Object.entries(_groupBy(mergeResponse.notDeletedEntities, e => e.entityLabel))
+    .map(([entityLabel, list]) => `${entityLabel} (${list.length})`).join(', ')
+
+  return <>
+    <Typography.P>
+      {intl.formatMessage({id: 'client.merge.summary.success'})}
```
MethodenMann, line 16: use react component. see https://formatjs.io/docs/react-intl/components

packages/merge/src/modules/merge/actions.js
```diff
@@ -64,9 +64,11 @@ export const executeMerge = () => ({
   type: EXECUTE_MERGE
 })
 
-export const setMergeResponse = mergeResponse => ({
+export const setMergeResponse = (notCopiedRelations, notDeletedEntities, showPermissionMessage) => ({
```
MethodenMann, line 67: Wouldn't it be simpler and cleaner to just pass a "mergeResponse" object like before?

This would have the advanced that only one parameter is used and no custom reducer is required.

"done" could be selected like this:

```
const mapStateToProps = (state, props) => ({
  mergeDone: !!state.merge.mergeResponse
})	})
```

But that required mergeResponse to be initialized with null.

packages/merge/src/modules/merge/sagas.js
MethodenMann: sorry ...my fault :)

packages/merge/src/modules/merge/sagas.js
```diff
@@ -76,13 +82,13 @@ export function* getMergeBody() {
...
+  const resource = 'merge/merge'
   const options = {
     method: 'POST',
     body: yield call(getMergeBody)
   }
 
   const response = yield call(rest.requestSaga, resource, options)
-
-  yield put(actions.setMergeResponse(response))
+  yield put(actions.setMergeResponse(response.body.notCopiedRelations, response.body.notDeletedEntities,
```
MethodenMann, line 92: with the change from above this would look like this: 

`yield put(actions.setMergeResponse(response.body))`

MethodenMann (approved):

PR #1296, opened 2021-02-02
handle forced two factor authentication

PadPalon:
- include two-factor-connector in login form when two-factor authentication is forced from backend
- make two-factor-connector work with fixed secret and without session


packages/two-factor-connector/README.md
```diff
@@ -14,4 +14,8 @@ React-registry name: `two-factor-connector`
 | Name                   | Mandatory | Description
 |------------------------|:---------:|-------------
 | `onSuccess`            |           | Callback if 2fa setup was successful.
+| `username`             |           | Username to use, will be loaded from principal if not available
+| `password`             |           | Password to use, only necessary if forced
+| `secret`               |           | Secret to use for activation
```
MethodenMann, line 19: "if not set, a secret will be loaded from the server."

packages/two-factor-connector/src/modules/sagas.js
```diff
@@ -9,35 +9,73 @@ const twoFactorActiveState = '1'
...
   yield put(actions.setSetupSuccessful(codeVerifyResponse.status === 204))
+}
+
+export function* verifyCode({payload: {userCode}}) {
+  const username = yield select(usernameSelector)
+  const {secret} = yield select(secretSelector)
+  const {password, forced} = yield select(inputSelector)
+
+  if (forced) {
+    yield call(loginAndActivateTwoFactor, secret, userCode, username, password)
```
MethodenMann, line 66: Would be nice to verify this code with the same endpoint. If not possible a rename of this function and maybe a comment in the code will be helpful.

MethodenMann (approved):

packages/login/src/modules/sagas.spec.js
```diff
@@ -63,6 +64,23 @@ describe('login', () => {
           expect(gen.next().done).to.eql(true)
         })
 
+        test('should handle two step activation response', () => {
```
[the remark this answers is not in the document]
PadPalon: Usually do, but since the rest of the tests here were written this way, I thought I'd just copy those for now

PR #1312, opened 2021-02-12
refactor(tocco-ui): make panel clickable

yilkur:
Refs: TOCDEV-3193
Changelog: Make panel header clickable so it can also expand the panel

### DIFF
--- a/packages/tocco-ui/src/Panel/PanelHeaderFooter.js
+++ b/packages/tocco-ui/src/Panel/PanelHeaderFooter.js
@@ -8,37 +8,33 @@ import Ball from '../Ball'
  * <Panel.Header/> and <Panel.Footer/> contain by default a button to toggle the visibility state of <Panel.Body>.
  * Header and footer can contain any content. If both are displayed is up to the implementer.
  */
-class PanelHeaderFooter extends React.PureComponent {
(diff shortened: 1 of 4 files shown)


packages/tocco-ui/src/Panel/PanelHeaderFooter.js
```diff
@@ -78,4 +76,4 @@ PanelHeaderFooter.propTypes = {
   })
 }
 
-export default PanelHeaderFooter
+export default memo(PanelHeaderFooter)
```
MethodenMann, line 79: I think memo is not really necessary here?

packages/tocco-ui/src/Panel/PanelHeaderFooter.js
```diff
@@ -8,21 +8,20 @@ import Ball from '../Ball'
...
+const PanelHeaderFooter = ({
+  children,
+  isFramed,
+  isOpen,
+  isToggleable,
+  showToggler,
+  toggleOpenState,
+  options
+}) => {
+  return (
```
MethodenMann, line 20: return could be omitted

MethodenMann (approved):

PR #1345, opened 2021-03-02
handle information exceptions

PadPalon:
Refs: TOCDEV-3097
Changelog: Handle information exceptions from the server in forms, delete action and a general fallback.

### DIFF
--- a/packages/app-extensions/src/rest/rest.js
+++ b/packages/app-extensions/src/rest/rest.js
@@ -1,8 +1,10 @@
-import {call} from 'redux-saga/effects'
+import {call, put} from 'redux-saga/effects'
 import {v4 as uuid} from 'uuid'
 
 import {sendRequest} from './request'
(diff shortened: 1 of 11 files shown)


packages/app-extensions/src/rest/rest.js
```diff
@@ -42,15 +44,30 @@ export const setNullBusinessUnit = value => {
...
+      sendRequest,
+      requestData.url,
+      requestData.options,
+      options.acceptedErrorCodes,
+      options.acceptedStatusCodes
+    )
+    response = yield call(handleClientQuestion, response, requestData, options)
+    return response
+  } catch (error) {
+    if (error instanceof InformationError) {
```
MethodenMann, line 58: As far as i understand this is the fallback handling. Would move it to a separate function and set a function name that suggest, that its a fallback.

MethodenMann (approved):

PR #1371, opened 2021-03-11
feat(two-factor-connector): Add external event `onResize`

rzueger:
This event is fired whenever the size of the screen (resp. its height)
changes substantially (similar to the `resize` event of the login
and password update app).

This event, for instance, is useful if the app is rendered in a popup
window which should be centered whenever the height of the app changes.

Refs: TOCDEV-3048
Changelog: Introduce onResize external event


rzueger:
this is a cherry pick from 2.28

PR #1381, opened 2021-03-16
feat: add different form for root level in dms

stewue:
Refs: TOCDEV-3175
Changelog: add different form for root level in dms

### DIFF
--- a/packages/entity-list/src/modules/list/sagas.spec.js
+++ b/packages/entity-list/src/modules/list/sagas.spec.js
@@ -71,8 +71,8 @@ describe('entity-list', () => {
             expect(gen.next({entityName, formName}).value).to.eql(select(sagas.listSelector))
             const nextValue = gen.next({formDefinition, entityModel, initialized}).value
             expect(nextValue).to.eql(all([
-              call(sagas.loadEntityModel, entityName, entityModel),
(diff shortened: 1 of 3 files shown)


packages/entity-list/src/modules/list/sagas.spec.js
```diff
@@ -114,6 +114,20 @@ describe('entity-list', () => {
...
+          test('should set parent, reload entity model, form definition and data', () => {
+            return expectSaga(sagas.setParent)
+              .provide([
+                [select(sagas.entityListSelector), {formName: 'UserTest', entityName: 'Docs_list_item'}],
+                [select(sagas.listSelector), generateState({}, 1)],
+                [matchers.call.fn(sagas.loadEntityModel), {}],
+                [matchers.call.fn(sagas.loadFormDefinition), {}],
+                [matchers.call.fn(sagas.reloadData), {}]
+              ])
+              .run()
```
rzueger, line 127: actually this test doesn't verify anything :)

the providers in `provide` are only mocks, this test would also pass if the `setParent` function was completely empty, right?

actual assertions should be added after the `provide()` before `run()` is called

rzueger (approved):

PR #1399, opened 2021-03-22
refactor(tocco-ui): add icon to buttons

yilkur:
Refs: TOCDEV-3414
Changelog: Add icons to action buttons

### DIFF
--- a/packages/tocco-ui/src/Icon/mapping.json
+++ b/packages/tocco-ui/src/Icon/mapping.json
@@ -337,6 +337,41 @@
       "trash"
     ]
   },
+  "comments": {
+    "id": "fal, comments",
+    "lib": "fa",
+    "tags": [
+      "comments"
+    ]
+  },
+  "cog-light": {
+    "id": "fal, cog",
+    "lib": "fa",
+    "tags": [
+      "cog"
+    ]
+  },
+  "file-export-light": {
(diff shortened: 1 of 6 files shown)


packages/tocco-ui/src/Menu/ButtonMenu.js
```diff
@@ -84,7 +105,11 @@ ButtonMenu.propTypes = {
   /**
    * cypress selector string
    */
-  'data-cy': PropTypes.string
+  'data-cy': PropTypes.string,
+  /**
+   * Icon Type to be shown on the button
+   */
+  'iconType': PropTypes.string
```
MethodenMann, line 112: please move the icon mapping outside this tocco ui component and just add a icon property

packages/tocco-ui/src/Icon/mapping.json
```diff
@@ -337,6 +337,41 @@
       "trash"
     ]
   },
+  "comments": {
```
MethodenMann, line 340: i think its important to support a clear icon set. right now it seems that every other icons is getting a light version as well. Can't we use one variation everywhere? Or dont you think a clear icon set is needed for clarity and consistency?

MethodenMann (changes requested):

PR #1409, opened 2021-03-24
allow simple action to download files

PadPalon:
Changelog: simpleAction can now handle download links returned from the backend

### DIFF
--- a/packages/app-extensions/src/actions/modules/actionHandlers/simpleAction.js
+++ b/packages/app-extensions/src/actions/modules/actionHandlers/simpleAction.js
@@ -1,4 +1,5 @@
 import {put, call} from 'redux-saga/effects'
+import {download} from 'tocco-util'
 
 import errorLogging from '../../../errorLogging'
 import rest from '../../../rest'
@@ -46,6 +47,18 @@ export function* invokeRequest(definition, selection, parent, params) {
         notifier.info('error', 'client.component.actions.validationError', validationErrorCompact(response.body.errors),
           'exclamation')
       )
+    } else if (response.body && response.body.params.downloadUrl) {
+      const fileResponse = yield call(rest.requestBytesSaga, response.body.params.downloadUrl, {
+        method: 'POST',
+        body: {
+          entity: selection.entityName,
+          selection,
+          parent,
+          ...params,
+          formProperties: definition.properties
+        }
+      })
+      yield call(download.downloadReadableStream, fileResponse.body, response.body.params.filename)
     } else {
       const success = response.body.success === true
       const type = success ? 'success' : 'warning'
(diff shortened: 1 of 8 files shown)


packages/app-extensions/src/rest/rest.js
```diff
@@ -74,6 +74,22 @@ export function* requestSaga(resource, options = {}) {
   }
 }
 
+export function* requestBytesSaga(resource, options = {}) {
```
MethodenMann, line 77: maybe add a description here

packages/tocco-util/src/download/download.js
```diff
@@ -13,6 +13,22 @@ export const downloadUrl = (url, fileName) => {
   }
 }
 
+export const downloadReadableStream = (readableStream, fileName) => {
```
MethodenMann, line 16: maybe add a short description here as well

packages/app-extensions/src/actions/modules/actionHandlers/simpleAction.js
```diff
@@ -46,6 +47,18 @@ export function* invokeRequest(definition, selection, parent, params) {
...
+          entity: selection.entityName,
+          selection,
+          parent,
+          ...params,
+          formProperties: definition.properties
+        }
+      })
+      yield call(download.downloadReadableStream, fileResponse.body, response.body.params.filename)
     } else {
       const success = response.body.success === true
```
MethodenMann, line 63: No success message needed here? or do you think a file download is sufficient?

packages/app-extensions/src/rest/request.js
```diff
@@ -41,3 +54,9 @@ export function sendRequest(url, options, acceptedErrorCodes = [], acceptedStatu
     .then(response => (extractBody(response)))
     .then(response => (handleError(response, acceptedErrorCodes, acceptedStatusCodes)))
 }
+
+export function sendByteRequest(url, options, acceptedErrorCodes, acceptedStatusCodes) {
+  return fetch(url, options)
+    .then(response => (extractBlobBody(response)))
```
MethodenMann, line 60: bracket not needed

`.then(response => extractBlobBody(response))`

packages/app-extensions/src/rest/request.js
```diff
@@ -30,6 +30,19 @@ const extractBody = response => {
     })
 }
 
+const extractBlobBody = response => {
```
MethodenMann, line 33: since this is almost a copy of extractBody: i think it should be possible to move the code to one function an pass the function name (blob()/json()) as argument

packages/app-extensions/src/actions/modules/actionHandlers/simpleAction.js
PadPalon: In my opinion the download is enough, a message that the file has started downloading doesn't seem to add any value. But I don't mind either way.

Of course, assuming this gets ported to a more general way of downloading files from actions with the whole websocket / notification business this will be obsolete anyway.

MethodenMann (approved):

PR #1416, opened 2021-03-30
Pr/dms move

stewue:


packages/docs-browser/src/components/Action/actions/Move.js
```diff
@@ -0,0 +1,134 @@
...
+    margin-right: 0;
+  }
+`
+
+const MoveAction = ({selection, onSuccess, onError, intl, context, close, moveElements, isWaiting, isDone}) => {
+  const initialNode = getNode(context.history.location.pathname)
+  const initialParent = {model: 'Docs_list_item', key: `${initialNode.model}/${initialNode.key}`}
+  const [parent, setParent] = useState(initialParent)
+
+  useEffect(() => {
```
MethodenMann, line 28: i would move such code in a saga that is triggered on init. I think this component is the wrong place for a input validation

packages/docs-browser/src/components/Action/actions/Move.js
```diff
@@ -0,0 +1,134 @@
...
+    if (selection.type === 'QUERY') {
+      onError({
+        message: intl.formatMessage({id: 'client.actions.dms-move.explicitSelection.message'})
+      })
+    }
+  }, [])
+
+  useEffect(() => {
```
MethodenMann, line 36: also this. why not call all the callback directly in the sage instead of setting a state attribute and trigger the callback in the react component. imo a components main concern should be the visualization of data.

packages/docs-browser/src/components/DocsView/DocsView.js
```diff
@@ -90,16 +105,18 @@ const DocsView = props => {
...
+        formName={formName === null ? getFormName(parent, keys) : formName}
+        limit={limit || 25}
         onRowClick={handleRowClick}
         searchFormPosition="left"
-        searchFormType="admin"
+        searchFormType={searchFormType || 'admin'}
         parent={parent}
         onSearchChange={onSearchChange}
-        store={viewPersistor.viewInfoSelector(storeKey).store}
+        store={disableViewPersistor ? undefined : viewPersistor.viewInfoSelector(storeKey).store}
```
MethodenMann, line 115: maybe `!disableViewPersistor && viewPersistor.viewInfoSelector(storeKey).store`

packages/docs-browser/src/components/Action/actions/Move.js
```diff
@@ -0,0 +1,106 @@
...
+  moveElements
+}
+
+const mapStateToProps = state => ({
+  isWaiting: state.docs.move.isWaiting
+})
+
+const MoveActionContainer = connect(mapStateToProps, mapActionCreators)(injectIntl(MoveAction))
+
+export default injectIntl(props => <MoveActionContainer {...props} />)
```
MethodenMann, line 106: to be a bit consistent to other code i would create the MoveActionContainer in a separate file (same folder) and avoid two injectIntls.

packages/docs-browser/src/modules/list/index.js
```diff
@@ -0,0 +1,12 @@
+import {fork, all} from 'redux-saga/effects'
+
+import moduleSagas from './sagas'
+import reducer from './reducer'
+
+export function* sagas() {
+  yield all([
```
MethodenMann, line 7: i think `export {sagas} `should be sufficient?

packages/tocco-util/src/validation/validation.js
```diff
@@ -0,0 +1,16 @@
+// extracts the first error message of a validation failed response object
```
MethodenMann, line 1: There is a common format for function comments. See:
https://en.wikipedia.org/wiki/JSDoc

MethodenMann (approved):

PR #1417, opened 2021-03-31
chore: add automatic releasing

stewue:
- if the flag "--auto" is passed questions will be automatically answered
- suggest next version
- add script to release all packages where the package.json has a property "releasePackage" which is true and the changelog is not empty

### DIFF
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -44,9 +44,9 @@ function setGitVars() {
 
 
 function gitPush() {
+  git push --tags
   git checkout -b $1
   git push
-  git push --tags
   git checkout ${current_branch}
   echo "${color_green}Commits and tags pushed!${color_reset}"
 }
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -29,11 +29,26 @@ function setColors(){
 
 function setGitVars() {
   current_branch=$(git rev-parse --abbrev-ref HEAD)
+  remote_branch=$(git rev-parse --abbrev-ref --symbolic-full-name @{u} | cut -d '/' -f2-)
   last_release_tag=$(git describe --tags --match 'tocco-'${package}'@*' --abbrev=0 ${current_branch})
   last_version=$(echo ${last_release_tag} | awk -F '@' '{print $2}')
 
+  if [[ $remote_branch != "master" && $remote_branch != nice-releases/* ]]; then
+    echo "Running release script is not allowed if remote branch is not master or nice-releases/*"
+    exit
+  fi
+
   greps=$(getDevDependenciesGreps)
   changelog=$(git log --pretty='%b' "${last_release_tag}"..HEAD --grep="${package}" ${greps} --reverse | grep -E '^Changelog:' | awk '{gsub("Changelog:", "-", $0); print}')
+}
+
+
+function gitPush() {
+  git checkout -b $1
+  git push
+  git push --tags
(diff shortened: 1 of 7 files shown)


packages/admin/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.4.30",
   "description": "Alpha version of new tocco admin.",
   "main": "dist/index.js",
+  "releasePackage": true,
```
MethodenMann, line 6: i think you can work with the private attribute here:
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private

scripts/release-all-packages.sh
```diff
@@ -0,0 +1,19 @@
...
+for dir in packages/*/ ; do
+  if [[ $(json -f "${dir}package.json" releasePackage) == true ]]; then
+    releasePackages+=("$(cut -d'/' -f2 <<<${dir})")
+  fi
+done
+
+for package in "${releasePackages[@]}"
+do
+	echo "Start releasing package $package"
+	yarn release-package $package --auto
```
MethodenMann, line 15: this will run "yarn setup" for every package since its part of "release-package". maybe this could be fixed by separating this two calls

scripts/release-all-packages.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+source ./scripts/helpers.sh
+
+releasePackages=()
```
MethodenMann, line 5: i know im not a good example but maybe in bash files we could add some describing comments here and there since its not JS and we dont have any unit tests. 

like

// Gather all packages with releasePackage equals true in package.json

scripts/release-package.sh
```diff
@@ -15,13 +24,25 @@ else
...
+
 echo "---------------------"
 echo "info latest release tag: ${color_blue}${last_release_tag}${color_reset}"
 echo "info latest version: ${color_blue}${last_version}${color_reset}"
 echo "release tag: ${color_blue}${release_tag}${color_reset}"
 echo -e  "Generated changelog:\n${color_blue}${changelog}${color_reset}"
 
-read -p "question New version: : " new_version
+if [[ $auto = false ]]; then
+  read -p "question New version [${next_version}] : " new_version
```
MethodenMann, line 39: I think this question could be a bit confusing. I would write that the value inside the square brackets is the suggested version. Also it would be nice to confirm this version without rewriting it as answer. But overwriting should still be possible.

Also the word "question" can be removed. I think this was added by me by mistake.

scripts/helpers.sh
```diff
@@ -36,14 +36,44 @@ function setGitVars() {
...
+
 function setCurrentReleaseTag() {
-  nice_version=$(head -n 1 nice-current-version.txt)
-  release_tag="nice${nice_version//[(.0).]/}"
+  setNiceVersion
+  release_tag="nice${nice_version}"
+}
+
+function setNextVersion() {
+  if [[ $(git rev-parse --abbrev-ref --symbolic-full-name @{u} | cut -d '/' -f2-) == "master" ]]; then
```
MethodenMann, line 65: Maybe throw an error if branch is not master or nice-releases/* because then someones tries to publish a package on a feature branch?

scripts/release-package.sh
stewue: If you leave the new version empty the suggested version is taken

MethodenMann (approved):

PR #1418, opened 2021-03-31
feat(docs-browser): Add input prop `documentDetailFormName`

rzueger:
Can be defined to override the default document detail form name
(which is "DmsResource")

Refs: TOCDEV-3458

### DIFF
--- a/packages/docs-browser/src/main.js
+++ b/packages/docs-browser/src/main.js
@@ -128,7 +128,8 @@ DocsBrowserApp.propTypes = {
   rootNodes: PropTypes.arrayOf(PropTypes.shape({
     entityName: PropTypes.string,
     key: PropTypes.string
-  }))
+  })),
+  documentDetailFormName: PropTypes.string
 }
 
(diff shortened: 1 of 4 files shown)


packages/docs-browser/src/main.js
```diff
@@ -128,7 +128,8 @@ DocsBrowserApp.propTypes = {
   rootNodes: PropTypes.arrayOf(PropTypes.shape({
     entityName: PropTypes.string,
     key: PropTypes.string
-  }))
+  })),
+  documentDetailFormName: PropTypes.string
```
stewue, line 132: With https://github.com/tocco/tocco-client/pull/1416 I added two input props: listLimit and listFormName. So we should use a consistent naming. Should we rename documentDetailFormName to detailFormName?

rzueger: that's what I wanted to call the prop first, but there are also detail forms for the domains and folders which we might want to be able to override too at some point

stewue (approved):

PR #1423, opened 2021-03-31
Pr/fix search mode

stewue:


rzueger (approved):
Looks good to me, but make sure this condition doesn't evaluate to true too often. `forceDocsViewUpdate()` should be called only if it's really necessary.

PR #1425, opened 2021-04-01
chore: fix intl and txt ressource

yilkur:
Refs: TOCDEV-3216


packages/admin/src/routes/entities/utils/navigationStrategy.js
```diff
@@ -78,7 +80,7 @@ export default (history, match) => {
   }
 }
 
-DetailLinkRelative.propTypes = {
+DetailLinkRelativeWithOutIntl.propTypes = {
```
MethodenMann, line 83: i know you applied my patch but in "without" the "O" should be lowercase

MethodenMann (approved):

PR #1434, opened 2021-04-13
refactor(app-extensions): prevent stacking of modal overlay

yilkur:
Refs: TOCDEV-3494
Changelog: Prevent stacking of modal overlays


MethodenMann (approved):
Do we need to fix this on the pr/notification-center branch as well?

yilkur:
> Do we need to fix this on the pr/notification-center branch as well?

Don't think so. Can't think of a scenario of stacking overlays with notifications...

PR #1435, opened 2021-04-13
fix(tocco-ui): fix empty integer and number edit

stewue:
if an integer or number edit the value should be null and not undefined

Refs: TOCDEV-3375


packages/tocco-ui/src/EditableValue/typeEditors/IntegerEdit.js
```diff
@@ -15,7 +15,11 @@ const IntegerEdit = ({value, onChange, options, immutable, name, id}) => {
   } = options || {}
 
   const handleChange = ({value, floatValue}) => {
-    onChange(allowLeadingZeros ? value : floatValue)
+    let newValue = allowLeadingZeros ? value : floatValue
+    if (!newValue) {
```
MethodenMann, line 19: I think this will not accept 0 as a valid value?

MethodenMann (approved):

PR #1436, opened 2021-04-13
fix(tocco-ui): no initial onChange

MethodenMann:
- in HTML Edit a onChange should only be triggered after user Input and never of api changes. Such api changes can be triggered if the value has content that is not wrapped in an element.

Changelog: Fix html edit initial change
Refs: TOCDEV-336


packages/tocco-ui/src/EditableValue/typeEditors/HtmlEdit.js
```diff
@@ -34,8 +34,8 @@ class HtmlEdit extends React.Component {
     this.lazyQuill = lazy(() => import(/* webpackChunkName: "quill" */ 'react-quill'))
   }
 
-  handleChange = value => {
-    if (this.props.onChange) {
+  handleChange = (value, delte, source) => {
```
stewue, line 37: typo "delte" should be delta?

stewue (approved):

PR #1501, opened 2021-05-04
refactor(docs-browser): adjust breadcrumbs padding

yilkur:
Refs: TOCDEV-3571
Changelog: Adjust left padding of Breadcrumbs in list view

### DIFF
--- a/packages/docs-browser/src/components/DocsBrowser/StyledComponents.js
+++ b/packages/docs-browser/src/components/DocsBrowser/StyledComponents.js
@@ -19,7 +19,7 @@ export const StyledContent = styled.div`
 
 export const StyledBreadcrumbs = styled.div`
   grid-area: breadcrumbs;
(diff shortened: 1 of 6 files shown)


packages/docs-browser/src/components/DocsBrowser/StyledComponents.js
```diff
@@ -19,7 +19,7 @@ export const StyledContent = styled.div`
 
 export const StyledBreadcrumbs = styled.div`
   grid-area: breadcrumbs;
-  ${({embedded}) => embedded && `
+  ${({embedded, hasLeftPadding}) => embedded && !hasLeftPadding && `
```
MethodenMann, line 22: why check for embedded? I think the hasLeftPadding should not depend on the embedded property or you should document it.

MethodenMann (dismissed):

MethodenMann (approved):

PR #1524, opened 2021-05-11
fix(docs-browser): fix navigation problem in move action

stewue:
parent was changed before form name was updated but docsView was only rerendered if parent was changed. instead of passing the form name directly pass a function which determine the form name which results in a synchronous update of the parent and form name

Refs: TOCDEV-3623


packages/docs-browser/src/components/Action/actions/Move.js
```diff
@@ -65,7 +65,7 @@ export const MoveAction = ({
       selectionStyle="none"
       searchFormType="none"
       disableViewPersistor={true}
-      listFormName={parent === null ? 'Move_root_docs_list_item' : 'Move_docs_list_item'}
+      getListFormName={(parent, _) => parent === null ? 'Move_root_docs_list_item' : 'Move_docs_list_item'}
```
MethodenMann, line 68: getListFormName={parent => parent === null ? 'Move_root_docs_list_item' : 'Move_docs_list_item'}
should be fine
underscore for a not used parameter is not a thing we are doing

MethodenMann (approved):

MethodenMann (approved):

PR #1537, opened 2021-05-12
refactor(docs-browser): set hasLeftPadding default to true

yilkur:
Refs: TOCDEV-3568


packages/docs-browser/src/components/DocsBrowser/DocsBrowser.js
```diff
@@ -101,4 +101,8 @@ DocsBrowser.propTypes = {
   hasLeftPadding: PropTypes.bool
 }
 
+DocsBrowser.defaultProps = {
```
MethodenMann, line 104: you should set the initial value in the store. (see reducer.js file with hasLeftèadding in it)

MethodenMann (approved):

PR #1540, opened 2021-05-18
use date field for datetime search when range is not expanded

PadPalon:


packages/tocco-ui/src/Range/Range.js
```diff
@@ -65,20 +47,48 @@ const Range = props => {
     }
   }))
 
+  const typeMapping = rangeTypeMappings[type]
+
+  const getToOptions = (options, fromValue) => {
```
MethodenMann, line 52:
```
const getToOptions = (options, fromValue) => 
   typeMapping && typeMapping.getToOptions
  ? typeMapping.getToOptions(options, fromValue)
   : options
```

would be shorter

MethodenMann (approved):

PR #1590, opened 2021-06-03
chore: add auto cherrypick script

stewue:
- target-branch.txt: branch in which the commits should be cherrypicked
- release script add error codes
- release all packages script do not push (done by cherrypick script)

Refs: TOCDEV-392


scripts/cherrypick.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+set -eu # Exit with nonzero exit code if anything fails
```
MethodenMann, line 2: please add a short comment here what this script is made for (like in tagging.sh) and how it roughly works

scripts/cherrypick.sh
```diff
@@ -0,0 +1,109 @@
...
+
+function getTagCommitId() {
+  # get commit id of tag
+  tag_commit=$(curl -s -H "Authorization: token ${GITHUB_BOT_TOKEN}" \
```
MethodenMann, line 6: Getting a commit id by tagname can be done by git. i did the same in tagging script. i think we should avoid the git api whenever possible to make github interchangeable

PR #1630, opened 2021-06-11
refactor(admin): harmonize menu spacing

yilkur:
Refs: TOCDEV-3749
Changelog: Harmonize menu tree spacing and switch icon size


packages/admin/src/components/MenuTree/MenuTree.js
```diff
@@ -31,7 +14,7 @@ const MenuItem = ({item, typeMapping}) => {
   }
 
   const Component = mappedType.component
-  return <StyledMenuEntry {...item}>
+  return <StyledMenuEntry {...item} childrenCount={item.children.length}>
```
MethodenMann, line 17: I checked it. item can never be undefined. nevertheless i would add a check. Otherwise someone might refactor a part and we get a error here.

PR #1632, opened 2021-06-13
fix(app-extensions): Lower modal z-index to make ExtJS modal visible

rzueger:
- We need to be able to show legacy ExtJS modals on top of our modals
  (e.g. legacy action "Edit rights" on Folder/Domain edit form which
  is shown in a modal)
- The mask of ExtJS modals has z-index 100, we have to remain below it

Refs: TOCDEV-3647


MethodenMann (approved):

MethodenMann:
cherry pick up?

PR #1644, opened 2021-06-16
refactor(tocco-ui): fix popper menu on safari

yilkur:
Refs: TOCDEV-3792
Changelog: Fix popper menu on safari browsers

### DIFF
--- a/packages/tocco-ui/src/Menu/StyledComponents.js
+++ b/packages/tocco-ui/src/Menu/StyledComponents.js
@@ -13,6 +13,10 @@ export const StyledIconButtonWrapper = styled(Button)`
   padding-right: .9rem;
 `
 
+export const StyledPopperWrapper = styled.div`
+  position: relative;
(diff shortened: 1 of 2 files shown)


MethodenMann:
Eventough it might work like this, i'll prefer having the "Cherry-pick: Up" keywork in the message body. And please pay attention to space and uppercase letters.

packages/tocco-ui/src/Menu/StyledComponents.js
```diff
@@ -13,6 +13,10 @@ export const StyledIconButtonWrapper = styled(Button)`
   padding-right: .9rem;
 `
 
+export const StyledPopperRef = styled.div`
```
MethodenMann, line 16: Why "Ref"?  Do you apply position relative because of the react ref prop?
I would rename it

MethodenMann (approved):

PR #1653, opened 2021-06-17
refactor(admin): align nav switch button

yilkur:
Refs: TOCDEV-3749
Changelog: Align nav switch button horizontally


MethodenMann:
i like the new changes but when on top, the menu still cuts of parts of the label
image

MethodenMann (approved):

PR #1756, opened 2021-07-19
feat(entity-browser): allow entity-browser to run with specific bu

PadPalon:
Refs: TOCDEV-3885


packages/entity-browser/src/main.js
```diff
@@ -140,6 +142,7 @@ EntityBrowserApp.propTypes = {
   simpleSearchFields: PropTypes.string,
   initialKey: PropTypes.string,
   nullBusinessUnit: PropTypes.bool,
+  runInBusinessUnit: PropTypes.string,
```
MethodenMann, line 145: please add a description in readme for this new prop

MethodenMann (approved):

PR #1790, opened 2021-07-29
refactor(entity-list): add collapsed bool to store

MethodenMann:
- store search form collapsed property in store. Otherwise it would be ignored when a store is loaded since the state is only represented in a locale state of a component.

Refs: TOCDEV-3985

### DIFF
--- a/packages/entity-list/src/components/EntityList/EntityListContainer.js
+++ b/packages/entity-list/src/components/EntityList/EntityListContainer.js
@@ -2,7 +2,7 @@ import {connect} from 'react-redux'
 import {injectIntl} from 'react-intl'
 
 import EntityList from './EntityList'
-import {initialize, searchFormCollapsedChange} from '../../modules/entityList/actions'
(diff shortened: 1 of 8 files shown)


packages/entity-list/src/components/EntityList/EntityListContainer.js
```diff
@@ -2,21 +2,23 @@ import {connect} from 'react-redux'
...
-  searchFormCollapsedChange
+  setSearchFormCollapsed
 }
 
-const mapStateToProps = state => ({
-  searchFormType: state.input.searchFormType,
-  searchFormPosition: state.input.searchFormPosition,
-  searchFormCollapsedDefault: state.input.searchFormCollapsed
-})
+const mapStateToProps = state => {
```
stewue, line 16: why syntax ({ ... }) is replaced with { return { ... } }?

stewue (approved):

PR #1804, opened 2021-08-04
chore: properly pass artifacts between jobs

pgerber:
Stop misusing the cache for passing build artifacts. For our
runners we use per-host caches. So, one can't assume a cache
from a previous job to be available.


MethodenMann:
Change looks good to me. Could you please change that in nice-release/227 branch and add "Cherry-pick: Up" to the commit message body. Otherwise the CI will still fail for the older version.

PR #1861, opened 2021-08-24
feat(tocco-ui): allow negative integers

PadPalon:
Refs: TOCDEV-4137


PadPalon:
@MethodenMann We discussed this in the refinement, any reason from your side why integers are never allowed to be negative? Previously we'd just configure the field to have a minimal value of 0 if negative numbers weren't allowed.

PR #1888, opened 2021-09-13
Pr/devcon sqllog

rzueger:


rzueger:
This is the same PR as https://github.com/tocco/tocco-client/pull/1887, which was already approved by @PadPalon (new PR was created, because the branch was renamed).

PR #1894, opened 2021-09-21
feat(app-extensions): check cached business unit on session check

PadPalon:
Refs: TOCDEV-4211

### DIFF
--- a/packages/app-extensions/src/login/sagas.js
+++ b/packages/app-extensions/src/login/sagas.js
@@ -1,4 +1,4 @@
-import {consoleLogger} from 'tocco-util'
+import {consoleLogger, cache} from 'tocco-util'
 import {takeLatest, call, all, put} from 'redux-saga/effects'
 
 import notification from '../notification'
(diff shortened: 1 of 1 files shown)


packages/app-extensions/src/login/sagas.js
```diff
@@ -28,9 +28,13 @@ export function getOptions() {
 }
 
 export function* sessionCheck() {
-  const sessionResponse = yield call(doSessionRequest)
-  yield put(actions.setLoggedIn(sessionResponse.success))
-  if (sessionResponse.success) {
+  const {success, businessUnit} = yield call(doSessionRequest)
+  const cachedPrincipal = cache.getShortTerm('session', 'principal')
+  if (cachedPrincipal !== undefined && cachedPrincipal.currentBusinessUnit.id !== businessUnit) {
```
rzueger, line 33: IMO, type-safe checking for `undefined` often makes code more unstable instead of more stable, because it leaves out `null` which eventually also leads to a null pointer.

For this reason, I personally prefer `if (cachedPrincipal && ...` to type-safe checks for `undefined` or `null` for cases like this one. We actually don't care if `getShortTerm()` returns `undefined` or `null` here (or do we?), we only want to avoid a null pointer.

rzueger (approved):

PR #1912, opened 2021-09-23
Pr/remove duplicate router

rzueger:


PadPalon (approved):
Some tests seem to fail, but rest seems good to me

PR #2041, opened 2021-11-03
feat(tocco-ui): table can be refreshed

izzappel:
added refresh button for reloading page data with keeping filter and current page

Changelog: table can be refreshed
Refs:TOCDEV-4290

### DIFF
--- a/packages/tocco-ui/src/Pagination/Pagination.js
+++ b/packages/tocco-ui/src/Pagination/Pagination.js
@@ -11,15 +11,14 @@ import {StyledPagination, StyledPaginationButton} from './StyledComponents'
  */
 const Pagination = ({totalCount, recordsPerPage, currentPage, onPageChange, intl}) => {
   const totalPages = Math.ceil(totalCount / recordsPerPage)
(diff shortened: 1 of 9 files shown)


packages/tocco-ui/src/Pagination/Pagination.js
```diff
@@ -11,15 +11,14 @@ import {StyledPagination, StyledPaginationButton} from './StyledComponents'
  */
 const Pagination = ({totalCount, recordsPerPage, currentPage, onPageChange, intl}) => {
   const totalPages = Math.ceil(totalCount / recordsPerPage)
-  const start = (currentPage - 1) * recordsPerPage + 1
-  const to = currentPage * recordsPerPage
+  const start = Math.min((currentPage - 1) * recordsPerPage + 1, totalCount)
```
stewue, line 14: Math.min for start necessary?

izzappel: not anymore, since we don't show the info text when there is no content. But before it was saying `1 bis 0 von 0` because `start` was at least `1`.

stewue (approved):

PR #2064, opened 2021-11-05
Refresh relations view

PadPalon:


packages/admin/src/routes/entities/modules/path/sagas.js
```diff
@@ -220,6 +223,18 @@ export function* initMultiRelations(model, key) {
...
 }
 
+export function* reloadRelationsInfo({payload: {location}}) {
+  const {pathname} = location
+  const {currentViewInfos, relations} = yield select(entitiesPathSelector)
+  const currentViewInfo = currentViewInfos[pathname]
+  if (currentViewInfo.type === 'detail' && !currentViewInfo.error) {
+    yield spawn(loadRelationInfos, currentViewInfo.model.name, currentViewInfo.key)
+    // adjusting the relations causes the RelationsView to re-render and reload potential newly created entities
+    yield put(actions.setRelations([]))
```
rzueger, line 233: having to call the same action twice with different arguments to achieve (eventually) 1 state change and having to comment it looks a bit hacky to me :)

IMO, this should either be fixed in the component or in the reducer so that the desired re-render and reload takes place with only 1 `setRelations()` action dispatch (or maybe introduce a new additional action for the reload/reinitialization explicitly).

rzueger (approved):

PR #2071, opened 2021-11-09
Pr/code editor

PadPalon:


packages/tocco-ui/src/CodeEditor/AceEditor.js
```diff
@@ -0,0 +1,99 @@
...
+const AceEditor = props => {
+  const {
+    mode,
+    theme = 'textmate',
+    value,
+    onChange,
+    showGutter = true,
+    editorOptions = {}
+  } = props
+  const [reference, setReference] = useState(null)
```
izzappel, line 75: Have you used `useState` here for a specific reason? 
I'm used to use `useRef` for such cases

packages/tocco-ui/src/CodeEditor/AceEditor.js
```diff
@@ -0,0 +1,99 @@
...
+  const [editor, setEditor] = useState(null)
```
izzappel, line 76: I'm not sure if having here a state is ok, since on line 91 you're changing the reference of it without updating the state (with `setEditor`). Also you don't need to get notified by editor changes, as far I can see. 
Isn't here a `ref` the better option?

packages/tocco-ui/src/CodeEditor/AceEditor.js
PadPalon: I forgot that was a thing :)

packages/tocco-ui/src/CodeEditor/AceEditor.js
```diff
@@ -72,26 +72,26 @@ const AceEditor = props => {
...
       aceEditor.getSession().setValue(value)
       aceEditor.on('change', () => onChange(aceEditor.getValue()))
       setEditorConfiguration(aceEditor, props)
-      setEditor(aceEditor)
+      editorReference.current = aceEditor
 
       return () => aceEditor.destroy()
     }
-  }, [reference])
+  }, [containerReference])
```
izzappel, line 88: This not needed anymore => you can use an empty array `[]`
it will then run this once "on mount"

izzappel (approved):

stewue (approved):

PR #2083, opened 2021-11-11
fix(docs-browser): Show create form for remote fields again

rzueger:
No create form was shown in the folder create form when the user clicked
on the "+" of the remote fields for the mailbox senders and receivers.

The barrier that swallowed some propagated actions in the CreateDomain
and CreateFolder actions was too greedy and swallowed the modal notification
actions that should show the create form for the mailbox sender and receiver
roles.

The barrier should *only* swallow the action that shows the notification
with the title "client.entity-detail.createSuccessfulTitle".

Refs: TOCDEV-4410
Changelog: Show create form for remote fields again in the folder creation form


stewue (commented):
also fix packages/docs-browser/src/components/Action/actions/Edit.js

stewue (approved):

PR #2097, opened 2021-11-15
chore: update references to documentation

pgerber:
• https://docs.tocco.ch has been the canonical URL for some time.
• Replace badge as documentation is now built via GitLab CI. Rather
  than using the GitLab CI pipeline badge, I decided to go with a
  static badge. Using the GitLab CI badge, we'd end up with two
  badges labeled "pipeline" which would be more confusing than
  helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 [![gitlab-ci Build Status](https://gitlab.com/toccoag/tocco-client/badges/master/pipeline.svg)](https://gitlab.com/toccoag/tocco-client/-/pipelines/latest)
 [![codecov](https://codecov.io/gh/tocco/tocco-client/branch/master/graph/badge.svg)](https://codecov.io/gh/tocco/tocco-client)
 [![devDependencies Status](https://david-dm.org/tocco/tocco-client/dev-status.svg)](https://david-dm.org/tocco/tocco-client?type=dev)
-[![Documentation Status](https://readthedocs.org/projects/tocco-docs/badge/?version=latest)](http://tocco-docs.readthedocs.io/?badge=latest)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 [![gitlab-ci Build Status](https://gitlab.com/toccoag/tocco-client/badges/master/pipeline.svg)](https://gitlab.com/toccoag/tocco-client/-/pipelines/latest)
 [![codecov](https://codecov.io/gh/tocco/tocco-client/branch/master/graph/badge.svg)](https://codecov.io/gh/tocco/tocco-client)
 [![devDependencies Status](https://david-dm.org/tocco/tocco-client/dev-status.svg)](https://david-dm.org/tocco/tocco-client?type=dev)
-[![Documentation Status](https://readthedocs.org/projects/tocco-docs/badge/?version=latest)](http://tocco-docs.readthedocs.io/?badge=latest)
+[![Documentation Status](https://img.shields.io/static/v1?label=docs&message=Sphinx&color=blue&logoColor=white&logo=readthedocs)](https://gitlab.com/toccoag/tocco-client/-/pipelines/latest)
```
izzappel, line 5: Isn't this the wrong url?
Shouldn't it be `http://docs.tocco.ch/` instead the link to the pipeline which we have already 3 lines above?
Or then the pipeline of the tocco-docs repo? `https://gitlab.com/toccoag/tocco-docs/-/pipelines/latest`

pgerber: Pointed it to the pipeline as the previous badge did so too. Does make more sense to just point to the documentation though.

izzappel (approved):

PR #2104, opened 2021-11-17
fix dms actions if user has no permission

stewue:


packages/docs-browser/src/modules/move/sagas.spec.js
```diff
@@ -38,14 +38,31 @@ describe('admin', () => {
                 ]
               }
 
+              const requestSaga = {method: 'POST', acceptedStatusCodes: [400, 403], body: body}
```
PadPalon, line 41: requestOptions vllt?

PadPalon (approved):

PadPalon (approved):

PR #2112, opened 2021-11-18
refactor(tocco-ui): adjust minimum postcode field width

yilkur:
Refs: TOCDEV-4390
Changelog: adjust min width of location edit postcode field and refactor component


izzappel (approved):
👌

PR #2120, opened 2021-11-22
feat(admin): reload breadcrumbs after entity update

izzappel:
Breadcrumbs are cached and whenever an entity updates
data which is used in the default display the breadcrumbs
are outdated. Therefore on each entity update the breadcrumbs
will be invalidated.

Changelog: reload breadcrumbs after entity update
Refs: TOCDEV-4401


packages/admin/src/routes/entities/modules/path/sagas.js
```diff
@@ -234,6 +244,77 @@ export function* reloadRelationsInfo({payload: {location}}) {
...
+}
+
+export function* getLastBreadcrumbsPath(pathname) {
+  const path = yield call(getPathInfo, pathname)
+
+  if (path === null || !path.entity) {
+    return ''
+  }
+  
+  return [path.entity, path.key, path.relation, path.view].filter(Boolean).join('/')
```
stewue, line 279: add comment

packages/admin/src/routes/entities/modules/path/sagas.js
```diff
@@ -234,6 +244,77 @@ export function* reloadRelationsInfo({payload: {location}}) {
...
+}
+
+export function* getLastEntityInPath(path) {
```
stewue, line 282: remove exports if function is not externally used

stewue (changes requested):

stewue (approved):

PR #2121, opened 2021-11-22
Pr/fix after action reset

PadPalon:


packages/entity-detail/src/modules/entityDetail/sagas.js
```diff
@@ -387,7 +387,7 @@ export function* updateMarked({payload: {entityName, entityId, marked}}) {
 export function* reloadAfterAction({payload}) {
   const {definition} = payload
   if (definition.id !== 'delete') {
-    yield call(loadData, false)
+    yield call(loadData, true)
```
stewue, line 390: yield call(loadData) also possible as true is default value of the reset argument

stewue (approved):

PR #2129, opened 2021-11-24
feat(entity-list): focus first search field in admin search form

PadPalon:
- filtering for inputs with type text does not only include our text fields, but pretty much anything that isn't a checkbox
-- this als works for things like selects and remote fields, since the search box is still a html text input

Refs: TOCDEV-4463
Changelog: the first search field in admin search forms now gets automatically focused on opening

### DIFF
--- a/packages/tocco-util/src/react/useAutofocus.js
+++ b/packages/tocco-util/src/react/useAutofocus.js
@@ -1,0 +1,14 @@
+import {useEffect} from 'react'
+
+const useAutofocus = reference =>
+  useEffect(() => {
+    if (reference.current) {
+      const firstInput = reference.current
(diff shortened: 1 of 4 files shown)


packages/tocco-util/src/react/useAutofocus.js
```diff
@@ -0,0 +1,11 @@
+import {useEffect} from 'react'
+
+const useAutofocus = reference =>
+  useEffect(() => {
+    const firstInput = reference.current.querySelector('input[type = "text"]:not([disabled]), textarea:not([disabled])')
```
izzappel, line 5: I would do a null check for `reference.current` here

izzappel (approved):

PR #2146, opened 2021-11-29
refactor(app-extensions): change notification title

yilkur:
cherry-pick: up
Refs: TOCDEV-4628
Changelog: change notification title and refactor NotificationCenter

### DIFF
--- a/packages/app-extensions/src/notification/modules/center/NotificationCenter/NotificationCenter.js
+++ b/packages/app-extensions/src/notification/modules/center/NotificationCenter/NotificationCenter.js
@@ -1,11 +1,14 @@
 import React, {useEffect, useRef} from 'react'
 import PropTypes from 'prop-types'
-import {LoadingSpinner, Typography} from 'tocco-ui'
+import {LoadingSpinner} from 'tocco-ui'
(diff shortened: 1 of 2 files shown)


packages/app-extensions/src/notification/modules/center/NotificationCenter/NotificationCenter.js
```diff
@@ -26,39 +29,44 @@ const NotificationCenter = (
...
+      loadNotifications(notificationCount)
     }
   }
 
-  const sortedNotifications = Object.keys(notifications)
+  const SortedNotifications = Object.keys(notifications)
     .map(k => notifications[k])
-    .filter(n => !n.result || n.result.type !== resultTypes.outputjob || n.result.file)
-    .filter(n => !n.taskProgress || n.taskProgress.status !== 'cancelled')
+    .filter(n =>
```
stewue, line 43: I think there should be two filters. for example if _n.result.type === resultTypes.entities_ and _n.taskProgress.status === 'cancelled'_ in the old one the element is filtered out. With only one filter the notification is still in the list (which is wrong)

stewue (approved):

PR #2177, opened 2021-12-08
refactor(login): refactor login app

stewue:
Refs: TOCDEV-4477

### DIFF
--- a/packages/login/src/LoginApp.js
+++ b/packages/login/src/LoginApp.js
@@ -1,0 +1,60 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import {appFactory, cache, errorLogging, externalEvents} from 'tocco-app-extensions'
+
+import LoginContainer from './containers/LoginContainer'
+import * as login from './modules/login/actions'
+import * as passwordRequest from './modules/passwordRequest/actions'
+import * as passwordUpdate from './modules/passwordUpdate/dialog/actions'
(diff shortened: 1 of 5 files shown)


packages/login/src/PasswordUpdateApp.js
```diff
@@ -0,0 +1,66 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import {appFactory, errorLogging, externalEvents} from 'tocco-app-extensions'
+import {consoleLogger} from 'tocco-util'
+
+import LoginContainer from './containers/LoginContainer'
+import * as passwordUpdate from './modules/passwordUpdate/dialog/actions'
+import loginReducers, {sagas} from './modules/reducers'
+
+const packageName = 'login'
```
izzappel, line 10: I would use 'passwort-update' here and then you could use `packageName` on line 48

packages/login/src/LoginApp.js
```diff
@@ -0,0 +1,60 @@
...
+    textResourceModules: ['login'],
+    customTheme
+  })
+}
+
+export const LoginApp = props => {
+  const {component} = appFactory.useApp({
+    initApp: initLoginApp,
+    props,
+    packageName: 'id',
```
izzappel, line 44: Here we could now use `packageName` as well, right?

izzappel (approved):

PR #2190, opened 2021-12-10
Pr/anonymous notifications

rzueger:


izzappel (approved):

packages/tocco-util/src/originId/originId.js
```diff
@@ -1,14 +1,14 @@
...
 
 const getOriginId = () => {
-  if (sessionStorage.getItem(originIdName)) {
-    return sessionStorage.getItem(originIdName)
+  if (window[originIdName]) {
+    return window[originIdName]
   }
   const id = `${originIdPrefix}_${uuid()}`
-  sessionStorage.setItem(originIdName, id)
+  window[originIdName] = id
```
rzueger, line 11: Storing the origin ID in the session storage isn't good enough, because `getItem()` and `setItem()` aren't executed atomically. So, race conditions occur quite often so that the origin ID is generated multiple times, which leads to a messy state.

Storing the origin ID prevents these race conditions (because now everything runs atomically). However, this solution might not be good enough.

Use case, that doesn't work anymore:

In the admin app with the notification center notifications are shown as toasters if they're coming from a background app that was started from the current browser tab. Notifications that pertain to the current user too, but weren't "triggered" from the current browser tab, are just added to the notification center silently without showing a toaster. Now that we're storing the origin ID on the window scope, it *changes* when the browser is reloaded, so the notifications which were triggered from the current browser tab before the reload, aren't shown as toasters anymore but just added to the notification center silently. Before with the session storage, the ID didn't change, because the session storage is kept when the page is reloaded.

I created another issue to look into this and to decide if this is an actual problem or if it can be ignored: https://toccoag.atlassian.net/browse/TOCDEV-4729

rzueger: I meant: "... Storing the origin ID *on the global window scope* prevents these race conditions ..."

PR #2198, opened 2021-12-14
Pr/table scrollbehaviour

izzappel:
- entity-list, entity-browser and docs-browser with new `scrollBehaviour` param
- entity-list with new `tableMinHeight` param

**scrollBehaviour=none**: List / Table takes as much space as needed. Scroll behaviours has to be implemented in container or page level. Mainly used for widgets.
**scrollBehaviour=inline**: List / Table takes as much space as given explicitly by the container. When content of table is bigger the tables has a scrollbar itself.
Whenever the container does not provide a specific height the new `tableMinHeight` prop can be used to set a minimum height so that the table shows at least some content.

https://tocco.sharepoint.com/:w:/s/Produkt-Gilde/EZJWWJtCCE1AojZWAjSE1rABFNsYdvWPBc_06qKSNOswjQ?e=4M9Cuv


packages/entity-list/src/containers/TableContainer.js
```diff
@@ -34,7 +34,8 @@ const mapStateToProps = (state, props) => ({
   positions: state.preferences.positions,
   widths: state.preferences.widths,
   markable: state.list.entityModel.markable && state.list.formDefinition.markable,
-  disablePreferencesMenu: state.list.disablePreferencesMenu
+  disablePreferencesMenu: state.list.disablePreferencesMenu,
+  scrollBehaviour: state.entityList.scrollBehaviour
```
rzueger, line 38: Side note (to discuss someday):
Is scrollBehaviour intended to change, once initialised? If not, I'm not sure if we should copy input params like this from `state.input` to another part of the state, or just read it from `state.input` when needed (like we do it a few lines above with the `navigationStrategy`). I think we don't follow a clear guideline here right now.

rzueger (approved):

packages/entity-list/src/containers/TableContainer.js
izzappel: Yes, good catch!
I wanted to test it within Storybook. That's why I added it here so that I could change the property there directly. 
But I was not sure if this is too good since I just added it for testing purposes.

PR #2200, opened 2021-12-14
feat(widget-utils): add widget attribute `data-tocco-package`

rzueger:
By default, the package JS file is derived from the widget name
(which is declared in the `data-tocco-widget` attribute).

There are some cases, where the widget and package name do
not match (e.g. because a package exports multiple widgets,
like the `login` package does (provides `login` and
`update-password` widget)). In these cases, the package name
can be declared explicitly now.

Refs: TOCDEV-4569
Changelog: add widget attribute `data-tocco-package`

### DIFF
--- a/packages/widget-utils/src/bootstrap/bootstrapWidgets.js
+++ b/packages/widget-utils/src/bootstrap/bootstrapWidgets.js
@@ -1,26 +1,31 @@
 import {buildInputFromDom, loadScriptAsync} from './utils'
 
+const getWidgetName = container => container.getAttribute('data-tocco-widget')
+
+const getPackageName = container => container.getAttribute('data-tocco-package') || getWidgetName(container)
+
 const bootstrapWidgets = async params => {
   const {backendUrl} = params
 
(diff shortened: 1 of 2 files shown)


packages/widget-utils/src/bootstrap/bootstrapWidgets.js
```diff
@@ -5,22 +5,29 @@ const bootstrapWidgets = async params => {
...
-    apps.map(app => {
-      return loadScriptAsync(`${backendUrl}/js/tocco-${app}/dist/index.js`)
+    packages.map(packageName => {
+      return loadScriptAsync(`${backendUrl}/js/tocco-${packageName}/dist/index.js`)
     })
   )
 
   widgetContainers.forEach(container => {
     const app = container.getAttribute('data-tocco-widget')
+    const packageName = container.getAttribute('data-tocco-package') || app
```
izzappel, line 24: Small detail... I would prefer to have a small function to get the package name.
This could be used on line 11 as well

izzappel (approved):

PR #2276, opened 2022-01-04
feat(widget-utils): handle errors and log to remote

izzappel:
- handle all kinds of errors while initializing the widgets
- try to log errors on remote

Changelog: handle errors and log to remote
Refs: TOCDEV-4757

### DIFF
--- a/packages/widget-utils/src/bootstrap/bootstrapWidgets.js
+++ b/packages/widget-utils/src/bootstrap/bootstrapWidgets.js
@@ -1,22 +1,19 @@
-import {consoleLogger} from 'tocco-util'
-
 import {ATTRIBUTE_WIDGET_KEY, ERROR_CODE_INVALID_DOMAIN} from './constants'
+import * as remoteLogger from './remoteLogger'
 import {executeRequest, enhanceExtractedBody} from './requests'
 import {getEventHandlers, getTheme, loadScriptAsync} from './utils'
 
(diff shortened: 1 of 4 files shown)


packages/widget-utils/src/bootstrap/bootstrapWidgets.js
```diff
@@ -28,16 +25,21 @@ const initializeWidget = async (backendUrl, container) => {
...
   if (widgetConfig) {
     const {appName, packageName, locale, config} = widgetConfig
-    await loadScriptAsync(`${backendUrl}/js/tocco-${packageName}/dist/index.js`)
+
+    try {
+      await loadScriptAsync(`${backendUrl}/js/tocco-${packageName}/dist/index.js`)
+    } catch (error) {
+      remoteLogger.logException(backendUrl, `Could not fetch package 'tocco-${packageName}'.`, error)
+      return
+    }
```
rzueger, line 42: should we display something to the user if a widget couldn't be loaded  (e.g. an exclamation mark in the widget container) or should it just remain empty/invisible?

rzueger (approved):

packages/widget-utils/src/bootstrap/bootstrapWidgets.js
izzappel: Hmm good question 🤔

izzappel: Same question for errors during the render... or in general.
I will merge this as is and start the discussion with Juan next week.

PR #2288, opened 2022-01-12
feat(widget-utils): run bootstrap script only once

izzappel:
- it can happen that the bootstrap script gets added multiple
times on a page
- the widgets should get initialized only once to
prevent issues and strange behaviours

Changelog: run bootstrap script only once
Refs: TOCDEV-4804

### DIFF
--- a/packages/widget-utils/src/bootstrap/bootstrapWidgets.js
+++ b/packages/widget-utils/src/bootstrap/bootstrapWidgets.js
@@ -1,7 +1,9 @@
-import {ATTRIBUTE_WIDGET_KEY, ERROR_CODE_INVALID_DOMAIN} from './constants'
+import {consoleLogger} from 'tocco-util'
+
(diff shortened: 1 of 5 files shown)


packages/widget-utils/src/bootstrap/bootstrapWidgets.js
```diff
@@ -60,7 +62,35 @@ const initializeWidget = async (backendUrl, container) => {
   }
 }
 
+/**
+ * Widgets should get initialized only once.
+ * Sets global variable to block other embedded bootstrap scripts from execution.
+ * Only the first bootstrap script gets executed.
```
rzueger, line 68: what happens if multiple widgets are embedded like this in the same page:

```
<div data-tocco-widget-key="1"></div>
<script src=".../bootstrap.js"></script>

foo bar...

<div data-tocco-widget-key="2"></div>
<script src=".../bootstrap.js"></script>
```

Is the second widget rendered as well or is it ignored because the second `<div>` is not yet in the DOM when the first bootstrap script is executed? I think we should wait with the bootstrapping until the DOM is fully loaded?

rzueger (approved):

PR #2290, opened 2022-01-12
Pr/widget methods

izzappel:


packages/widget-utils/README.md
```diff
@@ -19,14 +19,56 @@ To embed a widget in the website, just add a `<div>` container and decorate it w
...
+```html
+<div data-id="login" data-tocco-widget-key="12"></div>
+```
+```js
+window.toccoMethods = {
+  login: {
+    setLocale: () => {}
+    ...
+  }
+}
```
rzueger, line 69: This `toccoMethods` obj is the structure we're going to set up for the methods. To me it looks like the user should provide this `toccoMethods` object.

I'd replace it with (or at least add) an example of how they can be called.

Like for example:

```js
// e.g. update locale of the login widget
window.toccoMethods.login.setLocale('en')
```

rzueger (approved):

rzueger (approved):

PR #2295, opened 2022-01-17
feat(app-extensions): set custom client type for rest requests

PadPalon:
Refs: TOCDEV-4648
Cherry-pick: Up


packages/app-extensions/src/form/asyncValidation.js
```diff
@@ -41,7 +41,7 @@ const validateRequest = (formValues, initialValues, mode) => {
   const options = {
     queryParams: {_validate: true},
     method: mode === 'create' ? 'POST' : 'PATCH',
-    headers: {'X-Client-Questions': 'false'},
+    headers: {'X-Client': 'rest'},
```
stewue, line 44: add comment why header is explicitly set (=disable client questions)

stewue (approved):

PR #2315, opened 2022-01-25
feat(entity-list): add query view to AdminSearchForm

PadPalon:
Refs: TOCDEV-4806
Changelog: Add second tab to search form that allows executing custom queries. Add additional actions to open search as query and save query as a new filter.


PadPalon:
The relevant endpoints are already merged into master, but textresources aren't.

packages/entity-list/src/modules/searchForm/sagas.js
```diff
@@ -388,3 +399,85 @@ export function* resetSearchFields() {
   yield call(loadSearchForm, true)
   yield call(resetSearch)
 }
+
+export function* loadSearchAsQuery() {
+  const {where: condition, filter: filters} = yield call(getBasicQuery, false, false)
```
izzappel, line 404: Idea:
Can we use here `getSearchViewQuery` directly instead of `getBasicQuery(false, false)` and therefore remove the second optional parameter `regardQueryView` in `getBasicQuery`?

izzappel (approved):

PR #2349, opened 2022-02-08
fix(tocco-util): fix cache clearance to keep notifications session

izzappel:
- origin id was originally stored in session storage
- since session storage was completely cleared
  origin id was cleared as well
- websocket call and client-id header on rest requests were out of sync
- clearing cache should only consider cache keys
  instead of whole session storage
- keeping origin id in session storage will hydrate notifications
  instead of creating a new session

Refs: TOCDEV-4729
Changelog: fix cache clearance to keep notifications session


rzueger (approved):
well done, thx 🙏 😊

PR #2362, opened 2022-02-11
fix(tocco-ui): fix filtering of select boxes

PadPalon:
Refs: TOCDEV-5003
Changelog: Fixed searching in select boxes. Removed default searchOptions from Select.
Cherry-pick: Up

### DIFF
--- a/packages/tocco-ui/src/Select/Select.js
+++ b/packages/tocco-ui/src/Select/Select.js
@@ -48,9 +48,12 @@ const Select = ({
 
   const getOptions = () => [...(options || [])]
 
+  const searchFunction = useCallback(searchOptions || (() => { /* _throttle expects a function */ }), [searchOptions])
+  const throttledSearchFunction = useCallback(_throttle(searchFunction, 800, {trailing: true}), [searchFunction])
+
   const handleInputChange = (searchTerm, event) => {
     if (searchOptions && searchTerm) {
-      throttledSearchOptions(searchTerm)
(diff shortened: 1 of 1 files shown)


packages/tocco-ui/src/Select/Select.js
```diff
@@ -48,9 +48,12 @@ const Select = ({
 
   const getOptions = () => [...(options || [])]
 
+  const searchFunction = useCallback(searchOptions || (() => { /* _throttle expects a function */ }), [searchOptions])
+  const throttledSearchFunction = useCallback(_throttle(searchFunction, 800, {trailing: true}), [searchFunction])
```
izzappel, line 52: I'm not sure if the `_trottle` really works when you have `[searchFunction]` as a deps here.. or.. if you have `[searchOptions]` as a deps on the line above.
I can remember that there was a problem with that and that the _throttle was not working anymore.

PadPalon: Anything I need to test in particulat? Handling of remote fields and select boxes seemed pretty similar to test300 while testing. @yilkur

izzappel: Yes, there should be a delay when typing before the lookup request is triggered. So the lookup request should not get triggered on every keystroke.

PadPalon: I've removed the dependencies

izzappel (approved):

yilkur (approved):

PR #2363, opened 2022-02-11
Pr/widget theme

yilkur:


packages/app-extensions/src/appFactory/App.js
```diff
@@ -2,19 +2,15 @@ import PropTypes from 'prop-types'
 import React, {useCallback} from 'react'
 import {IntlProvider} from 'react-intl-redux/lib'
 import {Provider} from 'react-redux'
-import styled from 'styled-components'
+import {ToccoTheme} from 'tocco-theme'
+import widgetTheme from "tocco-theme/src/ToccoTheme/widgetTheme"
```
izzappel, line 6: Hm we should not have imports via `src` folder, we should only have top level imports.
That's why I would suggest that you create a new Folder `WidgetTheme` next to `ToccoTheme` and add the widget theme there (and add a default export for it. So that you can import it like
```
import {ToccoTheme, WidgetTheme} from 'tocco-theme'
```

packages/app-extensions/src/appFactory/App.js
```diff
@@ -45,7 +43,8 @@ App.propTypes = {
   store: PropTypes.object.isRequired,
   initIntlPromise: PropTypes.object.isRequired,
   content: PropTypes.node.isRequired,
-  theme: PropTypes.object
+  theme: PropTypes.object,
+  themeType: PropTypes.string
```
izzappel, line 47: I would prefer here explicit prop types for better understanding:
`PropTypes.oneOf(['WIDGET', 'ADMIN'])`

packages/app-extensions/src/appFactory/appFactory.js
```diff
@@ -29,7 +29,8 @@ export const createApp = (
     const initIntlPromise = setupIntl(input, store, name, textResourceModules)
 
     return {
-      component: <App store={store} initIntlPromise={initIntlPromise} content={content} theme={theme} />,
+      component: <App store={store} initIntlPromise={initIntlPromise} content={content} theme={theme}
+                      themeType={input.themeType === 'WIDGET' ? 'WIDGET' : 'ADMIN'}/>,
```
izzappel, line 33: tiny and picky comment:
```
input.themeType || 'ADMIN' 
```
For me, with the `||` logic its's more clear that `'ADMIN'` is the "fallback" or "default". Also, when we would have more types later this code would still be valid.

izzappel (approved):

PR #2413, opened 2022-02-18
fix(app-extensions): legacy actions ignore some exceptions

stewue:
Changelog: legacy actions ignore some exceptions
Cherry-pick: Up
Refs: TOCDEV-5058

### DIFF
--- a/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
+++ b/packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
@@ -6,6 +6,11 @@ import newNotification from '../../../notification'
 import remoteEvents from '../../../remoteEvents'
 import rest from '../../../rest'
 
+const ignoredExceptionMessages = [
+  'Single selection only, aborting...',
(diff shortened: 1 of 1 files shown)


packages/app-extensions/src/actions/modules/actionHandlers/legacyAction.js
```diff
@@ -234,7 +239,13 @@ export default function* (definition, selection, parent, params, config) {
   action.getSelectionNumber = () => selection.type === 'ID' ? selection.ids.length : 0
   action.getSelection = () => legacySelection
 
-  action.perform()
+  try {
+    action.perform()
+  } catch (e) {
+    if (!ignoredExceptionMessages.contains(e)) {
```
rzueger, line 245: - Is there a `contains` method on the array? shouldn't it be `includes(str)`?
- usually `Error` objects are thrown which have the message on the `message` property. Maybe we should check if `e` is an instance of `Error` or of `String` and then act accordingly

rzueger (approved):

PR #2414, opened 2022-02-18
clean up search form types (entity-list and entity-browser)

rzueger:
Changes for the searchFormType prop:
- "none" -> stays the same (no search form at all)
- "simple" -> renamed to "fulltext" (just 1 fulltext search field,
  nothing more)
- "basic" -> renamed to "simple_advanced" (displays the simple form with
  an arrow to expand to the advanced search form)
- "admin" -> stays the same (available for the admin only for
  layouting reasons)
- new type "simple" -> displays to simple search only (not expandable)
- new type "advanced" -> displays the advanced search only (not
  expandable; so this is the same as former type "basic" combined with
  the prop `disableSimpleSearch: true`

The input prop `disableSimpleSearch` now isn't needed anymore
and therefore has been removed.

Refs: TOCDEV-4941
Changelog: cleaned up search form types

### DIFF
--- a/packages/entity-browser/README.md
+++ b/packages/entity-browser/README.md
@@ -10,8 +10,7 @@ React-registry name: `entity-browser`
 | Name                   | Mandatory | Description
 |------------------------|:---------:|-------------
 | `entityName`           |x          | Entityname of records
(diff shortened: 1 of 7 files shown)


izzappel (approved):
👌  well done

PR #2436, opened 2022-02-24
refactor(entity-browser): use route component instead of config

izzappel:
- preparation for react router upgrade
- react router v6 does only support `element` on route component
- much easier to migrate with route components

Refs: TOCDEV-4660


izzappel:
Screenshot 2022-03-01 at 08 29 07
The `index.js` is without code splitting on route-level 13KB bigger.

izzappel:
I don't think that there will be much difference in loading performance in comparison with the version on master.

packages/entity-browser/src/input.js
```diff
@@ -0,0 +1,44 @@
...
+  {
+    name: 'entityName',
+    action: setFormBase,
+    argsFactory: input => [input.formBase || input.entityName]
+  },
+  {
+    name: 'formBase',
+    action: setFormBase,
+    argsFactory: input => [input.formBase]
+  },
```
rzueger, line 33: block starting on line 29 unnecessary? already handled in block on line 24? `entityName` is a mandatory prop, so the block on line 24 will always be executed an do everything that's needed for formBase, doesn't it?

rzueger (approved):

packages/entity-browser/src/input.js
izzappel: hm what happen if just `formBase` has changed without `entityName`?

rzueger: yeah, you're right. 🙈  nevermind

PR #2458, opened 2022-03-03
refactor(entity-detail): convert to func comp

yilkur:
Refs: TOCPROD-41
Changelog: convert entity-detail class compomeents to functional components


izzappel (approved):
👌

PR #2459, opened 2022-03-03
refactor(resource-scheduler): remove react-split-pane

yilkur:
Refs: TOCDEV-4759
Changelog: remove react-split-pane dependency as we already have a library called react-split with almost the same features and us it instead


packages/resource-scheduler/src/components/ResourceScheduler/ResourceScheduler.js
```diff
@@ -1,22 +1,28 @@
...
-
-  .react-bs-container-body {
-    height: auto !important;
-    margin-bottom: ${scale.space(0)};
-  }
-`
+const getGutter = () => () => {
+  const gutterEl = document.createElement('div')
+  ReactDOM.render(
+    <StyledGutter tabIndex={0}>
```
izzappel, line 19: I'm surprised that this is working. I would've assumed that there is no `Theme` available for `StyledGutter` since we are creating a complete new react context with `render`

izzappel: But I have no better idea so far.. I guess when it's working as expected it should be ok for now

izzappel (approved):

PR #2475, opened 2022-03-11
Pr/entity report action

stewue:


packages/core/entity-list/src/modules/list/sagas.js
```diff
@@ -417,6 +417,10 @@ export function* remoteEvent(action) {
     }
   }
 
+  if (event.type === 'trigger-action-event') {
```
izzappel, line 420: other remote events have naming schema from `<name>-<verb>-event`. Should we maybe rename this one to `action-trigger-event` instead?

packages/actions/entity-report/src/main.js
```diff
@@ -0,0 +1,72 @@
...
+
+const EntityReportApp = props => {
+  const {component} = appFactory.useApp({initApp, props, packageName, externalEvents: EXTERNAL_EVENTS})
+  return component
+}
+
+EntityReportApp.propTypes = {
+  selection: selection.propType.isRequired,
+  actionProperties: PropTypes.shape({
+    path_to_reports: PropTypes.string.isRequired
```
izzappel, line 64: Could we rename this to `pathToReports`? Or why do we have this snake case naming here?

izzappel (approved):

PR #2531, opened 2022-03-24
fix(app-extensions): do not reuse same tab for different reports

izzappel:
Changelog: do not reuse same tab for different reports
Cherry-pick: Up
Refs: TOCDEV-5251


yilkur (commented):
Have you deliberately removed the title attr from the anchor tags? This attr provided a handy tooltip when hovering.

yilkur (approved):

PR #2535, opened 2022-03-25
refactor(tocco-ui): convert class component

yilkur:
Refs: TOCDEV-4236

### DIFF
--- a/packages/core/tocco-ui/src/FormattedValue/typeFormatters/LibPhoneFormatter.js
+++ b/packages/core/tocco-ui/src/FormattedValue/typeFormatters/LibPhoneFormatter.js
@@ -1,0 +1,20 @@
+import parsePhoneNumber from 'libphonenumber-js'
+import _isEmpty from 'lodash/isEmpty'
+import PropTypes from 'prop-types'
+import React from 'react'
+
+import Typography from '../../Typography'
+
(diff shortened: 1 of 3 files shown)


izzappel:
I'm wondering if it would make more sense to do something similar like we have here:
https://github.com/tocco/tocco-client/blob/master/packages/core/tocco-ui/src/CodeEditor/LazyCodeEditor.js


```
// file: PhoneFormatter.js
import React, {lazy, Suspense} from 'react'

const LibPhoneFormatter = lazy(() => import(/* webpackChunkName: "phone-formatter" */ './LibPhoneFormatter'))

const LazyPhoneFormatter = props => (
  <Suspense fallback={...}>
    <LibPhoneFormatter {...props} />
  </Suspense>
)
...
export default LazyPhoneFormatter

------------------

// file: LibPhoneFormatter.js
...
import LibPhoneNumber form 'libphonenumber-js'
...

const PhoneFormatter = ({value, breakWords}) => {
  const parsedValue = LibPhoneNumber.parseNumber(value)
  const formattedInput =  _isEmpty(parsedValue) ? value : LibPhoneNumber.formatNumber(parsedValue, 'International')

  return <Typography.Span breakWords={breakWords}>{formattedInput}</Typography.Span>
}
```

@yilkur what do you think?

izzappel (approved):
👌

PR #2571, opened 2022-04-04
feat(entity-list): add prop to use separate list form for search mode

rzueger:
The list form definition that should be used as soon as the user
starts to search can be defined by the input prop `searchListFormName`.
If not defined, the regular list form is used.

Needed for the DMS for the following reasons:
- The first level shows the domains and there should also be a column
  for the type of the domain. Once the user starts to search, there
  not only domains displayed, but also folders and resources;
  therefore, this column should not be visible then anymore.
- Once the user starts to search, a "flat" list is displayed - there's no
  parent node anymore. Therefore, the "new" actions don't make sense
  here and we don't want to display them.

Refs: TOCDEV-5168, TOCDEV-5167
Changelog: new prop `searchListFormName` to use separate list form for search mode


packages/core/entity-list/src/modules/list/sagas.js
```diff
@@ -139,11 +137,17 @@ export function* prepareEndpointUrl(endpoint, searchEndpoint, hasUserChanges) {
...
 export function* countEntities() {
   const {entityName} = yield select(inputSelector)
-  const {endpoint, searchEndpoint} = yield select(listSelector)
+  const {formDefinition, searchListFormDefinition} = yield select(listSelector)
   const {showSelectedRecords, selection} = yield select(selectionSelector)
 
   const regardSelection = !showSelectedRecords
   const query = yield call(getBasicQuery, regardSelection)
+
+  const formDefinitionToUse =
```
izzappel, line 146: could we somehow use the `getFormDefinition` function in `forms.js` here as well. So that we have this logic with `hasUserChanges` only once?

izzappel, line 146: could we somehow use the `getFormDefinition` function in `forms.js` here as well. So that we have this logic with `hasUserChanges` only once?

packages/core/entity-list/src/modules/list/sagas.js
```diff
@@ -281,6 +285,12 @@ export function* fetchEntitiesAndAddToStore(page) {
       limit,
       paths
     }
+
+    const formDefinitionToUse =
```
izzappel, line 289: same here

izzappel, line 289: same here

rzueger:
markable: only main list form is considered (you can't have different marking settings for the main list and the search list form -> should be documented in form model xsd file

packages/core/entity-list/README.md
```diff
@@ -11,6 +11,7 @@ React-registry name: `entity-list`
 |-------------------------------- | :---------: | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------| -------- | --------------|
 | `entityName`                    | *           | Name of the entity that should be listed                                                                                                                                                                  | String   |               |
 | `formName`                      | *           | `formName`/`scope` and `formName`/search will be the used forms. Will only be used for the search form and not for the others, if `listFormDefinition` is set.                                            | String   |               |
+| `searchListFormName`                      | *           | If a different list form should be used once the user started a search, it can be provided using this prop                                            | String   |               |
```
izzappel, line 14: Could you add `searchListFormName` in the `propTypes` in `main.js`?

izzappel (approved):

PR #2596, opened 2022-04-14
refactor(tocco-ui): correct dimension calculation

yilkur:
Refs: TOCDEV-5206
Changelog: correct dimension calculation of select menu since in some situations it does not seem to calculate the correct width/height

### DIFF
--- a/packages/core/tocco-ui/src/Select/Select.js
+++ b/packages/core/tocco-ui/src/Select/Select.js
@@ -1,6 +1,6 @@
 import _throttle from 'lodash/throttle'
 import PropTypes from 'prop-types'
-import {useRef, useCallback} from 'react'
+import {useRef, useState, useCallback} from 'react'
 import ReactSelect from 'react-select'
 import {withTheme} from 'styled-components'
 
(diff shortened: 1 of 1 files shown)


packages/core/tocco-ui/src/Select/Select.js
```diff
@@ -46,6 +46,10 @@ const Select = ({
 }) => {
   const selectComponent = useRef(null)
   const selectWrapper = useRef(null)
+  const getWrapperWidth = () => selectWrapper.current?.clientWidth || 300
+  const getWrapperHeight = () => selectWrapper.current?.clientHeight || 35
+  const [wrapperWidth, setWrapperWidth] = useState(getWrapperWidth)
+  const [wrapperHeight, setWrapperHeight] = useState(getWrapperHeight)
```
izzappel, line 52: Is the initial state working properly?
Shouldn't it be `useState(getWrapperWidth())`?

packages/core/tocco-ui/src/Select/Select.js
```diff
@@ -80,11 +84,10 @@ const Select = ({
 
   const handleFocus = () => {
     selectComponent.current.focus()
+    setWrapperWidth(getWrapperWidth)
+    setWrapperHeight(getWrapperHeight)
```
izzappel, line 88: `setState` function do allow to pass in a function instead of a value. However the idea behind is, that you can access the `prevState` as a param:
```
setState((prev) => [...prev, addedItem]);
```
Therefore I would suggest to just pass in a value here for simplicity reasons: `setWrapperHeight(getWrapperHeight())`

izzappel (approved):

PR #2604, opened 2022-04-19
Pr/adress update

izzappel:


packages/widgets/address-update/src/util/api/entities.js
```diff
@@ -0,0 +1,49 @@
+import {SubmissionError} from 'redux-form/es/SubmissionError'
```
stewue, line 1: add tests?

packages/widgets/address-update/src/modules/addressUpdate/sagas.js
```diff
@@ -0,0 +1,126 @@
...
+  yield call(loadData)
+}
+
+export function* updateFormSubmit(entity, fieldDefinitions) {
+  const {
+    appContext: {widgetConfigKey}
+  } = yield select(inputSelector)
+  const updateResponse = yield call(updateAddress, widgetConfigKey, entity, fieldDefinitions)
+  if (updateResponse.status === 404) {
+    // record was most likely moved to another business unit
```
stewue, line 65: shouldn't we show a toaster here?

packages/widgets/address-update/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "version": "0.1.0",
+  "description": "",
+  "main": "dist/index.js",
+  "scripts": {
+    "compile:prod": "cd ../../../ && npm run compile:prod -- --package=address-update",
+    "prepublish": "npm run compile:prod"
+  },
+  "devDependencies": {
+    "tocco-app-extensions": "*",
+    "tocco-docs-browser": "*",
```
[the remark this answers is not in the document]
izzappel: I think without the dependency the local development is not working properly because the `tocco-docs-browser` is not in the node_modules folder from `address-update` and therefore the files cannot be found.
But yes, if don't need the `chooseDocument` at all, we can remove it completely from `address-update`. But I thought it's a nice "Standardfeature" for document types.

packages/widgets/address-update/src/modules/addressUpdate/sagas.js
izzappel: 👍  removed `404` as `acceptedStatusCodes` in the `updateAddress` fn. so there should be an error toaster now.

stewue (approved):

PR #2608, opened 2022-04-20
fix(entity-list): use preformatted error message for tql

izzappel:
Error message for TQL are already preformatted
in regards of line breaks and spacing. Using the same
format helps the user to find the error way faster.

Refs: TOCDEV-5040
Changelog: use preformatted error message for tql
Cherry-pick: Up

### DIFF
--- a/packages/entity-list/src/components/AdminSearchForm/QueryView.js
+++ b/packages/entity-list/src/components/AdminSearchForm/QueryView.js
@@ -4,7 +4,15 @@ import {FormattedMessage} from 'react-intl'
 import {Ball, BallMenu, EditableValue, FormattedValue, MenuItem, StatedValue} from 'tocco-ui'
 import {react as customHooks} from 'tocco-util'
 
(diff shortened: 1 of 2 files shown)


packages/entity-list/src/components/AdminSearchForm/QueryView.js
```diff
@@ -31,6 +39,11 @@ const QueryView = ({
   }
   const queryExists = !!query && query.trim().length > 0
   const queryHasErrors = queryError && Object.entries(queryError).length !== 0
+
+  const queryErrorValues = Object.keys(queryError || {}).map(key => queryError[key])
```
PadPalon, line 43: Is there a reason to not just use Object.values?

PadPalon (approved):

packages/entity-list/src/components/AdminSearchForm/QueryView.js
izzappel: 🤔  hmm good question.. not really. I will change it 👍

PadPalon (approved):

yilkur (approved):

PR #2621, opened 2022-04-22
Pr/wage view

stewue:


packages/core/entity-list/src/components/ListView/ListView.js
```diff
@@ -62,9 +63,14 @@ const ListView = ({
...
+  const modifiedFormDefinition = useMemo(
+    () => (formDefinition ? modifyFormDefinition(formDefinition) : formDefinition),
+    [formDefinition, modifyFormDefinition]
+  )
+
   const ActionBar = useMemo(() => {
-    if (formDefinition) {
-      const actionBar = getActionBar(formDefinition)
+    if (modifiedFormDefinition) {
+      const actionBar = getActionBar(modifiedFormDefinition)
```
izzappel, line 73: What about on Line 31 ?
Don't we need to use the `modifiedFormDefinition` there as well? I guess for your usecase only changes to the `ActionBar` is need. But the "feature: modifyFormDefintionFn" on the `entity-list` seems to be "not finished" for me.

izzappel: Or we could call the `modifyFormDefinition` function right before we save the `formDefinition` in the state. Then all followup functions / components will use automatically the modified version

packages/core/entity-detail/src/components/DetailForm/DetailForm.js
```diff
@@ -52,16 +53,21 @@ const DetailForm = props => {
...
+
   const embedType = env.getEmbedType()
 
   return (
     <StyledForm {...formEventProps} ref={formEl}>
       <form.FormBuilder
         entity={entity}
         formName={formName}
-        formDefinition={formDefinition}
+        formDefinition={modifiedFormDefinition}
```
izzappel, line 68: Here we have the same situation as we have for the `entity-list`. The `formDefinition` is also used in `DetailView` for the validation...

packages/core/app-extensions/src/form/formModifier.js
```diff
@@ -0,0 +1,101 @@
+/**
+ * remove create action group from main action bar
+ */
+export const removeCreate = formDefinition => ({
+  ...formDefinition,
+  children: formDefinition.children.map(rootItem => {
+    if (rootItem.id === 'main-action-bar') {
```
izzappel, line 7: I would prefer to have a `constants.js` file which includes all these special, const values. (e.g. `main-action-bar`, `output`). It will be easier to find and change them if needed.

packages/core/app-extensions/src/reports/sagas.spec.js
```diff
@@ -0,0 +1,64 @@
+import {expectSaga} from 'redux-saga-test-plan'
+import * as matchers from 'redux-saga-test-plan/matchers'
+import {all, takeLatest} from 'redux-saga/effects'
+import {rest} from 'tocco-app-extensions'
+
+import * as actions from './actions'
+import rootSaga, * as sagas from './sagas'
+
+describe('app-extensinos', () => {
+  describe('form', () => {
```
izzappel, line 10: `describe('reports', ...)`

packages/widgets/wage-view/src/main.js
```diff
@@ -0,0 +1,68 @@
...
+import {env, appContext} from 'tocco-util'
+
+import WageView from './components/WageView'
+
+const packageName = 'wage-view'
+
+const initApp = (id, input, events, publicPath) => {
+  env.setInputEnvs(input)
+  if (input.nullBusinessUnit) {
+    env.setBusinessUnit(env.NULL_BUSINESS_UNIT)
```
izzappel, line 12: We should align those input params. Now we have different versions.  The `docs-browser` only has `businessUnit` and the `entity-browser` has the same as here. And I think `runInBusinessUnit` is the same as `businessUnit`. Maybe we can defined how we want to name it and adjust the `setInputEnv` accordingly and create a follow up ticket to change the "wrong" one

packages/widgets/wage-view/src/components/WageView/WageView.js
```diff
@@ -0,0 +1,45 @@
...
+    return formDefinition
+  }
+
+  return (
+    <LoadMask required={[reports]}>
+      <EntityBrowserApp
+        entityName="Wage"
+        formBase="WageView"
+        searchFilters={searchFilters}
+        modifyFormDefinition={modifyFormDefinition}
```
izzappel, line 30: I think we need to pass in the `appContext` here. At leas the `embedType` is needed otherwise the wrong `theme` will be used

packages/widgets/wage-view/src/main.js
stewue: I already create a new issue https://toccoag.atlassian.net/browse/TOCDEV-5389

izzappel (approved):

PR #2630, opened 2022-04-25
fix(entity-list): fix loading of data when switching modules in admin

rzueger:
Since https://github.com/tocco/tocco-client/pull/2571 the entity-list
didn't load data correctly anymore after switching modules (e.g. opening
Users, then Addresses, then Users again -> list of data wasn't loaded).

Reason for the bug was a race condition where one of the actions
that the `initialize` saga waited for on lines 28 to 30 was already
dispatched before the `take` events were yielded. As a result, the
`initialize` saga was stuck on these lines and
`searchFormActions.executeSearch()` to load the data was never
dispatched.

To fix the issue, the initialization sagas and corresponding
"initialized" actions are handled in sequence instead of parallel.
The downside to this could be a minor loss in performance - but my tests
didn't show a noticeable impact (however with everything (backend server
and database) on my local machine).

Refs: TOCDEV-5168


izzappel (approved):
👌

PR #2665, opened 2022-05-05
fix(tocco-ui): fix too many debounced callbacks being called

PadPalon:
- inline functions would always technically be different on re-renders and triggered dependencies in Debouncer

Refs: TOCDEV-5412
Changelog: fix too many field callbacks being called
Cherry-pick: Up


packages/tocco-ui/src/EditableValue/typeEditorFactory.js
```diff
@@ -52,6 +52,12 @@ const isFlatpickrType = type => ['date', 'datetime'].includes(type)
 const TypeEditorFactory = ({type, value, options, id, events, readOnly = false}) => {
   const flatpickrBlurValue = useRef(undefined)
 
+  const onChange = events.onChange
```
izzappel, line 55: I think `events` could be potentially undefined.

izzappel (dismissed):

izzappel (approved):

PR #2684, opened 2022-05-10
fix(entity-detail): handle save shortcut globally

izzappel:
Refs: TOCDEV-5457
Changelog: handle save shortcut globally
Cherry-pick: Up


izzappel:
We don't want to adjust the shortcuts for now.

PR #2731, opened 2022-05-18
Pr/entity browser business unit

stewue:


packages/widgets/entity-browser/src/main.js
```diff
@@ -67,11 +67,6 @@ const initApp = (id, input, events, publicPath) => {
   input = {...input, id}
 
   env.setInputEnvs(input)
-  if (input.nullBusinessUnit) {
-    env.setBusinessUnit(env.NULL_BUSINESS_UNIT)
-  } else if (input.runInBusinessUnit) {
-    env.setBusinessUnit(input.runInBusinessUnit)
-  }
```
rzueger, line 74: following sequence not needed? where is the BU set now?


```
if (input.businessUnit) {
  env.setBusinessUnit(input.businessUnit)
}
```

stewue: Already handled in [https://github.com/tocco/tocco-client/blob/master/packages/core/tocco-util/src/env/helpers.js](packages/core/tocco-util/src/env/helpers.js)

rzueger (approved):

PR #2736, opened 2022-05-20
fix: add missing props to Table usages

PadPalon:
- somehow when these were undefined useSelection would keep creating new functions which in turn would keep calling a useEffect in Table
- look at this again when someone is around that understand more about it

Refs: TOCDEV-5453


PadPalon:
Replaced by #2754 for proper fix

PR #2746, opened 2022-06-03
fix(app-extensions): check if min or max is defined in validators

PadPalon:
- backend always supplies both, but when only one is configured, the other will be null
- this resulted in fields that accept no values whatsoever

Refs: TOCDEV-5578
Cherry-pick: Up
Changelog: Length and size validators no longer trigger if the max or min is not defined


stewue (approved):

PadPalon:
Last change is just a rebase, merging

PR #2754, opened 2022-06-07
fix(tocco-ui): prevent infinite re-render loop on table

izzappel:
- table without selection feature passed undefined selection as prop
- useSelection default value triggered infinite re-render loop

Refs: TOCDEV-5453
Cherry-pick: Up


izzappel:
Should replace https://github.com/tocco/tocco-client/pull/2736

PadPalon (approved):

PR #2762, opened 2022-06-09
refactor(tocco-ui): refactor phoneEdit

yilkur:
Refs: TOCPROD-41
Changelog: convert phoneEdit to functional component

### DIFF
--- a/packages/widgets/sso-login/src/components/LoginBox/LoginBox.js
+++ b/packages/widgets/sso-login/src/components/LoginBox/LoginBox.js
@@ -8,18 +8,19 @@ import {StyledButtonContainer} from './StyledComponents'
 
 const LoginBox = props => {
   const {loadProviders, autoLogin, providers, loginEndpoint, loginCompleted} = props
-  loadProviders()
-  const prevProps = react.usePrevious(props)
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => loadProviders(), []) // only invoke the loadprovider func once at the beginning
+  const prevProps = react.usePrevious(providers)
 
   useEffect(() => {
-    if (autoLogin && prevProps.providers.length === 0 && providers.length > 0) {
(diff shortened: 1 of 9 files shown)


packages/widgets/sso-login/src/components/LoginBox/LoginBox.js
```diff
@@ -1,51 +1,31 @@
...
+import {useEffect} from 'react'
+import {react} from 'tocco-util'
 
 import {openLoginWindow} from '../../utils/loginWindow'
 import ProviderButton from '../ProviderButton/ProviderButton'
+import {StyledButtonContainer} from './StyledComponents'
+
+const LoginBox = props => {
+  const {loadProviders, autoLogin, providers, loginEndpoint, loginCompleted} = props
+  loadProviders()
```
izzappel, line 11: I think this should be done within a `useEffect`: 
```
useEffect(() => loadProviders(), [])
```
Right now it will be invoked on every render. Before it was only invoked once when the component was created.

packages/widgets/sso-login/src/components/LoginBox/LoginBox.js
```diff
@@ -1,51 +1,31 @@
...
+  const prevProps = react.usePrevious(props)
```
izzappel, line 12: I would store here only the `providers` and not all props

packages/widgets/sso-login/src/components/LoginBox/LoginBox.js
```diff
@@ -1,51 +1,31 @@
...
-            key={idx}
-            provider={provider}
-            loginEndpoint={this.props.loginEndpoint}
-            loginCompleted={this.props.loginCompleted}
-          />
-        ))}
-      </StyledButtonContainer>
-    )
-  }
+  })
```
izzappel, line 22: and then add a dependency array for `[providers, autoLogin, loginEndpoint, loginCompleted]`

izzappel (approved):

PR #2785, opened 2022-06-17
create mailing-list widget

PadPalon:


PadPalon:
@stewue @izzappel I've added both of you as reviewers because it's a bit of a chonker of a PR :)

packages/core/app-extensions/src/form/formModifier.js
```diff
@@ -5,6 +5,70 @@ export const ACTION_SAVE_ID = 'save'
...
+
+export const removeActions = (formDefinition, actionIds) => ({
+  ...formDefinition,
+  children: formDefinition.children.map(rootItem => {
+    if (rootItem.id === MAIN_ACTION_BAR_ID) {
+      return {
+        ...rootItem,
+        children: rootItem.children
+          .map(group => {
+            if (group.componentType === 'action-group') {
```
izzappel, line 29: Can we move those static strings to a constant (like we have it for MAIN_ACTION_BAR_ID)?

packages/widgets/mailing-list/src/components/MailingList/MailingList.js
```diff
@@ -0,0 +1,84 @@
+import _get from 'lodash/get'
+import PropTypes from 'prop-types'
+import {useEffect} from 'react'
+import {rest, form} from 'tocco-app-extensions'
+import EntityBrowserApp from 'tocco-entity-browser/src/main'
+import {searchFormTypePropTypes} from 'tocco-entity-list/src/main'
+import {LoadMask} from 'tocco-ui'
+import {appContext as appContextPropType} from 'tocco-util'
+
+export function* modifyFormDefinition(
```
izzappel, line 10: Is it necessary to have it as generator function? can't we use a normal function in sagas?

izzappel (approved):

packages/widgets/mailing-list/src/main.js
```diff
@@ -0,0 +1,66 @@
...
+  limit: PropTypes.number,
+  backendUrl: PropTypes.string,
+  businessUnit: PropTypes.string,
+  appContext: appContextPropType.propTypes.isRequired,
+  reportIds: PropTypes.arrayOf(PropTypes.string),
+  allowEmail: PropTypes.bool
+}
+
+export default MailingListApp
+export const app = appFactory.createBundleableApp(packageName, initApp)
```
izzappel, line 66: I changed the `createBundleableApp` function and it needs a third parameter now.. 
```
export const app = appFactory.createBundleableApp(packageName, initApp, MailingListApp)
```

packages/core/app-extensions/src/form/formModifier.js
```diff
@@ -5,6 +5,70 @@ export const ACTION_SAVE_ID = 'save'
...
+          })
+          .filter(group => group.componentType !== 'action-group' || group.children.length > 0)
+      }
+    }
+
+    return rootItem
+  })
+})
+
+export const adjustAction = (formDefinition, actionId, adjuster) => ({
```
stewue, line 45: add a comment on how the function works

packages/core/app-extensions/src/form/formModifier.js
```diff
@@ -5,6 +5,70 @@ export const ACTION_SAVE_ID = 'save'
...
+    .map(item => {
+      if (item.componentType === 'layout') {
+        return removeBoxes(item, boxIds)
+      }
+
+      return item
+    })
+})
+
+export const removeActions = (formDefinition, actionIds) => ({
```
stewue, line 21: you currently assume that an action is in an action group. i don't now if we want this assumption (possibly an action can be placed directly in the main action bar for widgets). Additionally if we have "action group A -> action group A1 -> action" we cannot remove the action.

packages/widgets/mailing-list/src/components/MailingList/MailingList.js
```diff
@@ -0,0 +1,84 @@
...
+) {
+  if (formDefinition.id === 'Mailing_list_detail_relUser_list') {
+    let userListFormDefinition = formDefinition
+    if (reports?.length > 0) {
+      userListFormDefinition = form.addReports(userListFormDefinition, reports, intl)
+    }
+    if (!showEmailAction) {
+      userListFormDefinition = form.removeActions(userListFormDefinition, ['mailing-list-mail-action'])
+    } else {
+      userListFormDefinition = form.adjustAction(userListFormDefinition, 'mailing-list-mail-action', action => ({
```
stewue, line 23: maybe add a comment here that we late easily know why this is necessary

packages/widgets/mailing-list/src/components/MailingList/MailingList.js
stewue: i think the problem is the "yield rest.fetchEntity(...)" so depending on this response the form definition is differently modified

stewue (approved):

PR #2788, opened 2022-06-22
Pr/customer custom actions

izzappel:


stewue (approved):

packages/core/app-extensions/src/actions/modules/dynamicActions/sagas.spec.js
```diff
@@ -0,0 +1,48 @@
...
+              const actionPackages = [{packageName: 'package', appName: 'app', actionName: 'action'}]
+              return expectSaga(sagas.fetchActionPackages)
+                .provide([[matchers.call.fn(rest.requestSaga), {body: {actionPackages}}]])
+                .call(rest.requestSaga, '/client/actionPackages', {method: 'GET'})
+                .put(actions.setActionPackages(actionPackages))
+                .run()
+            })
+
+            test('should store empty actionPackages when malformed body is returned', () => {
+              return expectSaga(sagas.fetchActionPackages)
```
rzueger, line 37: (`return` as single statement not necessary)

packages/core/tocco-util/src/bundle/ErrorComponent.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from 'prop-types'
+
+import nice from '../nice'
+
+const ErrorComponent = ({packageName, appName}) => {
+  if (__DEV__ || nice.getRunEnv() !== 'PRODUCTION') {
+    return (
+      <div>
+        Fehler beim Laden der &quot;{appName}&quot; App im Package &quot;{packageName}&quot;.
```
rzueger, line 9: i18n?

packages/core/tocco-util/src/bundle/utils.js
```diff
@@ -0,0 +1,8 @@
+export const getDistPath = packageName => {
+  return `/js/tocco-${packageName}/dist/`
```
rzueger, line 2: (`return` as single statement not necessary)

rzueger (approved):

packages/core/app-extensions/src/actions/actionFactory.js
```diff
@@ -0,0 +1,84 @@
...
+const actionFactory = actions => {
+  const LazyActionWrapper = props => {
+    const {appId, fetchActionPackages, actionPackages, ...actionProps} = props
+
+    useEffect(() => {
+      fetchActionPackages()
+    }, [fetchActionPackages])
+
+    // wait until all actions have beend loaded
+    if (typeof actionPackages === 'undefined' || actionPackages === null) {
```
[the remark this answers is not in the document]
izzappel: I learned that a check to `undefined` is more save when used with `typeof` operator. Since then I always does it like this.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#typeof_operator_and_undefined

PR #2819, opened 2022-07-04
adjustements for Own_time_table widget

PadPalon:


packages/core/app-extensions/src/reports/sagas.js
```diff
@@ -22,22 +24,46 @@ export function* loadReport({payload: {reportIds, entityName, scope}}) {
...
+    showConfirmation: !report.ignore_selection,
+    icon: yield call(getReportIcon, report)
+  }
+}
 
-  yield put(actions.setReports(reports))
+function* getReportIcon(report) {
+  const reportIcons = yield call(loadReportIcons)
+  const fileType = report['relOutput_template.relReport_file_format.unique_id']
+  return reportIcons[fileType]
```
stewue, line 55: if a file type has no icon defined in the backend should there be a generic fallback icon or just no icon?

stewue (approved):

packages/core/app-extensions/src/reports/sagas.js
PadPalon: I'd go with there being no icon, since that is what ReportActionFormInterceptor in the backend does currently.

PR #2839, opened 2022-07-11
refactor(tocco-ui): style new date picker

yilkur:
Refs: TOCDEV-5695
Changelog: style new date picker like the old one


izzappel (approved):
👌

PR #2845, opened 2022-07-12
fix(app-extensions): correct usage of textresource in validation

PadPalon:
Refs: TOCDEV-5681
Cherry-pick: Up


izzappel (dismissed):
@PadPalon shouldn't we add a changelog to the commit message here?

PadPalon:
I wasn't sure what a good changelog would be for something like this. "XY works again" in changelogs always feels kinda redundant to me. But maybe I can think of something.

izzappel:
Maybe: "translate fallback error message properly"
When we do not have a Changelog message in the commit message, the package will not be released.

izzappel (approved):

PR #2903, opened 2022-07-29
chore: fix changelog generation

stewue:
currently only dependencies in the package.json be taken into account for the changelog generation. Finding the dependencies improved that transitive dependencies are now also considered

Refs: TOCDEV-5746


izzappel (approved):
👌 nice 😅

PR #2981, opened 2022-08-23
Pr/button responsiveness

izzappel:


packages/core/tocco-ui/src/Button/ButtonContext.js
```diff
@@ -0,0 +1,87 @@
...
+      if (node) {
+        resizeObserverRef.current = new ResizeObserver(entries => handleResize(entries))
+        resizeObserverRef.current.observe(node)
+      }
+    },
+    [handleResize]
+  )
+
+  useEffect(() => {
+    return () => {
```
rzueger, line 31: (`return` not necessary -> inline)

packages/core/tocco-ui/src/Button/ButtonContext.js
```diff
@@ -0,0 +1,87 @@
...
+}
+
+ButtonContextProvider.propTypes = {
+  children: PropTypes.func.isRequired
+}
+
+export const useButtonContext = () => {
+  const buttonContext = useContext(ButtonContext)
+  return buttonContext
+}
```
rzueger, line 85: (could be simplified?)

`export const useButtonContext = () => useContext(ButtonContext)`

packages/core/tocco-ui/src/Button/ButtonContext.js
rzueger, line 31: (`return` as single statement not necessary)

packages/core/tocco-ui/src/Button/buttonStyle.js
```diff
@@ -63,13 +63,25 @@ export const buttonStyle = css`
...
       }
+    `
+
+    if (icon && showIconOnly) {
+      return hideLabelStyles
+    }
+
+    return (
+      icon &&
+      typeof showIconOnly === 'undefined' &&
```
rzueger, line 79: this line looks a bit confusing to me. Is there a difference if `showIconOnly` is `undefined`, `null` or `false` or would it all be the same? Checking for `undefined` explicitly makes me expect that `undefined` really should be handled differently than e.g. `false`, but I can't see why this would be the case?

packages/core/entity-list/src/components/ListView/getActionBarContent.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  selection: [],
+  refresh: EMPTY_FUNC,
+  currentPageIds: ['1', '4'],
+  showSelectionController: true,
+  columnDisplayPreferences: {},
+  preferencesLoaded: true
+}
+
+describe('entity-list', () => {
+  describe('components', () => {
```
rzueger, line 39: (`ListView` namespace ignored intentionally?)

packages/core/tocco-ui/src/Button/buttonStyle.js
izzappel: So when having a valid value (`true` | `false`) the icon visibility is set explicitly. When having no value (`undefined`) the icon visibility is not set explicitly and the default responsiveness behaviour is used

izzappel: I've changed the context value, so that it easier to read 👍

packages/core/tocco-ui/src/Button/ButtonContext.js
```diff
@@ -0,0 +1,94 @@
+import PropTypes from 'prop-types'
+import React, {useContext, useMemo, useState, useCallback, useEffect, useRef} from 'react'
+
+export const LabelVisibility = {
+  visible: 'visible',
+  hidden: 'hidden',
+  responsive: 'responsive'
+}
```
rzueger, line 8: I like that :) much clearer than a boolean with 3 states 👍

yilkur (approved):

PR #2998, opened 2022-08-29
feat(app-extensions): generate reports without dialog in widgets

stewue:
Refs: TOCDEV-5672
Changelog: generate reports without dialog in widgets


packages/core/app-extensions/src/actions/modules/actionHandlers/report.spec.js
```diff
@@ -61,6 +75,34 @@ describe('app-extensions', () => {
             })
           })
 
+          describe('generateReportWithoutSettings', () => {
+            test('generate a without opening the modal', () => {
```
rzueger, line 79: ('generate a **report** without ...')

rzueger (approved):

PR #3001, opened 2022-08-30
fix(entity-list): handle changed props again

PadPalon:
Refs: TOCDEV-5934
Changelog: let entity list handle changed props again


PadPalon:
Correctly fixed in #3060

PR #3041, opened 2022-09-09
Pr/mobile menu

yilkur:


packages/apps/admin/src/components/Admin/Admin.js
```diff
@@ -95,7 +95,7 @@ const Admin = ({
     </>
   )
 
-  const adminForbiddenContent = adminAllowed === false && (
+  const adminForbiddenContent = !adminAllowed && (
```
stewue, line 98: I think `adminAllowed === false` was explicitly used here as `adminAllowed` is initially `undefined`. After a rest request it is set to true/false.

stewue (approved):

PR #3047, opened 2022-09-12
fix(tocco-ui): fix date picker icon

stewue:
variable was renamed in version 3.4

Changelog: fix date picker icon
Refs: TOCDEV-5930


yilkur (approved):
lgtm

PR #3078, opened 2022-09-22
refactor(tocco-ui): replace touch helper function

yilkur:
Refs: TOCDEV-5890


packages/core/tocco-ui/src/utilStyles/detectTouchDevice.js
```diff
@@ -0,0 +1,4 @@
+const detectTouchDevive = () =>
```
izzappel, line 1: `detectTouchDevive` => `detectTouchDeviCe`

izzappel (approved):

PR #3087, opened 2022-09-27
fix(entity-list): do not refresh list multiple times

izzappel:
- changedProps.store is undefined when store has not changed
- nevertheless it could be that store is defined on props
- to figure out if store has been removed from props it's needed to check if store is set on props and not on changedProps

Refs: TOCDEV-5912
Changelog: do not refresh list multiple times


izzappel:
Fixes also: https://toccoag.atlassian.net/browse/TOCDEV-6079

PadPalon (approved):

PR #3089, opened 2022-09-27
feat(app-extension): change HTTP method of `rest.fetchEntity` to POST

rzueger:
Caution: depends on https://git.tocco.ch/c/nice2/+/45699 (new POST REST
         endpoint must be merged first)

Details:
- When fetching an entity, the field and relation values needed to be specified in the `_paths` query param
- If there are many paths to request, the URL can become very long (longer than 2000 characters which is the de facto limit for the URL length) which causes requests to fail (see TOCDEV-6081).
- To get around the problem, the entity can be fetched via POST request now (send `paths` array in body instead of comma separated `_paths` query param) -> this is also the new default
- If you still want to use GET as the method, you can specify it in the request options for `rest.fetchEntity`

Refs: TOCDEV-6081
Changelog: change HTTP method of `rest.fetchEntity` to POST
Cherry-pick: Up

### DIFF
--- a/packages/core/app-extensions/src/rest/helpers.js
+++ b/packages/core/app-extensions/src/rest/helpers.js
@@ -12,17 +12,28 @@ import {requestSaga} from './rest'
  * @param entityName {String} Name of the entity
  * @param key {String} key of the entity
  * @param query {Object} see 'buildRequestQuery' function
+ * @param requestOptions {Object} An object which can contain the following options:
+ * - method {String} HTTP Method of request. Default is POST.
  * @param transformer {function} Function to directly manipulate the result. By default returns data attribute
  */
-export function* fetchEntity(entityName, key, query, transformer = json => json) {
+export function* fetchEntity(entityName, key, query, requestOptions = {}, transformer = json => json) {
+  const {method = 'POST'} = requestOptions
   const requestQuery = yield call(buildRequestQuery, query)
   const options = {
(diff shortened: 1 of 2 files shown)


packages/core/app-extensions/src/rest/helpers.js
```diff
@@ -12,17 +12,27 @@ import {requestSaga} from './rest'
  * @param entityName {String} Name of the entity
  * @param key {String} key of the entity
  * @param query {Object} see 'buildRequestQuery' function
+ * @param requestOptions {Object} An object which can contain the following options:
+ * - method {String} HTTP Method of request. Default is POST.
  * @param transformer {function} Function to directly manipulate the result. By default returns data attribute
  */
-export function* fetchEntity(entityName, key, query, transformer = json => json) {
+export function* fetchEntity(entityName, key, query, {method = 'POST'} = {}, transformer = json => json) {
```
PadPalon, line 19: Destructuring and Default Assignment both in the parameter list is pretty neat, but I wonder if we should just do this in the method itself to keep it easily understandable in the future?

PadPalon (approved):

packages/core/app-extensions/src/rest/helpers.js
rzueger: absolutely right 👍  moved it into the method body.

PR #3145, opened 2022-10-20
feat(login): redirect from login based on query parameter

PadPalon:
Refs: TOCDEV-6060
Changelog: allow login to redirect to another page from an url parameter


packages/widgets/login/src/modules/sagas.js
```diff
@@ -81,12 +81,22 @@ export function* handleFailedResponse() {
...
     window.location.href = redirectUrl
   }
 }
 
+export function* loadRedirectUrl() {
+  const {_redirect_url: queryUrl} = queryString.fromQueryString(window.location.search)
+  if (queryUrl) {
+    return queryUrl
+  } else {
+    const {inputUrl} = yield select(inputSelector)
```
stewue, line 95: question(blocking): is this still working after changing `redirectUrl` to `inputUrl`?

stewue (approved):

PR #3150, opened 2022-10-21
refactor(admin): correct height calculation

yilkur:
Refs: TOCDEV-6118
Changelog: correct height calculation method of admin to always ensure that the full viewport height is filled


packages/apps/admin/src/components/Admin/StyledComponents.js
```diff
@@ -2,23 +2,22 @@ import SlideMenu from 'react-burger-menu/lib/menus/slide'
...
+  const viewportHeight = Math.max(document.documentElement.clientHeight || 0, window.innerHeight || 0)
+  doc.style.setProperty('--doc-height', `${viewportHeight}px`)
 }
 
-calcViewportHeight()
-
-window.addEventListener('resize', () => calcViewportHeight())
-window.addEventListener('orientationchange', () => calcViewportHeight())
+documentHeight()
+window.addEventListener('resize', () => documentHeight())
```
izzappel, line 12: suggestion (blocking): Move height calculation into a separate component.

I know it was already like that before... but.. when having this logic in a react component we would have the `addEventListener` under control and we wouldn't need to add a css variable.

Does this then still work?

```
const Wrapper = ({children}) => {
  const docHeight = useRef(null)

  const documentHeight = () => {
    const viewportHeight = Math.max(document.documentElement.clientHeight || 0, window.innerHeight || 0)
    docHeight.current = `${viewportHeight}px`
  }

  useEffect(() => {
    documentHeight()
    window.addEventListener('resize', documentHeight)

    return () => {
      window.removeEventListener('resize', documentHeight)
    }
  }, [])

  return <StyledWrapper height={docHeight.current}>{children}</StyledWrapper>
}
```

packages/core/entity-list/src/components/SearchView/StyledComponents.js
```diff
@@ -3,7 +3,7 @@ import styled from 'styled-components'
 export const StyledAdminSearchFormWrapper = styled.div`
   display: flex; // enables StyledPlaceHolder to take full height
   flex-direction: column;
-  height: calc(var(--vh, 1vh) * 100 - 77px); // TODO: This is only a temporary workaround, height needs to be 100%
+  height: 100%;
```
izzappel, line 6: praise: very nice that you could remove this TODO!

izzappel (approved):
praise: 👏 nice!

PR #3176, opened 2022-11-01
feat(bettinggame): Add TournamentWidget

LorenzKieran:


packages/widgets/bettinggame/src/components/TournamentWidget.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from 'prop-types'
+import EntityDetailApp from 'tocco-entity-detail/src/main'
+
+const TournamentWidget = props => {
+  const {tournamentKey, formBase} = props
+  return (
+    <EntityDetailApp
+      id={`detail_${formBase}_${tournamentKey}`}
+      entityName="Tournament"
```
izzappel, line 9: suggestion(blocking): The `entityName` should come from the input props

The `formBase` and `entityName` are both immutable configs form the widget config. Please also use `entityName` from the config instead of using hardcoded name here.

packages/widgets/bettinggame/src/components/TournamentWidget.js
```diff
@@ -0,0 +1,22 @@
...
+      entityId={tournamentKey}
+      formName={formBase}
+      mode="update"
+    />
+  )
+}
+
+TournamentWidget.propTypes = {
+  tournamentKey: PropTypes.number,
```
izzappel, line 18: question(blocking): Are you sure that this is of type `number` and not of type `string`?

packages/widgets/bettinggame/src/main.js
```diff
@@ -0,0 +1,63 @@
+import {appFactory, selection} from 'tocco-app-extensions'
+import {reducer as reducerUtil, appContext} from 'tocco-util'
+
+import TournamentWidget from './components/TournamentWidget'
+import reducers, {sagas} from './modules/reducers'
+
+const packageName = 'bettinggame'
+
+const initApp = (id, input, events, publicPath) => {
+  const content = <TournamentWidget formBase={input.formBase} tournamentKey={input.tournamentKey} />
```
izzappel, line 10: suggestion(blocking): so that the components run in the correct environment the `env` has to be set correctly.

https://docs.tocco.ch/framework/configuration/widgets/widget.html#env 
- No footer information should be visible 
image
- correct theme has to be applied

packages/widgets/bettinggame/src/components/TournamentWidget.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from 'prop-types'
+import EntityDetailApp from 'tocco-entity-detail/src/main'
+
+const TournamentWidget = props => {
+  const {tournamentKey, formBase} = props
```
izzappel, line 5: nitpick: align with our coding guidelines

https://docs.tocco.ch/framework/client/code-styleguide.html#react-components
- destruct props in function parameter
```
const TournamentWidget = ({tournamentKey, formBase}) => {}
```

- return component immediately
```
const TournamentWidget = ({...}) => (<EntityDetailApp ... />)
```
instead
```
const TournamentWidget = ({...}) => {
    return (<EntityDetailApp ... />)
}
```

izzappel (changes requested):

izzappel (approved):

PR #3178, opened 2022-11-02
feat(entity-dms): add widget entity-dms

stewue:
Changelog: add widget entity-dms
Refs: TOCDEV-6084


packages/widgets/entity-dms/src/components/EntityDms/EntityDms.js
```diff
@@ -0,0 +1,55 @@
...
+      return 'DmsPublic_root_docs_list_item_specific'
+    } else {
+      return 'DmsPublic_root_docs_list_item'
+    }
+  }
+
+  const rootNodes = []
+
+  if (folderKey) {
+    rootNodes.push({
```
izzappel, line 20: nitpick: I personally do not like functions that mutate objects/arrays and I try to write code that only works with immutable objects

```
const rootNodes = folderKey ? [{...}] : []
```

izzappel (approved):

PR #3191, opened 2022-11-08
Pr/invalid session

stewue:


izzappel:
When logging in with a different user on the invalid session modal, the username and business unit is not changing on the top right corner. Only after a manual page reload these data are correct again.

packages/core/app-extensions/src/login/sagas.js
```diff
@@ -17,18 +14,3 @@ export function doRequest(url, options) {
...
 }
-
-export function* sessionCheck() {
-  const {success, businessUnit, adminAllowed} = yield call(doSessionRequest)
-  const cachedPrincipal = cache.getLongTerm('session', 'principal')
-  if (cachedPrincipal && cachedPrincipal.currentBusinessUnit.id !== businessUnit) {
-    yield call(cache.clearAll)
-  }
-  yield put(actions.setAdminAllowed(adminAllowed))
-  yield put(actions.setLoggedIn(success))
```
izzappel, line 28: So before we set `adminAllowed` before `loggedIn`. This caused that `adminAllowed` always had a value when Admin is rendered.
Now we set `loggedIn` separately and only set `adminAllowed` when we already render `Admin` component. This causes the prop types of `Admin` to be unhappy because `adminAllowed` prop is required.

I still try to understand all these changes.. nevertheless either the proptypes are not correct anymore or we need to be more careful with setting the loggedIn and adminAllowed values

izzappel (approved):

packages/core/app-extensions/src/login/sagas.js
stewue: `adminAllowed` is set to `undefined` in the file `packages/apps/admin/src/modules/session/sagas.js`. So I think should the proptype was wrong

PR #3193, opened 2022-11-08
feat(bettinggame): add custom react action

TryFresco:


packages/actions/scoreboard/src/components/Scoreboard.spec.js
```diff
@@ -0,0 +1,80 @@
...
+                  entity: 'User',
+                  key: 2
+                }
+              }
+            ]}
+            fetchData={() => {}}
+          />
+        )
+
+        formRender.debug()
```
izzappel, line 60: suggestion(blocking): remove debug()

`debug()` is actually similar to `console.log(..)`. It's useful during implementation but should not be pushed to master.

packages/actions/scoreboard/src/modules/scoreboard/sagas.spec.js
```diff
@@ -0,0 +1,75 @@
...
+import {all, select, takeEvery} from 'redux-saga/effects'
+import {rest} from 'tocco-app-extensions'
+
+import * as actions from './actions'
+import rootSaga, * as sagas from './sagas'
+
+describe('scoreboard', () => {
+  describe('modules', () => {
+    describe('sagas', () => {
+      test('should fork child sagas', () => {
```
izzappel, line 12: praise: nice once, without any explanation from my side

packages/actions/scoreboard/src/modules/scoreboard/sagas.spec.js
```diff
@@ -0,0 +1,75 @@
...
+import * as matchers from 'redux-saga-test-plan/matchers'
+import {all, select, takeEvery} from 'redux-saga/effects'
+import {rest} from 'tocco-app-extensions'
+
+import * as actions from './actions'
+import rootSaga, * as sagas from './sagas'
+
+describe('scoreboard', () => {
+  describe('modules', () => {
+    describe('sagas', () => {
```
izzappel, line 11: suggestion(non-blocking): the "describe-path" is not 100% correct

it should reflect the folder path (omit `packages/actions|core|apps|widgets|...` and `src`)
this would result into `scoreboard`, `modules`, `scoreboard`, `sagas`

packages/actions/scoreboard/src/components/Scoreboard.js
```diff
@@ -0,0 +1,68 @@
...
+    )
+  } else {
+    content = data.map(entry => {
+      const defaultDisplay = <p>{entry.participantBean.defaultDisplay}</p>
+      const rank = <p>Ranking: {entry.ranking}</p>
+      const points = <p>Points: {entry.points}</p>
+
+      const returnValue = [defaultDisplay, rank, points]
+      return returnValue
+    })
```
izzappel, line 44: suggestion(non-blocking): usually we would write this in jsx to have a clearer understanding of the output

e.g.
```
content = data.map(({ranking, points, participantBean}) => (
      <React.Fragment key={participantBean.key}>
        <p>{participantBean.defaultDisplay}</p>
        <p>Ranking: {entry.ranking}</p>
        <p>Points: {entry.points}</p>
      <React.Fragment/>
    )}
```

React.Fragment is needed when having multiple jsx-elements: https://reactjs.org/docs/fragments.html
`key` prop is needed for arrays: https://reactjs.org/docs/lists-and-keys.html#keys

packages/actions/scoreboard/src/components/Scoreboard.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+  return (
+    <StyledContainer>
+      <StyledContent>{content}</StyledContent>
+    </StyledContainer>
+  )
+}
+
+Scoreboard.propTypes = {
+  data: PropTypes.arrayOf(
```
izzappel, line 54: suggestion(blocking): I think this is not 100% correct.

Now it would be something like this:
```
data = [{key: 1, entity: '', defaultDisplay: ''}, {points: 1, ranking: 1}]
```
This means: an _Array_ of two separate _Objects_. First Object has three properties (_key_, _entity_, _defaultDisplay_) and second Object has two properties (_points_, _ranking_)

but it should be:

```
data = [{participantBean: {key: 1, entity: '', defaultDisplay: ''}, points: 1, ranking: 1}]
```
so you have an _Array_ of _Objects_. Those Objects have three properties (_participantBean_, _points_, _ranking_) and _participantBean_ is an Object with three properties (_key_, _entity_, _defaultDisplay_)

izzappel (changes requested):

izzappel (approved):

PR #3209, opened 2022-11-11
refactor(widgets): remove div wrapper

yilkur:
Cherry-pick: Up
Refs: TOCDEV-6197
Changelog: remove div wrapper of entity-browser as it was preventing it from filling the full available height when embedded in the old client

### DIFF
--- a/packages/widgets/entity-browser/src/main.js
+++ b/packages/widgets/entity-browser/src/main.js
@@ -18,11 +18,9 @@ const packageName = 'entity-browser'
 
 const LazyEntityBrowserComp = React.lazy(() => import('./components/EntityBrowser'))
 const LazyEntityBrowser = () => (
-  <div>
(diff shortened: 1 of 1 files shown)


packages/widgets/entity-browser/src/main.js
```diff
@@ -18,11 +18,11 @@ const packageName = 'entity-browser'
 
 const LazyEntityBrowserComp = React.lazy(() => import('./components/EntityBrowser'))
 const LazyEntityBrowser = () => (
-  <div>
+  <>
```
izzappel, line 21: nitpick: remove unused Fragment

actually `<></>` is not needed here

izzappel (approved):

izzappel:
@yilkur commit message is wrong!

izzappel (approved):

PR #3212, opened 2022-11-14
feat(app-extensions): use selection when loading report settings

PadPalon:
Refs: TOCDEV-6247
Changelog: send selection to adjusted POST endpoint when loading report settings


PadPalon:
wrong version

PR #3239, opened 2022-11-22
Pr/public dms folder selection action

izzappel:


packages/widgets/docs-browser/src/components/DocsView/DocsView.js
```diff
@@ -108,7 +110,6 @@ const DocsView = props => {
 
   useEffect(() => {
     changeListParent(parent)
-    setSelection([])
```
stewue, line 111: question(blocking): removing this line has no negative impact on the normal dms?

stewue (approved):

packages/core/tocco-ui/src/Select/StyledComponents.js
```diff
@@ -192,6 +193,9 @@ export const reactSelectStyles = outerTheme => {
     }),
     input: (base, state) => {
       let indicatorWidth = 50
+      if (state.selectProps.hasDocsTreeSearch) {
+        indicatorWidth += 25
+      }
       if (state.selectProps.hasAdvancedSearch) {
```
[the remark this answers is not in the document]
izzappel: Yes, this is because it should be possible to show advancedSearch and docsTreeSearch in the future.
Screen Shot 2022-11-24 at 08 05 13

packages/widgets/docs-browser/src/components/DocsView/DocsView.js
izzappel: Hmm I did not saw any issues.. I re-tested it an realized that we have a different behaviour now

stewue (approved):

PR #3245, opened 2022-11-23
feat(admin): Add system-menu shortcut to the Admin

LorenzKieran:


izzappel (changes requested):
👌 Changes are fine.
But the commit message is not yet correct.
This is a productive change now and we need the add the jira ticket ref and changelog message:
```
feat(admin): ....

Refs: TOCDEV-5237
Changelog: add system menu shortcut
```

The changelog should start with a lowercase character.

izzappel (approved):
todo: only one line for Changelog in commit message

PR #3274, opened 2022-12-01
refactor(scheduler): remove split pane

yilkur:
Refs: TOCDEV-6193
Changelog: remove the split pane feature as it doesn not improve UX and wrap scheduler in generic sidepanel component


izzappel (approved):
praise: 😍 🥳

PR #3289, opened 2022-12-05
allow non-fullscreen custom actions in menu

PadPalon:
Needs backend change https://git.tocco.ch/c/nice2/+/46780

### DIFF
--- a/packages/apps/admin/src/components/Navigation/menuType/ActionEntry.js
+++ b/packages/apps/admin/src/components/Navigation/menuType/ActionEntry.js
@@ -1,23 +1,49 @@
 import PropTypes from 'prop-types'
+import {actions} from 'tocco-app-extensions'
 import {Icon} from 'tocco-ui'
 
-import {StyledIconLink, StyledMenuLink, StyledMenuLinkWrapper} from './StyledComponents'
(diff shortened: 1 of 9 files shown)


packages/apps/admin/src/components/Navigation/menuType/ActionEntry.js
```diff
@@ -1,23 +1,49 @@
...
-  <StyledMenuLinkWrapper>
-    <StyledMenuLink data-quick-navigation={true} onClick={onClick} to={`/e/action/${item.name}`}>
-      {item.label}
-    </StyledMenuLink>
-    <StyledIconLink to={`/e/action/${item.name}`} target="_blank" rel="noreferrer">
-      <Icon icon="external-link" />
-    </StyledIconLink>
-  </StyledMenuLinkWrapper>
-)
+const ActionEntry = ({onClick, item}) =>
```
izzappel, line 7: suggestion(non-blocking): I personally prefer an if-statement to check for correct data

```
const ActionEntry = ({onClick, item}) => {
  if (!isCustomAction(item)) {
    return null // or return <></>
  }

  return (
    <StyledMenuLinkWrapper>
      {item.fullscreen ? (<>...</>) : (<StyledMenuText>...</StyledMenuText>)}
    </StyledMenuLinkWrapper>
  )
}
```

izzappel (approved):

PR #3290, opened 2022-12-05
feat(app-extensions): set default button in confirm modal

PadPalon:
Refs: TOCDEV-6222
Changelog: confirm modals now use whichever default action the backend declares, which means switching the styling and keyboard shortcut


PadPalon:
@izzappel First time using testing-library myself, do let me know if there are better ways to do things :)

izzappel (approved):

PR #3315, opened 2022-12-15
refactor: use template database in cypress

PadPalon:


scripts/nice2/helpers.sh
```diff
@@ -103,9 +103,20 @@ function waitForWidgetServer() {
   echo "widget-server is ready"
 }
 
-function emptyDB() {
+function setupDB() {
+  if doesTemplateExist;
```
izzappel, line 107: question: should we really support both versions?

izzappel: For backwards-compatibility reasons I would not.. Not sure if we really want to maintain the 3 tests we have on > 3.4

scripts/nice2/helpers.sh
```diff
@@ -103,9 +103,20 @@ function waitForWidgetServer() {
   echo "widget-server is ready"
 }
 
-function emptyDB() {
+function setupDB() {
```
izzappel, line 106: question: What is the difference between `setupDB` and `restoreDB`? 

Couldn't we just use one of those?

scripts/nice2/run.sh
```diff
@@ -25,32 +25,34 @@ function runDbRefactoring() {
...
 
 function initDB () {
-  waitForNice2
-
+  previousPwd=$(echo $currentPwd)
   cd $(echo $currentPwd)
 
   setupCypressUser
 
   createDump
```
izzappel, line 33: question: And then we can remove the `dump`?

scripts/nice2/helpers.sh
PadPalon: nothing anymore :)

scripts/nice2/helpers.sh
PadPalon: I've removed the creation of the dump, but I'd suggest we keep the reading of the dump for now. It's an easy way to run cypress on a specific database state without having to restore it yourself, and makes the transition to a template database seamless for anyone that already has cypress setup the old way.

(It's also a neat fallback in case I've included any breaking bugs ... :) )

izzappel (approved):

PR #3319, opened 2022-12-16
Pr/rtl migration

yilkur:


packages/core/tocco-ui/src/Button/Button.spec.js
```diff
@@ -1,97 +1,108 @@
...
     describe('RouterLinkButton', () => {
       test('should show label', () => {
-        const wrapper = enzymeUtil.mountEmbedded(
+        testingLibrary.renderWithIntl(
           <MemoryRouter>
             <RouterLinkButton label="test" />
           </MemoryRouter>
         )
-        expect(wrapper.find('a').text()).to.equal('test')
+        expect(screen.getByText('test')).exist
```
izzappel, line 84: suggestion(non-blocking): since this is expected to be a link we could get by role 

```
screen.getByRole('link', { name: 'test' })
```

packages/core/tocco-ui/src/Button/Button.spec.js
```diff
@@ -1,97 +1,108 @@
...
-        const wrapper = enzymeUtil.mountEmbedded(
+        testingLibrary.renderWithIntl(
           <MemoryRouter>
-            <RouterLinkButton icon="icon" />
+            <RouterLinkButton icon="icon-cog" />
           </MemoryRouter>
         )
-        expect(wrapper.find(Icon)).to.have.length(1)
+
+        expect(screen.getByRole('link').getAttribute('icon')).match(/icon-cog/)
```
izzappel, line 105: suggestion(non-blocking): instead of checking an attribute here we could check the actual icon

```
const cogIcon = await waitFor(() => screen.getByTestId('icon-cog'))
expect(cogIcon).to.exist
```

It's most probably a "bug" that the `icon` is an `attribute` of the `a` Element. It should have been removed since it's a prop not a HtmlAttribute. But there is this issue when we styled `Link` components that sometimes some props will be added to the HtmlElement.

See here:
https://github.com/tocco/tocco-client/commit/80d773d945c7cf89f7a39ec884e2329b80ce9c22

packages/core/tocco-ui/src/EditableValue/EditorProvider.spec.js
```diff
@@ -1,15 +1,15 @@
...
 describe('tocco-ui', () => {
   describe('EditableValue', () => {
     describe('typeFormatterProvider', () => {
       test('should render a type and set props', () => {
-        const wrapper = mount(<EditorProvider componentType="string" value="test" />)
+        render(<EditorProvider componentType="string" value="test" />)
 
-        expect(wrapper.find('StringEdit')).to.have.length(1)
-        expect(wrapper.find('StringEdit').props().value).to.eql('test')
+        expect(screen.getByTestId('string')).exist
```
izzappel, line 11: question(blocking): Is this really a helpful assertion?

I prefer to remove `data-testid` on the `EditorProvider` and remove this assertion since it does not tell us if the EditorProvider really choosed the right component.
In the old test it checked if the mapping was working and the `StringEdit` was rendered.

How about using this assertion instead:

```
expect(screen.getByRole('input')).to.exist
```

packages/core/tocco-ui/src/Panel/Panel.spec.js
```diff
@@ -1,79 +1,84 @@
-import {mount} from 'enzyme'
+import {screen, render, fireEvent} from '@testing-library/react'
```
izzappel, line 1: praise: 👌 

I think these tests are way more useful now

izzappel (approved):

